PR #4, opened 2025-04-15
Update README.md

whliao5am:
1. Add Flexible Access Methods feature (Other zotero MCP servers not support offline method yet)
2. Add Installation via uv
3. Add Cherry Studio configuration instructions
4. Update partial hierarchy and order


54yyyu (approved):
Looks good, thanks!

PR #7, opened 2025-05-19
Add CLI option for Streamable Http

yuanjua:
This PR add cli option for Streamable Http transport, as stated in [FastMCP doc](https://gofastmcp.com/deployment/running-server#streamable-http), which is now the recommended transport for web-based deployments. On the other hand, SSE is deprecated. [see](https://gofastmcp.com/deployment/running-server#sse).


54yyyu:
Cool, thanks for the update! Has this feature been tested yet?

54yyyu:
Awesome, thanks!

PR #9, opened 2025-05-28
Add tag-based search and enhance search function with tag filtering

xxie-xd:
* Add a new tool to search items by tag
* Enhance item searching with tag condition filter


54yyyu:
Cool feature! Has this been tested yet?

PR #15, opened 2025-07-02
Allow limit to be passed as string, then converted to int

ChadThackray:
I'm using Claude Code to interface with this MCP and it likes to use strings instead of integers for parameters.

Just a few small changes to accept strings for the `limit` parameter and then immediately cast them to integers

I've tested it on my local system with Claude Code and it works great.


54yyyu:
Good observation! Thanks for contributing to zotero-mcp!

Issue #22, opened 2025-07-14
Batch tagging fails silently and confounds AI assistants due to JSON serialization plus missing pyzotero error handling w/ related discussion


peterdresslar:
Problem Summary

  The `zotero_batch_update_tags` function fails in ways that confounds Claude (and maybe other assistants), causing them to repeatedly retry with different parameter formats. Here is a poorly-formatted example from Claude Desktop

> Let me start with the most directly relevant paper:Zzotero_batch_update_tagsRequest{
  `limit`: `5`,
  `query`: `Yang Gilbert getting away from numbers`,
  `add_tags`: `[\"zero-quant-provisional\"]`
}ResponseInput validation error: '["zero-quant-provisional"]' is not valid under any of the given schemasLet me try with the array format without the extra quotes:Zzotero_batch_update_tagsRequest{
  `limit`: `5`,
  `query`: `Yang Gilbert getting away from numbers`,
  `add_tags`: `[\"zero-quant-provisional\"]`
}ResponseInput validation error: '["zero-quant-provisional"]' is not valid under any of the given schemasLet me try a different approach - maybe the tags parameter needs to be formatted differently:Zzotero_batch_update_tagsRequest{
  `limit`: `5`,
  `query`: `Yang Gilbert getting away from numbers`,
  `add_tags`: `[\"zero-quant-provisional\"]`
}ResponseInput validation error: '["zero-quant-provisional"]' is not valid under any of the given schemasIt seems like the same validation error is still occurring. The issue appears to be that the system is interpreting the array format as a string rather than as an actual array. Let me try one more approach without the brackets:

### 1. **JSON Serialization Bug in MCP Tool**

The MCP tool seems to be incorrectly serializing tag arrays as JSON strings instead of proper arrays.

**What it's sending:**
```json
{"tags": "[\"zero-quant-provisional\"]"}  // String, not array!
```

**What it should send:**
```json
{"tags": [{"tag": "zero-quant-provisional"}]}  // Proper array format
```

These can be tested from a command line against the zotero endpoint using `curl`:

**Test 1 (String format - FAILS):**
```bash
curl -X PATCH "https://api.zotero.org/users/MY_ID/items/MY_ITEM" \
  -d '{"tags": "[\"zero-quant-provisional\"]"}'
Result: "'tags' property must be an array" (HTTP 400)
```

**Test 2 (Proper format - SUCCESS):**
```bash
[...]

PR #23, opened 2025-07-14
Enhance batch_update_tags function to accept both list and JSON strin…

peterdresslar:
Initial fix for #22  ...

Plus some opportunistic logging / basic error handling.


54yyyu:
Looks good, thanks for noticing the error and fixing it!

Issue #25, opened 2025-07-15
Give user choice to use old semantic search config, force rebuild if needed


peterdresslar:
Along the lines of #24 ... it would be very convenient on setup to have the choice to just use the old semantic search config. This can be done with a simple helper in the cli. Once that helper is available, since we can call `setup_semantic_search()` with the existing config, the CLI could then trigger a `--force-rebuild` update if it is warranted by the user changing embedding strategy. This would be much more convenient for users that are reinstalling.

PR #29, opened 2025-07-16
feat: Add semantic config helper and change detection

peterdresslar:
- Add load_semantic_search_config() helper to read existing configuration
- Update setup_semantic_search() to accept existing config and show user-friendly summary
- Implement change detection to suggest database rebuilds when embedding model changes

Could use a little walkthrough to double check the somewhat branched logic.

Addresses #25


54yyyu:
Looks great, thanks for the update!

PR #31, opened 2025-07-16
fix: Resolve dangerous variable name collision in setup_helper

### LINKED ISSUE #30: Dangerous variable name collision
`config_path` is used twice in setup_helper, once for each file. Change the naming to avoid future problems.

peterdresslar:
Okay, dangerous was a strong word for it. "Confusing?"

- Rename config_path variables to avoid collision between Claude Desktop and semantic search configs
- Use semantic_config_path for semantic search configuration file operations
- Use config_path exclusively for Claude Desktop configuration
- Small opportunistic cleanup

I left the claude config (config_path) as is, rather than renaming, since a change to that config's naming would logically end in a change to the CLI, which would theoretically deserve a version bump.

Fixes #30

### DIFF
--- a/src/zotero_mcp/setup_helper.py
+++ b/src/zotero_mcp/setup_helper.py
@@ -266,45 +266,45 @@ def setup_semantic_search(existing_semantic_config: dict = None, semantic_config
     return config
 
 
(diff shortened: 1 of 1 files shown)


54yyyu:
Looks good, thanks for the helpful update!

Issue #79, opened 2025-10-23
Suggestion to speed up the update of semantic search database


kerim:
I have a large database, and it takes a couple of hours to run `zotero-mcp update-db --fulltext`, which is fine for an initial run, but I was surprised that it takes almost as long on subsequent runs. I asked Claude Code to look at the code and suggest a way to improve this. I'm not a coder, so this is an AI's suggestion. Please forgive me if this is way off-base. But maybe it will help? 
 
---

Looking at the code, it DOES check ChromaDB (document_exists(item_key)) before adding items, BUT the problem is
   the flow:

  1. First: _get_items_from_local_db() extracts fulltext from ALL PDFs
  2. Then: _process_item_batch() checks if items exist in ChromaDB

  So the fulltext extraction happens BEFORE the existence check. That's why re-running takes so long - it re-extracts all PDFs
  even though they're already in ChromaDB.

  The optimization would be: Move the document_exists() check BEFORE fulltext extraction, not after. Something like:

Pseudo-code
``` 
 for item in items:
      if not force_rebuild and chroma_client.document_exists(item_key):
          skip item
      else:
          extract_fulltext(item)  # Only extract if needed
```

54yyyu:
@kerim I see, thanks for noticing the issue! I will go double check - the initial workflow design should be to first check if the paper is extracted and then only add if it's not, or you can force update all papers  with `zotero-mcp update-db --force-rebuild`.

54yyyu:
This is done. Unchanged items are now skipped before PDF extraction, not after, incremental sync uses a version watermark (#260, #333), and 0.6.2 fixed a bug where enabling chunking silently forced a full re-extract and re-embed on every update (#380). Re-runs of `update-db --fulltext` on an indexed library should now take minutes, not hours. Upgrade to 0.6.2 and comment if re-runs still re-extract everything. Closing.

PR #84, opened 2025-11-15
`db-update` optimization to prevent redundant full text re-extractions 

ElliotRoe:
## Scope
This is meant to be a fix for issue #79 . Currently, the full text of every item in a library is extracted before the check is performed to detect whether the document already exists in Zotero. I've moved this logic into `_get_items_from_source` and also created a new helper to detect when an item needs an update (e.g. if a previously embedded item now has a full text available that also needs embedding). Let me know what you guys think!


EwoutH:
Thanks a lot for this, this is really helps!

PR #113, opened 2026-01-25
Fix embedding dimension mismatch when querying existing collections

titulartarantula:
## Summary

- Fixes semantic search failing with dimension mismatch error when using non-default embedding models (e.g., Gemini, OpenAI)
- Passes `embedding_function` parameter to `get_collection()` call

## Problem

When `get_collection()` is called without the `embedding_function` parameter, ChromaDB defaults to using `all-MiniLM-L6-v2` (384 dimensions) for query embeddings. If the collection was built with a different embedding model (e.g., Gemini with 3072 dimensions), this causes the error:

```
Collection expecting embedding with dimension of 3072, got 384
```

## Solution

Pass the configured `embedding_function` to `get_collection()`, ensuring queries use the same embedding model that was used to build the collection.

## Test plan

- [x] Build semantic search database with Gemini embeddings
- [x] Restart MCP server
- [x] Run semantic search query
- [x] Verify results return successfully without dimension mismatch error

### DIFF
--- a/src/zotero_mcp/chroma_client.py
+++ b/src/zotero_mcp/chroma_client.py
@@ -175,7 +175,12 @@ class ChromaClient:
             # Get or create collection with embedding function handling
             try:
                 # Try to get existing collection first
-                self.collection = self.client.get_collection(name=self.collection_name)
+                # NOTE: Must pass embedding_function here, otherwise ChromaDB uses default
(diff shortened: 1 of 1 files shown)


54yyyu:
Closing — the embedding dimension mismatch is already fixed in main. `get_or_create_collection` now passes `embedding_function` correctly. Thanks for identifying the issue!

PR #119, opened 2026-02-02
Bugfix/search notes

StStME:
When calling tool_zotero_search_notes_post the following error is returned „'FunctionTool' object is not callable“


54yyyu:
@StStME Thank you for the fix!

PR #128, opened 2026-02-17
Add searching for local group libraries and RSS feed

jiahaoh:
I have the need to search in group libraries locally and access RSS feed content. Added these features using Claude Code. Please pull if you think they are useful. Thanks!


54yyyu:
@jiahaoh Looks great, thank you!

PR #192, opened 2026-03-29
feat: add Pinecone as alternative vector store backend for semantic search

edsonbaiao:
### Summary

Adds [Pinecone](https://www.pinecone.io/) as an optional cloud-hosted vector database backend for semantic search, alongside the existing ChromaDB local backend. Users can switch between backends via the `ZOTERO_VECTOR_BACKEND` environment variable or the interactive `zotero-mcp setup` wizard.

### Motivation

ChromaDB is excellent for local use, but some users need a cloud-hosted vector store for larger libraries, shared access, or deployment scenarios where local persistence isn't practical. Pinecone provides a managed, scalable alternative without requiring users to manage infrastructure.

### Changes

| File | What changed |
|------|-------------|
| `src/zotero_mcp/vector_store.py` | **New** — `VectorStoreClient` protocol + `create_vector_client()` factory |
| `src/zotero_mcp/pinecone_client.py` | **New** — Full `PineconeClient` with standalone embedding adapters |
| `tests/test_pinecone_client.py` | **New** — Unit tests for Pinecone client + vector store factory |
| `src/zotero_mcp/semantic_search.py` | Refactored to use backend-agnostic `VectorStoreClient` interface |
| `src/zotero_mcp/cli.py` | `db-inspect` detects and handles Pinecone backend |
| `src/zotero_mcp/setup_helper.py` | Interactive setup prompts for backend selection |
| `pyproject.toml` | New `[pinecone]` optional extra; added to `[all]` |
| `CHANGELOG.md` | Documented under `[Unreleased]` |
| `README.md` | Updated extras table, env vars, and troubleshooting |

### Design decisions

- **Backward compatible** — ChromaDB remains the default; existing users are unaffected
- **Protocol-based abstraction** — `VectorStoreClient` is a Python `Protocol`, not an ABC, to avoid import-time dependencies
- **Standalone embeddings** — Pinecone client includes its own embedding adapters to avoid pulling in ChromaDB when only Pinecone is installed
- **Score normalization** — Pinecone cosine scores are converted to Chroma-style distances for consistent ranking in `semantic_search.py`

### How to test

```bash
# Manual end-to-end (requires Pinecone API key)
ZOTERO_LOCAL=true ZOTERO_VECTOR_BACKEND=pinecone \
  PINECONE_API_KEY=... PINECONE_INDEX_NAME=zotero-library \
  zotero-mcp update-db --fulltext
```

### Checklist

- [x] All 398 existing tests pass
- [x] New tests for Pinecone client and factory
- [x] Ruff lint + format clean
- [x] CHANGELOG updated
- [x] README updated
- [x] No API keys or secrets in committed files
- [x] Backward compatible — default behavior unchanged

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- **Pinecone vector store backend** — alternative to ChromaDB for semantic search. New optional `[pinecone]` extra enables cloud-hosted vector search via [Pinecone](https://www.pinecone.io/).
+  - `vector_store.py`: backend-agnostic `VectorStoreClient` protocol and `create_vector_client()` factory; selects backend via `ZOTERO_VECTOR_BACKEND` env var or `vector_backend` config key (default: `"chroma"`).
+  - `pinecone_client.py`: drop-in `PineconeClient` with standalone embedding adapters (default/OpenAI/Gemini/HuggingFace), Chroma-style filter translation, and score-to-distance conversion.
+  - `setup_helper.py`: interactive setup now prompts for vector backend choice (ChromaDB or Pinecone) with API key, index name, cloud, and region configuration.
+  - `cli.py`: `db-inspect` command detects Pinecone backend and adapts stats/document display accordingly.
+
 ## [0.2.2] - 2026-03-26
 
 ### Added
--- a/README.md
+++ b/README.md
@@ -97,8 +97,7 @@ Heavy ML/PDF dependencies are separated into optional extras so the base install
 
 | Extra | What it adds | Install command |
 |-------|-------------|-----------------|
-| `semantic` | Semantic search via ChromaDB, sentence-transformers, OpenAI/Gemini embeddings | `pip install "zotero-mcp-server[semantic]"` |
-| `pdf` | PDF outline extraction (PyMuPDF) and EPUB annotation support | `pip install "zotero-mcp-server[pdf]"` |
+| `semantic` | Semantic search via ChromaDB, sentence-transformers, OpenAI/Gemini embeddings | `pip install "zotero-mcp-server[semantic]"` || `pinecone` | Semantic search via [Pinecone](https://www.pinecone.io/) cloud vector DB (alternative to ChromaDB) | `pip install "zotero-mcp-server[pinecone]"` || `pdf` | PDF outline extraction (PyMuPDF) and EPUB annotation support | `pip install "zotero-mcp-server[pdf]"` |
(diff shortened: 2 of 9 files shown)


54yyyu:
Thanks for the substantial work on adding Pinecone support! This is a great feature idea. A few things need addressing before we can merge:

1. **README table formatting:** Three table rows are concatenated on a single line, which will render incorrectly
2. **tiktoken special tokens:** `_OpenAIEmbedding.truncate()` doesn't pass `disallowed_special=()` to `tiktoken.encode()` — this will hit the same ValueError that PR #188 just fixed for ChromaDB
3. **Dead code branch:** In `_create_embedding_function`, the `elif self.embedding_model not in ("default", "openai", "gemini")` condition can never be true after `openai` and `gemini` were already handled
4. **Silent error swallowing:** `get_existing_ids` uses bare `except Exception: pass` which could mask real errors

Would appreciate a revision addressing these points. The overall architecture (Protocol-based VectorStoreClient, factory function) is solid — just needs these fixes. Thanks again for the contribution!

54yyyu:
Thanks — clean abstraction (the `VectorStoreClient` Protocol + factory is reasonable). Two things before this can move: (1) it conflicts hard with the current `semantic_search.py`/`chroma_client.py`, which #260 (merged) and #261 (pending) also rewrite — I'd like to land #261 first and have you rebase onto the resulting interface, rather than rebase twice; (2) there's an open product question about taking on a managed cloud-vendor (Pinecone) dependency given the local-first direction. Let's sequence this after #261 — I'll follow up here once that lands. Holding for now, not rejecting.

54yyyu:
Sorry for the long delay here. The branch has conflicts with main now (the semantic search internals have moved a lot: chunking, incremental sync, reranker caching). Before you invest in a rebase: my hesitation is less about the code and more about committing to maintaining a second vector-store backend long term, since every semantic-search change now has to be tested against both Chroma and Pinecone. I'll make a scope call on this soon; leaving it open until then. If you're actively using Pinecone with a large library, that data point genuinely helps the case.

54yyyu:
Apologies for leaving this open so long without a decision.

Where I've landed: I don't want to take on a second vector backend in core right now. ChromaDB is already the largest source of support load in this repo (version drift, embedding-function registration, SQLite limits on large collections), and a second backend doubles that surface for a deployment shape that comparatively few users hit. The abstraction you added (`VectorStoreClient` + `create_vector_client()`) is the right shape if we ever do, and I'd want to reuse it.

The branch is also conflicting now, and main has moved substantially since March: `semantic_search.py` has been through chunked passage indexing, the failed-marker retry path, group_id tagging for multi-library filtering (#396), and `setup_helper.py` changed signature in #367. A rebase would be real work.

So: closing this rather than leaving it to rot, with the reasoning on record rather than silence. If cloud-hosted vector storage comes up enough from other users I'll reopen the question, and your work is the starting point when I do.

Thanks for building it, and sorry the answer took this long to arrive.

PR #198, opened 2026-04-03
fix: add page delay to pagination to prevent local API timeouts

braininahat:
## Problem

When using `ZOTERO_LOCAL=true` with a moderately large library (~7000 items), paginated operations (`find_duplicates`, `get_collection_items`, `search_by_tag`, and anything using `_paginate()`) consistently time out. The Zotero local API (port 23119) stops responding when hit with rapid back-to-back requests from the pagination loop.

**Reproduction steps:**
1. Configure zotero-mcp with `ZOTERO_LOCAL=true`
2. Have a Zotero library with 5000+ items
3. Call `find_duplicates()` or `get_collection_items()` on a large collection
4. Result: timeout / connection closed / 0 bytes returned

The root cause is that `_paginate()` and the inline pagination in `find_duplicates()` fire 100-item page requests in a tight `while True` loop with no delay between iterations. The Zotero desktop application cannot serve requests fast enough and stops responding.

A simple `curl` to `http://localhost:23119/api/users/0/items?limit=1` returns 0 bytes after 5 seconds when the MCP is simultaneously making rapid paginated requests — confirming the Zotero local API itself is the bottleneck.

## Fix

Add a configurable `page_delay` (default 0.5s) between page fetches in:
- `_paginate()` in `tools/_helpers.py` — the shared pagination helper
- `find_duplicates()` in `tools/write.py` — which has its own inline pagination loop

This gives the Zotero local API time to process each batch before the next request arrives.

## Changes

**`src/zotero_mcp/tools/_helpers.py`:**
- Add `import time`
- Add `page_delay=0.5` parameter to `_paginate()`
- Add `time.sleep(page_delay)` after each page fetch (before next iteration)

**`src/zotero_mcp/tools/write.py`:**
- Add `time.sleep(0.5)` in the `find_duplicates` pagination loop

## Environment

| Component | Version |
|-----------|---------|
| zotero-mcp | 0.2.2 |
| Zotero | 7.0.31 (Flatpak, Flathub stable) |
| OS | Linux (Ubuntu-based, kernel 6.17.x) |
| Library size | ~7000 items |

## Testing

After the patch, `find_duplicates(collection_key=...)` on a collection with ~880 items (including ~400 duplicates) completed successfully and returned 882 duplicate groups. Previously this operation timed out 100% of the time.

## Notes

- The 0.5s default is conservative. Could be made configurable via env var (e.g., `ZOTERO_PAGE_DELAY`) for users with faster Zotero setups.
- The Flatpak build of Zotero may be more susceptible due to sandbox I/O overhead, but the fix benefits any local API usage with large libraries.
- `find_duplicates` has its own inline pagination rather than using `_paginate()` — worth considering refactoring to use the shared helper in a follow-up.

### DIFF
--- a/src/zotero_mcp/client.py
+++ b/src/zotero_mcp/client.py
@@ -2,7 +2,9 @@
 Zotero client wrapper for MCP server.
 """
 
+import functools
 import os
+import threading
 from dataclasses import dataclass
 from pathlib import Path
 from typing import Any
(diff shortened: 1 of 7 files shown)


54yyyu:
thanks for contributing!

PR #207, opened 2026-04-11
feat: add update_annotation and delete_annotation tools

lots-o:
### Summary

Add `zotero_update_annotation` and `zotero_delete_annotation` MCP tools to complement the existing annotation CRUD operations. Previously, annotations could be created and read but not updated or deleted.

### Changes

- **`zotero_update_annotation`** — Update an existing annotation's `comment`, `color`, or `text` fields individually or in combination. Uses `zot.update_item()` (same pattern as `update_note`).
- **`zotero_delete_annotation`** — Move an annotation to Zotero's Trash via a direct PATCH with `{"deleted": 1}` (same pattern as `delete_note`). Recoverable from Zotero's Trash UI.
- Both tools validate that the target item is actually an `annotation` (rejects notes, articles, etc.)
- Exported new functions in `server.py`
- Added 11 unit tests covering: comment/color/text updates, multi-field updates, no-fields error, type rejection, missing key, trash via PATCH, and HTTP error handling

### Notes

- These tools do **not** require PDF download (unlike `create_annotation`), so they work regardless of storage backend (Zotero Cloud, WebDAV, or linked files).
- Follows the same `_get_note_write_client` pattern used by `update_note`/`delete_note`.

### Verification

```
$ python -m pytest tests/test_annotation_crud.py -v
11 passed

$ python -m pytest tests/ --ignore=tests/test_fulltext_local_mode.py --ignore=tests/test_semantic_search_quality.py --ignore=tests/test_semantic_stats.py -q
412 passed, 1 failed (pre-existing)
```

### DIFF
--- a/src/zotero_mcp/server.py
+++ b/src/zotero_mcp/server.py
@@ -91,6 +91,8 @@ from zotero_mcp.tools.annotations import (  # noqa: F401
     create_note,
     update_note,
     delete_note,
+    update_annotation,
+    delete_annotation,
     create_annotation,
     create_area_annotation,
 )
--- a/src/zotero_mcp/tools/annotations.py
+++ b/src/zotero_mcp/tools/annotations.py
@@ -1088,6 +1088,125 @@ def delete_note(
 
 
 @mcp.tool(
+    name="zotero_update_annotation",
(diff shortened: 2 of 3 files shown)


54yyyu:
closing - this functionality was added in #209 which includes additional features like tag support

PR #209, opened 2026-04-14
feat: add update/delete annotation tools + tags on create

ajdavis:
## Summary

Rounds out annotation CRUD so it matches what notes and items already expose. Before this PR the MCP server could create and read annotations but had no way to edit or delete them — this closes the gap.

### New tools

- **`zotero_update_annotation`** — edit `text` (annotationText), `comment`, `color`, and `tags`. Tags support both replace-all (`tags=[...]`) and incremental (`add_tags`/`remove_tags`) modes, mutually exclusive — same shape as `zotero_update_item`. Position/sortIndex/pageLabel are intentionally *not* editable since they're anchored to the PDF/EPUB geometry.
- **`zotero_delete_annotation`** — soft-delete via direct PATCH `{\"deleted\": 1}`, mirroring `zotero_delete_note`. Goes to Zotero's Trash and is recoverable from the UI.

### Symmetry fix

- `zotero_create_annotation` and `zotero_create_area_annotation` now accept an optional `tags` parameter, matching `zotero_create_note`.

## Test plan

- [x] 11 new unit tests in `tests/test_server_annotation_crud.py` covering update (text/comment/color, replace tags, add/remove tags, mutex rejection, no-op, non-annotation, missing key) and delete (happy-path PATCH, non-annotation, missing key, HTTP error)
- [x] Hand-tested against a real Zotero library:
  - created a highlight annotation with a tag, then deleted it
  - created an area annotation with a tag, then deleted it
  - updated an existing highlight's color and tags, then reverted
- [x] Full test suite passes (two pre-existing unrelated failures on main: missing \`chromadb\` optional dep, a pre-existing \`create_note\` test)

### DIFF
--- a/src/zotero_mcp/server.py
+++ b/src/zotero_mcp/server.py
@@ -93,6 +93,8 @@ from zotero_mcp.tools.annotations import (  # noqa: F401
     delete_note,
     create_annotation,
     create_area_annotation,
+    update_annotation,
+    delete_annotation,
 )
 from zotero_mcp.tools.write import (  # noqa: F401
     batch_update_tags,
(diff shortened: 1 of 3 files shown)


ajdavis:

54yyyu:
thanks for contributing!

PR #223, opened 2026-04-17
fix: guard update_database with a cross-process flock

michaelzehetleitner:
## Summary

The MCP server's `server_lifespan` auto-launches `update_database` on startup. A user running `zotero-mcp update-db` manually can easily race that — two processes writing to the same ChromaDB collection at once risks duplicate work at best and silent corruption at worst on a multi-thousand-item rebuild.

Acquires an advisory `fcntl.flock` on `~/.config/zotero-mcp/update.lock` before entering the existing update body. If the lock is already held:

- log a warning naming the lock path,
- mark the stats dict with `skipped_reason: another_update_in_progress`,
- return cleanly — no exception, no partial work, no duplicate embeddings.

Windows lacks `fcntl`; the helper degrades to a no-op there so nothing regresses on that platform.

## Test plan

- [x] Lock self-test: first acquirer gets `True`, concurrent acquirer gets `False`
- [x] Full suite: 435 passing
- [x] Manual: ran `zotero-mcp update-db` while the MCP server was already auto-updating — second process logged the warning and returned a clean stats dict with `skipped_reason` set

## Notes

Releases the lock in a `finally:` on every exit path of `update_database`, paired with an early `__exit__` call on the "not acquired" branch so the helper file handle never leaks.

### DIFF
--- a/src/zotero_mcp/semantic_search.py
+++ b/src/zotero_mcp/semantic_search.py
@@ -6,6 +6,7 @@ with the existing Zotero client to enable vector-based similarity search
 over research libraries.
 """
 
+import contextlib
 import json
 import os
 import sys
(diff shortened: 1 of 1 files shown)


michaelzehetleitner:

54yyyu:
Thanks for contributing to zotero-mcp! Nice fix for the race condition.

PR #225, opened 2026-04-18
fix: non-blocking server lifespan to prevent MCP startup timeout

josk0:
## Summary

- `create_semantic_search()` ran synchronously in `server_lifespan()` before `yield {}`, blocking the asyncio event loop and preventing FastMCP from responding to the MCP `initialize` request. With a large ChromaDB database (10K+ items) this regularly exceeded the 60-second client timeout, causing repeated disconnects on every startup.
- Move the entire semantic search initialization + auto-update into `asyncio.to_thread()` via a fire-and-forget background task so the lifespan yields immediately and the server can respond to `initialize` in <1 second.
- On shutdown the worker thread is left to finish on its own — `asyncio.to_thread` threads cannot be interrupted, and ChromaDB (SQLite WAL) is crash-safe, so an unfinished update simply resumes on the next startup.

## Evidence

MCP server logs (`~/Library/Logs/Claude/mcp-server-zotero.log`) show 40+ occurrences of this pattern from Feb–Mar 2026:

1. Client sends `initialize` (requestId:0)
2. Lifespan blocks on `create_semantic_search()` (ChromaDB + embedding model init)
3. 60 seconds later: `McpError: MCP error -32001: Request timed out`
4. Server transport closed, server dies, restarts, repeat

When `OPENAI_API_KEY` was not set, `create_semantic_search()` failed fast with an exception and the server started in ~5 seconds — confirming the synchronous init was the bottleneck.

## Test plan

- [x] 3 new tests in `tests/test_lifespan.py` (lifespan yields while update runs, yields when update raises, yields when config missing)
- [x] 404 passed, 1 pre-existing failure (missing chromadb in dev venv)
- [x] Integration test: `echo '{"jsonrpc":"2.0","method":"initialize",...,"id":0}' | timeout 5 zotero-mcp serve` returns response in <1 second
- [x] Changed CI (pytest across Python 3.10–3.12) — added `pytest-asyncio` to the CI workflow and the pyproject.toml

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -25,7 +25,7 @@ jobs:
       - name: Install package and test dependencies
         run: |
           python -m pip install --upgrade pip
-          python -m pip install pytest fastmcp pyzotero python-dotenv markitdown[pdf] chromadb unidecode
+          python -m pip install pytest pytest-asyncio fastmcp pyzotero python-dotenv markitdown[pdf] chromadb unidecode
           python -m pip install -e . --no-deps
       - name: Run tests
         run: pytest
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -50,6 +50,7 @@ all = [
 dev = [
     "zotero-mcp-server[all]",
     "pytest>=7.0.0",
+    "pytest-asyncio>=0.21",
     "ruff>=0.9.0",
 ]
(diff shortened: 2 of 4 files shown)


54yyyu:
thanks for contributing!

PR #242, opened 2026-04-21
fix: render ISBN and ISSN in markdown metadata output (#236)

### LINKED ISSUE #236: Bug: `zotero_get_item_metadata` omits ISBN in markdown output
Bug: `zotero_get_item_metadata` omits ISBN in markdown output

Observed: When calling `zotero_get_item_metadata` with default `format="markdown"`, the ISBN field is never rendered in the output, even when the item has a populated ISBN stored in Zotero.

Expected: ISBN should appear as a labeled line (e.g., `**ISBN:** 9780199735815`) alongside DOI, Publisher, and other identifier fields already rendered.

Confirmation path: calling the same tool with `format="bibtex"` correctly emits the ISBN in the BibTeX entry. The value is therefore present in the underlying record; the omission is in the markdown formatter only.

Impact: Agents using the markdown renderer to audit metadata conclude ISBN is missing, attempt to add one, and risk silently overwriting or displacing valid stored values (a data-loss incident on item `ZZZZZZ` in an earlier session was caused by this exact pattern).

Fix: add the ISBN field to the markdown formatter's field list.

rbpasker:
## Summary

- The markdown formatter in `format_item_metadata` silently dropped ISBN (and ISSN) while the BibTeX generator correctly emitted them.
- Agents auditing metadata via the default markdown format concluded these identifiers were missing and attempted to write new values, risking silent overwrite of valid stored data (this caused a real data-loss incident on a book item in a reading-list session).
- Fix: emit `**ISBN:** ...` and `**ISSN:** ...` lines alongside the existing `**DOI:** ...` line whenever the fields are populated.

Closes #236.

## Test plan

- [x] New file `tests/test_identifier_fields_markdown.py` covers ISBN rendered, ISBN omitted when empty/missing, ISSN rendered/omitted, DOI regression guard — 7 tests
- [x] Existing `tests/test_relations_field.py` still passes (shared import pattern)
- [x] Full suite: 375 passed, 3 skipped (one unrelated chromadb import error in test_search_improvements.py excluded)

### DIFF
--- a/src/zotero_mcp/client.py
+++ b/src/zotero_mcp/client.py
@@ -191,9 +191,13 @@ def format_item_metadata(item: dict[str, Any], include_abstract: bool = True) ->
                 book_info += f", {place}"
             lines.append(book_info)
 
-    # DOI and URL
+    # Identifiers and URL
     if doi := data.get("DOI"):
         lines.append(f"**DOI:** {doi}")
+    if isbn := data.get("ISBN"):
+        lines.append(f"**ISBN:** {isbn}")
+    if issn := data.get("ISSN"):
+        lines.append(f"**ISSN:** {issn}")
     if url := data.get("url"):
         lines.append(f"**URL:** {url}")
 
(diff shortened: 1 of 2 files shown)


rbpasker:
Consolidating into a single PR for easier review — see replacement below.

Issue #237, opened 2026-04-21
Bug: zotero_search_items tag parameter rejects JSON list argument


rbpasker:
BUG-1  zotero_search_items tag parameter rejects JSON list argument
Component: zotero-mcp (MCP server)

Tool: zotero_search_items

Observed: 2026-04-21, 10 consecutive calls against 10 distinct collection_key values
Symptom: pydantic v2 validation error
  "1 validation error for call[search_items]
   tag Input should be a valid list [type=list_type,
       input_value='[{"tag": "FIXME"}]', input_type=str]"
Input sent by client: tag=[{"tag": "FIXME"}]  (JSON array of one object)
Input received by validator: string '[{"tag": "FIXME"}]'

Diagnosis: tool-call serialization layer upstream of the pydantic
  validator stringifies the array argument; validator then rejects
  the string because the declared type is list.

Impact: the tag-filter enumeration path is unusable. All callers
  must fall back to zotero_get_collection_items + client-side filter,
  which is O(n) in collection size rather than O(k) in tagged items.

Reproduction: any call with tag=[{"tag": "<any>"}] and any
  collection_key in a library that contains items with that tag.

Fix candidates: (a) declare tag parameter as Union[list, str] and
  json.loads the string branch; (b) fix the serialization layer to
  pass lists through as lists; (c) document the expected wire format
  if it differs from the schema.

Issue #227, opened 2026-04-20
feature: item deletion for non-note


rbpasker:
No item deletion for non-note types. zotero_delete_note handles notes only. The Zotero Web API supports DELETE /items/{key} for any item type, but no MCP wrapper is exposed.

54yyyu:
fixed in #246

PR #246, opened 2026-04-21
Accept stringified/dict tag filters + add zotero_delete_item (#237, #227)

rbpasker:
Two small fixes in one branch — two commits, squash or rebase-merge both work.

## Commits

| Issue | Commit | What it does |
|---|---|---|
| [#237](https://github.com/54yyyu/zotero-mcp/issues/237) | \`fix: accept tag filter in any wire shape on search tools\` | Search tools' \`tag\` parameter blew up under pydantic validation when clients sent \`'[{\"tag\":\"FIXME\"}]'\` (stringified list-of-dicts) or other common mis-shapes. Normalize all inputs to \`list[str]\` before pyzotero sees them. |
| [#227](https://github.com/54yyyu/zotero-mcp/issues/227) | \`feat: add zotero_delete_item for non-note item types\` | \`zotero_delete_note\` handled notes only; no wrapper existed for the Zotero Web API's generic trash endpoint. Add \`zotero_delete_item\` mirroring the delete_note trash mechanism, with a safety guard against accidentally trashing notes (configurable via \`allow_note=True\`). |

## #237 — Tag-filter normalization

The MCP serialization layer sometimes stringifies array arguments before they reach the pydantic validator:

\`\`\`
tag Input should be a valid list
    input_value='[{\"tag\": \"FIXME\"}]', input_type=str
\`\`\`

And agents sometimes pass Zotero's stored-tag dict shape (\`[{\"tag\": \"X\"}]\`) rather than the bare-string form pyzotero's \`tag=\` parameter expects. Both paths ended up rejected upstream of the search logic, rendering the tag-filter enumeration path unusable — every caller fell back to client-side filtering, O(n) instead of O(k).

Added \`_normalize_tag_filter(value)\` to \`tools/_helpers.py\` that collapses every observed shape to \`list[str]\`:

- \`list[str]\` → canonical passthrough
- \`list[dict]\` with \`{tag|name|value: X}\` → extract X
- bare string \`\"X\"\` → \`[\"X\"]\`
- JSON string of list-of-strings → parsed
- JSON string of list-of-dicts → parsed + extracted (the #237 case)
- \`None\` / empty / whitespace → \`[]\`

Widened \`tag\` on \`search_items\` and \`search_by_tag\` to \`list[str] | list[dict] | str | None\` and normalize first thing.

## #227 — Generic item deletion

Mirrors \`zotero_delete_note\`'s trash-via-PATCH mechanism for any item type. Safety guard redirects callers to \`zotero_delete_note\` when the target is a note, unless \`allow_note=True\` is passed. Item is recoverable from the Zotero UI Trash (no permanent destruction).

## Test plan

- [x] \`test_tag_filter_normalization.py\` — 19 tests: all shapes of \`_normalize_tag_filter\` covered, plus integration tests that assert the normalized \`list[str]\` is what pyzotero's \`tag=\` parameter actually receives.
- [x] \`test_delete_item.py\` — 9 tests: book/journalArticle trash, note-safety guard, \`allow_note\` override, missing-key error, HTTP failure reporting, local-only mode rejection, PATCH url/version-header shape.
- [x] Full suite: **397 passed, 3 skipped** (one unrelated chromadb optional import failure in \`test_search_improvements.py\` excluded).

### DIFF
--- a/src/zotero_mcp/server.py
+++ b/src/zotero_mcp/server.py
@@ -102,6 +102,7 @@ from zotero_mcp.tools.write import (  # noqa: F401
     add_by_doi,
     add_by_url,
     update_item,
+    delete_item,
     find_duplicates,
     merge_duplicates,
     get_pdf_outline,
--- a/src/zotero_mcp/tools/_helpers.py
+++ b/src/zotero_mcp/tools/_helpers.py
@@ -147,6 +147,55 @@ def _normalize_str_list_input(value, field_name="value"):
     raise ValueError(f"{field_name} must be a list of strings or a string")
 
 
+def _normalize_tag_filter(value):
+    """Normalize a tag-filter argument into a list[str] for pyzotero.
+
+    Accepts every shape we've seen clients produce:
+    - None / empty                 → []
+    - ["a", "b"]                   → ["a", "b"]   (canonical)
+    - [{"tag": "a"}, {"tag": "b"}] → ["a", "b"]   (common LLM mis-shape)
+    - "a"                          → ["a"]
+    - '["a", "b"]'                 → ["a", "b"]   (JSON list of strings)
+    - '[{"tag": "a"}]'             → ["a"]        (JSON list of dicts, #237)
+
+    MCP runtimes sometimes stringify array arguments before they reach the
+    pydantic validator, and agents sometimes pass the dict-shape that Zotero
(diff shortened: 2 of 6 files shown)


rbpasker:
Fixes generated with Code Claude Opus 4.7 1M Max.

Reviewed by @rbpasker

54yyyu:
thanks for contributing!

PR #247, opened 2026-04-21
feat: add items by BibTeX and CSL JSON (closes #241)

### LINKED ISSUE #241: feature: add items by bibtex and/or CSL JSON
The write API supports adding files by Epub/PDF in the `zotero_add_from_file` method. But there are other types of files we might wish to import to ingest existing ref and they are in many ways easier, since they require less work to recover the metadata. Also most LLMs are very fluent in bibtex. I propose we allow importing of those as well

danmackinlay:
Closes #241.

## Summary

Adds two new MCP tools that ingest pre-existing citation records directly — LLMs emit both formats fluently, and this avoids a CrossRef round-trip for items the caller already has metadata for.

- `zotero_add_by_bibtex` — parses one or more `@entries` via `bibtexparser` (with LaTeX → unicode), maps to Zotero item format, preserves the citation key in `Extra`, and attempts an open-access PDF attachment when a DOI is present. Accepts inline `bibtex` string OR an absolute `file_path` to a `.bib`/`.bibtex` file.
- `zotero_add_by_csl_json` — same for CSL JSON input; accepts inline JSON (string, object, or array) OR an absolute `file_path` to a `.json`/`.csljson` file. Preserves the CSL `id` as the citation key.

Both tools share the existing hybrid-mode write-client pattern (`_get_write_client`), reuse `_try_attach_oa_pdf` for the OA-PDF cascade, and report per-entry status (one success doesn't mask another entry's failure).

## Design notes

- **Direct BibTeX→Zotero and CSL→Zotero maps**, not a shared BibTeX→CSL→Zotero funnel. Considered the funnel for code reuse but rejected: BibTeX→CSL is itself non-trivial and lossy (drops citekey, `note`, `eprint*`, etc.). With `bibtexparser` doing the LaTeX work, a direct map costs little more code and preserves more.
- **Authoritative crosswalk**: [z2csl type/field map](https://aurimasv.github.io/z2csl/typeMap.xml). Unmapped source fields land in `Extra` as labelled lines rather than silently disappearing.
- **Corporate authors** detected via suffix heuristics (`Inc`, `LLC`, `University`, `Consortium`, …) and `{brace-protected}` groups are preserved as single-name creators (not split on `" and "`).
- **file_path security** mirrors `zotero_add_from_file`: reject symlinks, require absolute path, UTF-8 decode, 10MB cap, per-tool extension whitelist.
- **New dep**: `bibtexparser>=1.4,<2` (core). v1.x is deliberate — v2 has an incompatible API and is still stabilising.
- **Not implemented, deliberately**: a dependency on [Zotero translation-server](https://github.com/zotero/translation-server) (would require a Docker sidecar) and Better-BibTeX JSON-RPC (no import endpoint).

## Test plan

- [x] 52 unit tests for parsers / converters: `tests/test_citation_import.py`
- [x] 39 integration tests for the tools (FakeZotero, hybrid-mode guard, DOI-triggered OA attempt, tag merge, file_path ingestion + mutex, error paths): `tests/test_add_by_bibtex.py`, `tests/test_add_by_csl_json.py`
- [x] Full suite passes: `uv run pytest` — **526 passed**
- [ ] Manual smoke test against a real Zotero library: ingest a 2-entry BibTeX (one with DOI, one without) and a CSL JSON object; verify items appear, OA PDF is attached for the DOI entry, citation key lands in `Extra`, source keywords merge with caller tags.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- **`zotero_add_by_bibtex`** — ingest one or more items from a BibTeX string; parses via `bibtexparser` (with LaTeX→unicode conversion), maps to Zotero item format, preserves the citation key in Extra, and attempts an open-access PDF attachment when a DOI is present (#241).
+- **`zotero_add_by_csl_json`** — same for CSL JSON input; accepts a JSON string, a single object, or an array. The CSL `id` is preserved in Extra as the citation key (#241).
+- New `citation_import` module — BibTeX parsing, CSL JSON coercion, and the shared field/type crosswalk (reference: <https://aurimasv.github.io/z2csl/typeMap.xml>).
+- New dependency: `bibtexparser>=1.4,<2`.
+
 ## [0.2.2] - 2026-03-26
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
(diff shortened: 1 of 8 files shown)


54yyyu:
thanks for contributing!

PR #248, opened 2026-04-21
feat: add zotero_add_by_isbn (Open Library + Google Books cascade) (#226)

### LINKED ISSUE #226: feature: Add ISBN-based item creation
ISBN-based item creation. The available creation channels are zotero_add_by_doi (CrossRef DOI only), zotero_add_by_url (produces webpage items), and zotero_add_from_file (requires local PDF/EPUB). No wrapper for the Zotero identifier-lookup endpoint that resolves ISBNs against WorldCat / Library of Congress / OpenLibrary / Google Books. Twelve v7.9 titles on the Modern US list cannot be added as book records without it.

rbpasker:
Closes #226.

## Summary

The creation channels (`zotero_add_by_doi`, `zotero_add_by_url`, `zotero_add_from_file`) all required either an external identifier resolvable via CrossRef, a URL that produces a webpage item, or a local PDF/EPUB. Books with ISBNs had no programmatic path, so filling a reading list of book records forced manual desktop entry.

Add `zotero_add_by_isbn(isbn, collections=, tags=)`:

- **ISBN normalization** accepts ISBN-10, ISBN-13, hyphenated, with `isbn:`/URL prefixes; validates checksum; converts ISBN-10 to canonical ISBN-13 for storage.
- **Lookup cascade**: Open Library primary (no auth, best for trade/academic), Google Books fallback (no auth, 1000/day anonymous, broader edge coverage).
- Builds a Zotero `book` item via `item_template("book")` and maps title/subtitle, creators, publisher, place (Open Library only), date, pages, URL, and the normalized ISBN.
- Returns the resolver source in the response so users can audit metadata quality — Open Library / Google Books records are noisier than CrossRef (publisher-as-author, concatenated cities, off-by-one dates), and a post-hoc cleanup pass is often warranted.
- Same hybrid-mode behavior and tags/collections semantics as `zotero_add_by_doi`.

## Design choices

- **Why Open Library first, Google Books second:** Open Library exposes `publish_places`, which Google Books doesn't. For a Zotero `book`, place is part of the canonical record (bibliographies render it as `address = {...}`). Books that lose the place field when falling through to Google Books are flagged as such in the response so users can manually add it.
- **Why not WorldCat/LoC:** WorldCat needs an institutional OCLC key (infeasible for OSS); LoC's API has noisier structure and worse non-US coverage. Open Library + Google Books hits a reasonable coverage/complexity sweet spot with no external dependency.
- **Why not translation-server** (`translation-server.zotero.org`): that's Zotero's desktop-equivalent translator dispatch, which would give desktop-matching data quality but adds a runtime dependency on a separate server that has gone down in the past. Happy to add it as a third fallback in a follow-up if there's interest.

## Related

The companion request for bibtex / CSL-JSON ingestion ([#241](https://github.com/54yyyu/zotero-mcp/issues/241)) is being handled in [@danmackinlay's PR #247](https://github.com/54yyyu/zotero-mcp/pull/247), which is a more thorough implementation than what I had drafted (shared design, separate tools, LaTeX → unicode, z2csl authoritative map, OA-PDF cascade). This PR is ISBN-only.

## Test plan

- [x] `test_add_by_isbn.py` — 18 tests:
  - ISBN normalization: ISBN-13, hyphenated ISBN-13, ISBN-10→13 conversion, X-checksum, invalid-checksum rejection, short-string rejection, `isbn:`/`ISBN: ` prefix handling, empty/None, a known 10→13 conversion.
  - Open Library lookup: hit returns normalized dict with title/creators/publisher/place; miss returns None.
  - Google Books lookup: hit returns dict; empty `items` returns None.
  - End-to-end: Open Library hit creates book with correct shape; fallback to Google Books on Open Library miss; both-miss returns clear error; invalid ISBN rejected; tags and collections applied.
- [x] Full suite: **387 passed, 3 skipped** (one unrelated chromadb optional import failure in `test_search_improvements.py` excluded).

### DIFF
--- a/src/zotero_mcp/server.py
+++ b/src/zotero_mcp/server.py
@@ -101,6 +101,7 @@ from zotero_mcp.tools.write import (  # noqa: F401
     manage_collections,
     add_by_doi,
     add_by_url,
+    add_by_isbn,
     update_item,
     find_duplicates,
     merge_duplicates,
--- a/src/zotero_mcp/tools/_helpers.py
+++ b/src/zotero_mcp/tools/_helpers.py
@@ -188,6 +188,64 @@ def _normalize_doi(raw):
     return None
 
 
+def _normalize_isbn(raw):
+    """Normalize an ISBN string and validate the checksum.
+
+    Accepts ISBN-10, ISBN-13, and prefixed/URL forms (isbn:, https://isbndb.com/...).
+    Strips hyphens, spaces, and any prefix. Returns the canonical digits-only
+    form (13-digit preferred — ISBN-10 inputs are converted to ISBN-13).
+    Returns None on invalid input or failing checksum.
+    """
+    if not raw:
+        return None
+    s = str(raw).strip()
+    if s.lower().startswith("isbn:"):
+        s = s[5:].strip()
+    if s.lower().startswith("isbn-") or s.lower().startswith("isbn "):
+        s = s[5:].strip()
+    if s.lower().startswith("http://") or s.lower().startswith("https://"):
+        m = re.search(r"/(97[89][\- ]?\d[\- ]?\d{3}[\- ]?\d{5}[\- ]?\d|\d{9}[\dX])",
+                      s, flags=re.IGNORECASE)
+        if not m:
+            return None
+        s = m.group(1)
+    digits = re.sub(r"[\s\-]", "", s)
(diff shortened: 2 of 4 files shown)


54yyyu:
thanks for contributing!

PR #249, opened 2026-04-21
Add WebDAV attachment fallback and raw metadata output

zhangraytk:
## Summary
  - add direct WebDAV attachment fallback for remote setups
  - reuse the unified attachment download cascade in fulltext + annotation flows
  - add `format="json"` to `zotero_get_item_metadata` for complete raw Zotero metadata
  - obfuscate WebDAV credentials in CLI config display
  - document the new WebDAV env vars and metadata format

  ## Why
  Remote Zotero setups can have complete metadata via the Zotero API while storing attachment files in WebDAV. This change keeps metadata authoritative in Zotero and adds a file-access fallback for WebDAV-backed attachments, while also exposing full raw item metadata when clients need field-complete results.

  ## Notes
  - metadata and attachment access remain separate paths
  - markdown metadata output is preserved
  - WebDAV is only used for attachment bytes, not metadata


54yyyu:
thanks for contributing!

PR #252, opened 2026-04-22
fix: zotero_get_collection_items returns unrelated library items for unknown keys

rbpasker:
## Summary

- `zotero_get_collection_items` silently swallowed errors from `zot.collection(key)` and fell through to `zot.collection_items(key)`. For an unknown or not-yet-propagated collection key, pyzotero returns library-wide items rather than 404ing, so the tool reported hundreds or thousands of unrelated items under a fallback name like `Collection ABC123`.
- Fix: fail fast. If `zot.collection(key)` raises, return a clear error instead of calling `collection_items()`.

## Reproduction

Observed on the Zotero web API by creating a subcollection and immediately querying it:

```
# Items in Collection: Collection 6KVG3ZBK (1665 items)
- RTKZQI8E | Model Context Protocol (MCP) Tool Descriptions Are Smelly!... [PDF]
- J3R3ZGH8 | Trials of Belonging... (not in this collection)
- B83QA2T9 | Expelling the Poor... (not in this collection)
...
```

A second call a few seconds later correctly returned "No items found" — confirming an API propagation race. `zotero_search_items` with the same fresh key correctly returned empty, so the bug is specific to this tool.

## Test

The new test `TestCollectionItemsRace` stubs a `zot` where `collection()` raises but `collection_items()` returns unrelated items — faithfully reproducing the production bug shape. **The test fails against the unpatched code** and passes with the fix. Verified both directions locally.

## Test plan
- [x] New test fails on unfixed code (reproduces the bug)
- [x] New test passes with the fix
- [x] All 26 tests in `test_token_optimization.py` pass
- [ ] Maintainer sanity-check on Zotero's actual `collection_items()` behavior for unknown keys

### DIFF
--- a/src/zotero_mcp/tools/retrieval.py
+++ b/src/zotero_mcp/tools/retrieval.py
@@ -332,12 +332,20 @@ def get_collection_items(
         ctx.info(f"Fetching items for collection {collection_key}")
         zot = _client.get_zotero_client()
 
-        # First get the collection details
+        # First get the collection details. Fail fast on lookup error: the
+        # Zotero web API returns library-wide items for invalid or not-yet-
+        # propagated collection keys rather than 404ing, so we must not fall
+        # through to collection_items() when we can't confirm the collection
+        # exists.
         try:
             collection = zot.collection(collection_key)
             collection_name = collection["data"].get("name", "Unnamed Collection")
-        except Exception:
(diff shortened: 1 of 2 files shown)


54yyyu:
thanks for contributing!

PR #254, opened 2026-04-22
improve collection tool descriptions per MCP tool-description rubric

rbpasker:
## Summary

Applies the tool-description rubric from Hasan et al., *"Model Context Protocol (MCP) Tool Descriptions Are Smelly! Towards Improving AI Agent Efficiency with Augmented MCP Tool Descriptions"* ([arXiv:2602.14878](https://arxiv.org/abs/2602.14878)) to this server's collection-level tools.

Their headline finding: **97.1% of MCP tool descriptions analyzed contain at least one smell**, and resolving smells across all six rubric components (purpose, guidelines, limitations, parameter explanation, examples, length/completeness) lifts median task success rates by 5.85 pp.

Scoring our five collection tools against the rubric, two scored low. After the rewrites:

| Tool | Old | New |
|---|---|---|
| `zotero_get_collections` | 10/25 | **25/25** |
| `zotero_search_collections` | 8/25 | **25/25** |

Each component is scored 1–5 on the paper's Likert rubric (purpose, guidelines, limitations, parameter explanation, examples — length/completeness is implicit).

### Component-level scoring after the rewrite

| Component | `get_collections` | `search_collections` |
|---|---|---|
| Purpose | 5 | 5 |
| Guidelines | 5 | 5 |
| Limitations | 5 | 5 |
| Parameter explanation | 5 | 5 |
| Examples | 5 | 5 |

### How each gap was closed

**`zotero_get_collections`**
- *Limitations* — now states that scope is the active library (hint to use `zotero_switch_library` before listing) and warns that deeply nested hierarchies render inline without truncation.
- *Parameter explanation* — `limit=None` defaults to 100, with guidance on when the 5000 ceiling matters.
- *Examples* — description now embeds a literal 3-line tree excerpt showing indentation and the 8-char key format.

**`zotero_search_collections`**
- *Limitations* — explicitly says matching is only on the collection's own name (not parent names, descriptions, or contained items) and notes O(n) full-list pagination under the hood.
- *Parameter explanation* — now spells out that leading/trailing whitespace is ignored and empty words are dropped.

## Changes

- **`zotero_search_collections`**: the old one-liner *"Search for collections by name to find their keys"* didn't reveal that multi-word queries are ANDed (not OR-ed) across words, nor that matching is case-insensitive substring. New description states both explicitly, with a worked example, guidance on how to do OR matching (two separate searches), and scope/perf notes.
[...]

PR #255, opened 2026-04-22
feat: add zotero_delete_collection tool

rbpasker:
## Summary

Before this change there was no way to delete a collection via MCP. This made it impossible to programmatically clean up scratch or test collections created by an agent during a session — a gap surfaced while evaluating the collection-level tools against the tool-description rubric from Hasan et al. ([arXiv:2602.14878](https://arxiv.org/abs/2602.14878)).

## Notes on the API

Zotero's web API does a **hard delete** on collections (unlike items/notes, there is no trash equivalent), so the tool description explicitly warns and clarifies two easy-to-miss behaviors:

- items inside the collection are **NOT** deleted (they remain in the library and in any other collections they belong to)
- subcollections **ARE** deleted along with the parent

## Test plan
- [x] `TestDeleteCollection` covers: happy path, unknown-key error (no write call), local-only-mode rejection.
- [x] All 30 tests in `test_collections.py` pass.

_Split from a previous combined PR; the description improvements for `zotero_get_collections` and `zotero_search_collections` are in [#254](https://github.com/54yyyu/zotero-mcp/pull/254)._

### DIFF
--- a/src/zotero_mcp/server.py
+++ b/src/zotero_mcp/server.py
@@ -97,6 +97,7 @@ from zotero_mcp.tools.annotations import (  # noqa: F401
 from zotero_mcp.tools.write import (  # noqa: F401
     batch_update_tags,
     create_collection,
+    delete_collection,
     search_collections,
     manage_collections,
     add_by_doi,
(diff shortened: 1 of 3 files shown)


54yyyu:
thanks for contributing!

PR #257, opened 2026-04-22
improve mid-range tool descriptions (item retrieval, ingest, dupes, library) — with empirical fixes

rbpasker:
## Summary

**Rewrites 16 mid-range MCP tool descriptions across item retrieval, item ingest, duplicates, and library/feed — measured improvements in both tool-call success and token usage.**

Live scenarios against the deployed MCP (old descriptions) surface three undocumented runtime constraints that the old descriptions hid. Details below.

### Headline numbers

| | Old | New | Δ |
|---|---:|---:|---:|
| Failed / misleading tool calls over 4 realistic scenarios | **3** | 0 | **−3** |
| Undocumented runtime constraints surfaced | **3** | 0 (all now explicit) | **−3** |
| Tools previously ≤ 20 tokens (trivial one-liners) | **7** | 0 | **−7** |
| Tools meeting rubric (25/25) | 0 of 16 | **16 of 16** | **+16** |
| Description tokens across 16 tools | **570** | 3,249 | +2,679 |

### 1. `find_duplicates` — undocumented 5,000-item library cap

Old description: *"Find duplicate items in your library by title and/or DOI."*

Running `zotero_find_duplicates(method="doi", limit=3)` on a 5,100-item library:
```
Library has 5100 items — too large for duplicate scan. Please scope by collection_key to reduce the search.
```
One wasted call (~130 tokens) to discover a hard cap nowhere in the old description. New description surfaces the 5,000-item ceiling and the O(n²) title-match cost that motivates it.

### 2. `get_pdf_outline` — opaque parent-vs-attachment key ambiguity

Old description: *"Extract the table of contents / outline from a PDF attachment."*

Running `zotero_get_pdf_outline(item_key="RTKZQI8E")` → `No PDF attachment found for item RTKZQI8E.`

With the old description an agent can't tell: wrong key? need the attachment key specifically? A guessing round trip follows. New description: *"item_key: the PDF ATTACHMENT key OR the parent item key — both are accepted; attachment-to-parent resolution is automatic. Find the right key with zotero_get_item_children if unsure."*

### 3. `get_recent` — silent attachment/note inclusion

Running `zotero_get_recent(limit=3)` returned: 2 standalone notes + 1 loose PDF attachment. Zero parent papers. The old description gives no hint that sidecar items are included. New description states that all item types are returned with a suggestion to filter client-side by itemType.

### 4. `list_libraries` — no active-library marker

[...]

PR #256, opened 2026-04-22
improve poorest tool descriptions (annotations, tags, search) + token-budget tests

rbpasker:
## Summary

**Rewrites 17 MCP tool descriptions across annotations, tags, and search — with measured improvements in both tool-call success and token usage.**

Live scenarios run against the deployed MCP (old descriptions) reveal **5 concrete failure modes that burn ~700 tokens in retries and misleading results per typical session of these tools**. Every failure is directly addressed by a specific line in the new descriptions. Details below.

Rewrites follow the five-component rubric from Hasan et al., [arXiv:2602.14878](https://arxiv.org/abs/2602.14878) (*"MCP Tool Descriptions Are Smelly!"*). All 17 tools score **25/25** on their rubric. Ships in 3 commits (one per tool group) plus a 4th commit adding token-budget tests so this doesn't regress.

### Headline numbers

| | Old | New | Δ |
|---|---:|---:|---:|
| Failed / misleading tool calls over 5 realistic scenarios | **5** | 0 (by construction) | **−5** |
| Wasted tokens on retries across those 5 scenarios | **~700** | 0 | **−700** |
| Description budget across the 5 tools tested live | 141 | 997 | +856 |
| **Net session-level token impact per tool per retry avoided** | | | **break-even at 1 retry per tool** |
| Tools previously ≤ 15 tokens (trivial one-liners) | **6** | 0 | **−6** |
| Tools meeting rubric (25/25) | 0 of 17 | **17 of 17** | **+17** |

The paper's RQ-2 data: median **+5.85 pp success rate**, **+15.12% AE** across four benchmarks. The 5 live above are concrete instances of the failure modes that drive those aggregate numbers.

## Empirical reproductions (live MCP, old descriptions)

Five scenarios, each targeting a specific tool and description gap. Numbers in parentheses are wasted tokens per occurrence.

### 1. `advanced_search` — bad example in parameter docs (~110 tokens wasted)

Old parameter hint: *"operation: is, isNot, contains, etc."* The `etc.` implies more options exist. Calling with the natural guess:
```python
conditions=[{"field":"itemType","operation":"is","value":"preprint"},
            {"field":"date","operation":"isInTheLast","value":"30 days"}]
```
→ `Error: Unsupported operation 'isInTheLast' in condition 2. Supported: beginsWith, contains, doesNotContain, endsWith, is, isAfter, isBefore, isGreaterThan, isLessThan, isNot`

[...]

PR #258, opened 2026-04-22
improve Scite and ChatGPT-connector tool descriptions per MCP smell rubric

rbpasker:
## Summary

**Rewrites 5 tool descriptions across Scite enrichment and the ChatGPT custom-connector endpoints — completes the rubric-compliance pass for this server's tools.**

Two commits, one per group. These were the least-smelly tools going in (multi-sentence descriptions already), but live calls still reveal under-documented behavior the old copy hid.

### Headline numbers

| | Old | New | Δ |
|---|---:|---:|---:|
| Tool-intent confusion hazards eliminated | **2** (connector names) | 0 | **−2** |
| Undocumented behaviors surfaced | **4** | 0 | **−4** |
| Tools previously ≤ 20 tokens | **2** (search, fetch) | 0 | **−2** |
| Tools meeting rubric (25/25) | 0 of 5 | **5 of 5** | **+5** |
| Description tokens across 5 tools | **156** | 1,181 | +1,025 |

### 1. `search` / `fetch` — no guidance away from the wrong call site

Old descriptions: *"ChatGPT-compatible search wrapper. Performs semantic search and returns JSON results."* / *"ChatGPT-compatible fetch wrapper. Retrieves fulltext/metadata for a Zotero item by ID."*

The names `search` and `fetch` are required by the OpenAI MCP-over-web spec — but to a non-ChatGPT agent browsing tool names, `search` looks shorter and friendlier than `zotero_semantic_search`. Nothing in the old description warned agents away. New descriptions lead with a *"NAME is REQUIRED by the OpenAI MCP-over-web spec — not intended for general MCP clients; use zotero_semantic_search / zotero_get_item_fulltext instead"* banner.

Live `search(query='MCP tool descriptions')` returned 10 results in the ChatGPT envelope `{"results":[{id,title,url}, ...]}`; live `fetch(id='RTKZQI8E')` returned the full envelope with `{id,title,text,url,metadata:{...}}`. The shape the connector produces is now spelled out explicitly in the descriptions so regular-MCP agents can recognise when they've picked the wrong endpoint.

### 2. All three Scite tools — silent network reliability

Live `scite_check_retractions(limit=5)` returned: `Could not reach Scite API — try again later.` — Scite's public endpoint was transiently down. None of the old descriptions mentioned the third-party-reliability failure mode. New descriptions on all three tools call it out: *"uses the free public endpoints, so calls can fail transiently: expect 'Could not reach Scite API — try again later' when Scite is slow or unreachable (not a permanent error)"*.

### 3. `scite_enrich_item` — either/or semantics and DOI resolution

Old: *"Provide either a DOI or a Zotero item key."* — doesn't say what happens with both, how item_key → DOI resolution works, or what fails if the item has no DOI. Live `scite_enrich_item(doi='10.1162/tacl_a_00638')` returned a full citation tally for the matching paper — but the path the tool took to get there (DOI normalization, fallback on item_key lookup, failure mode when neither has a DOI) was invisible. New description spells out: doi wins when both passed; item_key pulls DOI from Zotero metadata and fails clearly when none exists; prefix/whitespace normalization is automatic.

### 4. `scite_check_retractions` — mutually-exclusive scoping modes

Old: *"Filter by collection, tag, or check recent items."* — doesn't tell an agent what happens when 2+ are passed, or what "recent" means. New description: three scoping modes (collection > tag > recent) are mutually exclusive; first non-null wins; recent defaults to most-recently-MODIFIED ordering; limit max is 500; items without DOIs skipped silently.

## Token footprint

Measured with `tiktoken` (`cl100k_base`):

| Tool | Old | New | Δ |
|---|---:|---:|---:|
| scite_enrich_item | 42 | 248 | +206 |
| scite_enrich_search | 42 | 228 | +186 |
| scite_check_retractions | 39 | 252 | +213 |
| search | **14** | 195 | **+181** |
| fetch | **19** | 258 | **+239** |
| **TOTAL** | **156** | **1,181** | **+1,025** |

Bolded were ≤ 20-token one-liners on the connector tools.

## Commits

| # | Group | Tools |
|---|---|---|
| 1 | Scite | `scite_enrich_item`, `scite_enrich_search`, `scite_check_retractions` (3) |
| 2 | ChatGPT connectors | `search`, `fetch` (2) |

## Rubric scoring

Every rewritten tool scores 25/25:

- **Purpose** — one-sentence "what it does" up front, with the connector pair carrying the explicit OpenAI-spec banner so non-ChatGPT clients route elsewhere
- **Guidelines** — Scite tools cross-reference each other (single → batch, batch → retraction-only), connectors point to their zotero_* replacements for regular MCP use
- **Limitations** — Scite network-reliability failure mode, 500-item cap on check_retractions, items-without-DOIs silently skipped, connector silent-fallback error path, fixed limit=10 on `search`
- **Parameter explanation** — either/or doi/item_key with explicit precedence, mutually-exclusive scoping modes on retractions, id semantics on fetch, url-field fallback rules
- **Examples** — every tool has a concrete inline example

## Token-budget tests

Not included to keep this PR off the dependency chain. When [#256](https://github.com/54yyyu/zotero-mcp/pull/256) merges, these 5 tools should be added to `TOOL_BUDGETS` in `tests/test_description_tokens.py`.

## Test plan

- [x] All 432 existing tests pass
- [x] Live scenarios recorded above
- [ ] Maintainer review

Depends on nothing; clean off `main`. Completes the rubric pass for this server — after [#254](https://github.com/54yyyu/zotero-mcp/pull/254) (collection descriptions), [#256](https://github.com/54yyyu/zotero-mcp/pull/256) (annotations/tags/search), and [#257](https://github.com/54yyyu/zotero-mcp/pull/257) (item retrieval/ingest/dupes/library), every `@mcp.tool` description in this server now hits 25/25.

### DIFF
--- a/src/zotero_mcp/tools/connectors.py
+++ b/src/zotero_mcp/tools/connectors.py
@@ -19,7 +19,26 @@ from zotero_mcp.tools.retrieval import get_item_fulltext
 
 @mcp.tool(
     name="search",
-    description="ChatGPT-compatible search wrapper. Performs semantic search and returns JSON results."
+    description=(
+        "ChatGPT custom connector SEARCH endpoint — name is REQUIRED by "
+        "the MCP-over-web spec (see platform.openai.com/docs/mcp); "
+        "do not rename. Not intended for general MCP clients — in Claude "
+        "or other regular MCP contexts use zotero_semantic_search or "
+        "zotero_search_items instead, which return richer markdown. "
+        "Performs semantic search over the active Zotero library and "
+        "returns a JSON string {\"results\":[{\"id\",\"title\",\"url\"}, "
+        "...]} matching the ChatGPT connector citation UI. URLs are "
+        "zotero://select/items/<key> deep-links. "
+        "query: topic string; natural language works (embedding match). "
+        "No limit parameter — fixed at 10 per the connector UI's "
+        "expected result-set size. "
+        "Requires the semantic search DB populated — run "
+        "zotero_update_search_database first if empty. "
+        "SILENT FALLBACK: any error returns {\"results\":[]} rather "
+        "than raising, to keep the ChatGPT connector stable. "
+        "Example (agent-invoked): search(query='mindfulness-based "
+        "therapy')."
+    )
 )
 def chatgpt_connector_search(
     query: str,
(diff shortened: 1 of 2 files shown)


54yyyu:
thanks for contributing!

PR #259, opened 2026-04-22
fix: surface trash status and list collection keys in item metadata

rbpasker:
## Summary

Two related gaps in the item-metadata formatters caused the MCP to return stale or misleading data after items were trashed or collections deleted:

1. **Trashed items were returned silently.** `format_item_metadata` in [client.py](src/zotero_mcp/client.py) and `format_item_result` in [utils.py](src/zotero_mcp/utils.py) ignored `data.deleted`. The Zotero web API surfaces this flag on every item lookup; we just weren't reading it. An agent reasoning about "current" library state would cite a paper the user had explicitly removed.

2. **Collections rendered as bare counts, not keys.** The line read `**Collections:** N collections` with no keys. The Zotero web API does NOT cascade collection-delete to items' `data.collections` arrays, so the array can contain dangling references to collections that no longer exist; the count is stale.

## Where the bugs live

Both are MCP-layer bugs. The Zotero web API provides everything needed:
- `data.deleted == 1` is documented for trashed items (see [Zotero API types & fields](https://www.zotero.org/support/dev/web_api/v3/types_and_fields))
- `data.collections` is an array of current collection keys

The formatters just didn't read `deleted` and didn't display the keys from `collections`. Fixed in both formatters — one line each.

## Reproduction

```
user (via Zotero desktop): Delete Collection and Items for "MCP Tool Description Eval"
  → item RTKZQI8E moved to Trash, collection 9SU943GB deleted

mcp call: zotero_get_item_metadata(item_key="RTKZQI8E")
  → returns full metadata; NO trash indicator
  → "Collections: 1 collections" (the 1 is 9SU943GB, which no longer exists)
```

After the fix:
```
**Status:** 🗑️ In Trash (recoverable from Zotero Trash view)
...
**Collections:** 9SU943GB
```

The agent can now (a) refuse to cite, and (b) verify the key via `zotero_search_collections` and discover it's dangling.

## Scope

Fixed in both metadata formatters:
- `format_item_metadata` (single-item markdown) — used by `zotero_get_item_metadata`, `zotero_get_item_fulltext` (for the header block)
- `format_item_result` (list-mode) — used by `zotero_search_items`, `zotero_get_collection_items`, `zotero_get_recent`, `scite_enrich_search`, etc.

pyzotero filters trashed items out of list endpoints by default, so the list-mode fix is defense-in-depth against callers that pass `includeTrashed=1` or route single-item lookups through the list formatter.

## Test plan

- [x] 11 new tests in `tests/test_item_metadata_status.py` cover both formatters and both fixes
- [x] Tests fail on unfixed code (verified) and pass with the fix
- [x] Full suite still green: 443 passing
- [x] `zotero_get_item_metadata` description updated to call out trashed-item inclusion and key-vs-count semantics

Independent of all other open PRs; clean off `main`.


rbpasker:
## Merge ordering note (updated)

This PR has been **rebased onto #257**. The `zotero_get_item_metadata` description now combines both edits — #257's rubric-compliant body with this PR's trash/collections addendum appended before the example block. The rebase resolved the conflict ahead of time so no manual resolution is required at merge.

**Current state**: this branch contains #257's 5 commits plus this one. GitHub's PR view will show all of those commits in the diff until #257 merges — at which point GitHub auto-detects the merged commits and the diff collapses to just the single fix commit.

**Suggested order**: merge #257 → GitHub cleans up this PR's diff → merge this.

The underlying code fix (trash status in [client.py](src/zotero_mcp/client.py)/[utils.py](src/zotero_mcp/utils.py) and its 11 tests) is independent of #257 and conflicts nowhere else. If for any reason #257 is not going to merge, this PR can be rebased back onto `main` with a one-line revert of the description-side change.

54yyyu:
hey, this looks good but there are merge conflicts now after recent merges. could you rebase against main when you get a chance?

54yyyu:
thanks for contributing!

PR #260, opened 2026-04-23
feat(semantic): fulltext via web API + since-based incremental sync

SipengXie2024:
## Problem

Users running `zotero-mcp-server` against the Zotero **web API** (e.g. on a headless Linux server with no local Zotero desktop client) currently get **metadata-only semantic search** — the extract_fulltext path is gated behind `is_local_mode()` and raises `RuntimeError` otherwise. But Zotero cloud already keeps the server-side-extracted PDF text for every PDF the desktop client has ever synced: `zot.fulltext_item(key)` returns it. There is a capability gap, not a data gap.

Additionally, every update rescans the whole library from scratch and never deletes items that have been removed in Zotero.

## Summary of changes

1. **Fulltext via web API** — `ZoteroSemanticSearch._fetch_fulltext_via_web_api(item_key)` tries `fulltext_item(key)` on the item itself, then walks PDF attachment children and tries each. Recorded source (`web-api:parent` or `web-api:attachment:<child_key>`) is stored in ChromaDB metadata for diagnostics.
2. **Web-API ingest wiring** — `_get_items_from_api(include_fulltext=False)` gains an optional flag that populates `data.fulltext` in place for each returned item. Downstream `_process_item_batch` already reads `data.fulltext` and writes the existing `has_fulltext` bookkeeping, so no per-item code path needed to change.
3. **New public kwarg `include_fulltext`** on `update_database`. Defaults from a new `semantic_search.include_fulltext` config key (default true, so existing users auto-upgrade). `extract_fulltext=True` (local mode) still wins and uses the local sqlite path — this flag only governs web-API ingest.
4. **Incremental sync** — `_get_changed_items_from_api(since_version, include_fulltext)` uses pyzotero's `item_versions(since=V)` for changes and an unfiltered `item_versions()` for deletion detection. `update_database` caches the library version as `semantic_search.last_sync_version` and fast-paths a noop when `last_modified_version()` hasn't advanced.
5. **Deletion handling** — entries in ChromaDB whose keys no longer appear in the current library are removed via a new `ChromaClient.get_all_ids()` + existing `delete_documents`. New `deleted_items` stat.
6. **Pre-search fire-and-forget** — the `zotero_semantic_search` tool kicks off a background daemon-thread sync when `auto_update` is on, with a 60s module-level debounce. Swallows all exceptions — a failed sync must never surface as a tool error. Future tool calls see fresh state without blocking the current query.
7. **Config default** — `setup_helper.setup_semantic_search` writes `include_fulltext: true` so new configs opt in from day one.

Existing call sites (`_app.py`'s `server_lifespan` and the `update_search_database` tool) were **not** modified. They already pass `extract_fulltext=is_local_mode()`; the new `include_fulltext=None` default resolves from config so behavior upgrades automatically.

## Verification

Ran on a 2402-item real Zotero library over web API.

* **Bootstrap path** — `zotero-mcp update-db --limit 20`: 20 items fully indexed in 18.3s, 10/20 items got fulltext (50% hit rate; misses are notes, webpages, datasets without PDFs). Attachment-child fallback works — e.g. parent `HYJ5IR99` had no fulltext but its PDF child `6HN4C9LI` did, yielding `fulltext_source=web-api:attachment:6HN4C9LI`.
* **last_sync_version persistence** — config after bootstrap: `last_sync_version: 7661` (matches live library).
* **Incremental noop** — second run with no library changes: 3.3s total, 0 items processed, log says \`Library unchanged since last sync (version 7661); no items to reindex.\`.
* **Unit tests** — 20 new tests in \`tests/test_fulltext_web_mode.py\` cover: parent-hit / attachment-fallback / empty-miss / non-PDF-skip for the fulltext fetcher; bootstrap vs incremental vs noop dispatch in `update_database`; deletion detection; explicit `include_fulltext=false` opt-out; force-rebuild watermark advance; config loader defaults.
* **Full suite** — \`uv run pytest tests/\` passes **454/454** (435 baseline + 19 new plus 1 later added for the watermark fix).

## Backward compatibility

* Local-mode users (`ZOTERO_LOCAL=true`) see no behavior change — `extract_fulltext=True` takes precedence and the existing local-sqlite extraction runs as before.
* Web-API users with existing ChromaDB collections get an automatic upgrade on their next sync: fulltext is fetched and items are upserted. No collection reset required.
* Users who prefer metadata-only can set `"include_fulltext": false` in their config.
* The new `get_all_ids()` method on `ChromaClient` is additive, not replacing any existing API.

## Commits

1. `feat(semantic): fulltext via web API + since-based incremental sync`
2. `fix(semantic): advance last_sync_version on force_full_rebuild` — caught in review: the force-rebuild branch was skipping the watermark capture, which would leave `last_sync_version` stale and cause the next incremental run to miss items not modified since the old watermark.

### DIFF
--- a/src/zotero_mcp/chroma_client.py
+++ b/src/zotero_mcp/chroma_client.py
@@ -628,6 +628,19 @@ class ChromaClient:
         except Exception:
             return set()
 
+    def get_all_ids(self) -> set[str]:
+        """Return every id currently stored in the collection.
+
+        Used by incremental sync to compute deletions: items in the local
+        collection but no longer present in the Zotero library.
+        """
+        try:
+            result = self.collection.get(include=[])
+            return set(result.get("ids", []))
+        except Exception as e:
+            logger.error(f"Error listing collection ids: {e}")
+            return set()
+
 
 def create_chroma_client(config_path: str | None = None) -> ChromaClient:
     """
--- a/src/zotero_mcp/semantic_search.py
+++ b/src/zotero_mcp/semantic_search.py
@@ -142,8 +142,46 @@ class ZoteroSemanticSearch:
 
         return config
 
-    def _save_update_config(self) -> None:
-        """Save update configuration to file."""
+    def _load_include_fulltext_setting(self) -> bool:
+        """Whether to fetch fulltext via the Zotero web API during indexing.
+
+        Defaults to True so existing users auto-upgrade to fulltext indexing on
+        their next sync. Users can opt out by setting
+        `semantic_search.include_fulltext: false` in the config file.
+        Local mode (`ZOTERO_LOCAL=true`) keeps using `extract_fulltext` via
+        the local sqlite DB; this setting only governs web-API ingestion.
+        """
+        if not self.config_path or not os.path.exists(self.config_path):
+            return True
+        try:
+            with open(self.config_path) as f:
+                file_config = json.load(f)
+                value = file_config.get("semantic_search", {}).get("include_fulltext", True)
+                return bool(value)
+        except Exception as e:
+            logger.warning(f"Error loading include_fulltext setting: {e}")
+            return True
+
+    def _load_last_sync_version(self) -> int:
+        """Last Zotero library version fully indexed into ChromaDB.
+
+        Zero means "no prior successful sync; bootstrap required". Used to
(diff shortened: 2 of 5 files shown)


54yyyu:
thanks for contributing!

PR #261, opened 2026-04-24
feat(semantic): chunked ingest + groupby rerank + gap-fill resume (builds on #260)

SipengXie2024:
## Relation to #260

This PR is **stacked on top of #260** (fulltext via web API + since-based incremental sync). The diff against `main` therefore includes #260's two commits; reviewers can focus on the **six commits unique to this branch**:

- `a50f28c feat(semantic): chunked ingest + groupby rerank + embedding throttle`
- `acad0cd fix(embeddings): sub-batch OpenAI requests to respect provider caps`
- `13c784b fix(semantic): rebuild when collection reset but last_sync_version stays`
- `75cb347 fix(throttle): enforce rate limit at every HTTP request, including retry`
- `e979704 feat(semantic): diff-driven incremental sync with gap-fill / resume`
- `679f43d fix(ingest): exclude annotation items; self-heal stale annotation chunks`

I'll rebase onto `main` once #260 merges so only those six remain.

## Motivation

Even after #260 lands, a single Zotero item still gets one truncated embedding per upsert (`_process_item_batch` calls `truncate_text` to 8k tokens then writes a single document whose id equals the item key). For a 15k-token conference paper that loses the entire second half of the body. Semantic search can only score "title/abstract + first-chunk relevance", not "does this specific paragraph match my query".

### 1. Chunked ingest
`ZoteroSemanticSearch._chunk_document(text, window=1500, overlap=225)` slices each document into overlapping token windows via tiktoken (cl100k_base — the tokenizer used by OpenAI and most OpenAI-compatible providers). 1500/225 fits comfortably inside `text-embedding-3-small`, `bge-m3`, etc. with room for prepended structured metadata. Falls back to char-based windowing when tiktoken is unavailable; clamps overlap < window to avoid infinite loops on malformed config.

`_process_item_batch` now emits one id per chunk (`<item_key>__<chunk_index>`); each chunk's metadata carries `parent_item_key`, `chunk_index`, `total_chunks`, plus the usual per-item fields. Before upsert, `delete_documents_by_parent(parent)` clears stale chunks so chunk count can shrink across runs when fulltext re-extraction yields less text.

### 2. Dedupe-then-rerank search handler
`search()` oversamples to `max(limit * rerank_multiplier * 5, 50)` hits from ChromaDB, collapses via a new `_dedupe_by_parent` that keeps the best chunk per parent, then runs the optional cross-encoder reranker on the deduped set. `_enrich_search_results` dereferences `parent_item_key` (not the chunk id) when calling `zot.item(key)` so callers never see chunk ids in the wrong place; result dicts gain `chunk_id` and `chunk_index` for diagnostics.

### 3. Sub-batched embedding requests
SiliconFlow caps `/v1/embeddings` at 64 inputs per POST; 25 items × 16 chunks = 400 inputs exceeds that. `OpenAIEmbeddingFunction.__call__` now slices `input` into `request_batch_size` (default 64, config-overridable) windows and concatenates vectors. Return order preserved.

### 4. Rate-limiting at the HTTP request boundary
Throttle moved from per-upsert (in `ZoteroSemanticSearch`) down to per-HTTP-request (in `OpenAIEmbeddingFunction._wait_for_rate_limit`). Two bugs that caused cascade failures on SiliconFlow free-tier rebuilds:

- Sub-batching inside a single upsert fired N back-to-back requests without throttle. Bursts hit TPM.
- The `update_database` retry loop called `chroma_client.upsert_documents([doc], ...)` directly, bypassing the throttle entirely. A rate-limit failure cascade dumped thousands of small requests into the retry pool unthrottled and logged permanent failures.

Both paths now call `_throttle_embedding_request` / `_wait_for_rate_limit` before every HTTP request. Config knob: `embedding_config.rate_limit_rps` (or the outer `semantic_search.embedding_rate_limit_rps` — the outer propagates into the embedding function at `create_chroma_client` time).

### 5. ID-format migration + empty-collection resume
When the chunked format is deployed on top of #260's single-embedding format, `update_database` samples existing ids on every run; if none contain `__`, it auto-triggers a `force_full_rebuild` so users don't have to invoke it manually.

### 6. Diff-driven incremental sync with gap-fill
The old fast-path noop in the incremental branch assumed ChromaDB was a consistent snapshot of the library at `last_sync_version`. That broke when:
- A rebuild was killed mid-run (watermark not advanced; ChromaDB partial)
- The embedding model changed (ChromaClient silently resets the collection; watermark still points at the old sync)
- Any crash left the collection inconsistent with the watermark

Users stuck with partial indexes had only `--force-rebuild` available, which throws away all prior progress.

Replaced with a diff-driven design: every run computes `changed ∪ missing` (ingest set) and `deleted`, where `missing` = items in the library but absent from ChromaDB (via a parent-key-aware scan of `get_all_ids()`). If all three sets are empty → true noop with a "Library fully synced" log; otherwise ingest runs. New helper `_fetch_items_by_keys(keys)` bulk-fetches via Zotero's `itemKey=K1,K2,...` query param, batched at 50 per request (Zotero's documented cap), with per-key fallback on batch failure. New stat `gap_filled_items`.

### 7. Annotation item self-heal
The original `_get_items_from_api` filter excluded only `itemType in {attachment, note}`, letting `annotation` items (PDF highlights and user comments, which Zotero exposes as top-level in the /items listing) slip through. On a representative 2402-item library, this produced **1018 zero-text annotation chunks** drowning out the 237 real research papers in dedup rankings. Tightened the filter to match the since-based and itemKey paths (all three now drop `annotation`) and added `delete_documents_by_item_type("annotation")` as a self-healing cleanup step on every `update_database` run. New stat `cleaned_annotation_chunks`.

## Verification

End-to-end on a 2402-item real Zotero library with web-API mode and `embedding_config` pointing at SiliconFlow bge-m3:

- **First pass (killed mid-run at ~17%)** to stress-test gap-fill resume. `update-db` restart with no `--force-rebuild`: detected 1631 missing items, fetched + indexed, ran to completion.
- **Second pass** triggered annotation cleanup (1018 chunks purged) plus true-noop report: `Library fully synced (version 7661); nothing to reindex`. Duration 1:28 with zero embedding API calls.
- **Final state**: 3952 chunks across **311 distinct parents**; of 237 research-paper parents, 100% have fulltext indexed. Remaining 74 non-fulltext parents are webpages / computer programs / blog posts / bookSections (genuinely have no PDF on Zotero cloud) or journal/conference papers whose PDFs haven't been opened in the user's desktop client yet (out of scope for this PR).
- **Paragraph-level semantic check** ("concurrent smart contract execution parallelism"): top-5 all highly relevant papers (cosine-sim 0.42–0.50 via bge-m3), different chunk indices per paper (`__0`, `__1`, `__2`...) confirming matches land in fulltext bodies, not just title/abstract.
- **Unit tests**: 486 passed. 23 new cases in `tests/test_chunking.py` + 12 new cases in `tests/test_fulltext_web_mode.py` cover chunking boundary behavior, groupby rerank, enrichment parent-key resolution, sub-batching, throttle, legacy-format migration, gap-fill resume, true-noop path, annotation filter + cleanup.

## Config surface added

- `semantic_search.chunking: {window: 1500, overlap: 225}` — tunable chunker parameters
- `semantic_search.embedding_rate_limit_rps: N` — per-HTTP-request throttle (defaults off)
- `semantic_search.embedding_config.request_batch_size: N` — per-provider inputs-per-POST cap (defaults 64, safe for SiliconFlow; real OpenAI can bump to 2048)

`setup_helper.setup_semantic_search` now writes the `chunking` and `embedding_rate_limit_rps` defaults into new config files so users discover and tune them.

## Backward compatibility

- Local-mode users (`ZOTERO_LOCAL=true`) see no behavior change — `extract_fulltext` still takes precedence and the local sqlite path is unaltered.
- Web-API users on PR #260's single-embedding format auto-upgrade on their next `update-db` run via the legacy-id detection → force_full_rebuild path.
- The `update_search_database` MCP tool and `zotero-mcp update-db` CLI signatures are unchanged — all new behavior is default-on but gated behind existing config.

### DIFF
--- a/src/zotero_mcp/chroma_client.py
+++ b/src/zotero_mcp/chroma_client.py
@@ -42,11 +42,16 @@ class OpenAIEmbeddingFunction(EmbeddingFunction):
                  model_name: str = "text-embedding-3-small",
                  api_key: str | None = None,
                  base_url: str | None = None,
-                 request_batch_size: int | None = None):
+                 request_batch_size: int | None = None,
+                 rate_limit_rps: float | None = None):
+        import threading as _threading
         self.model_name = model_name
         self.api_key = api_key or os.getenv("OPENAI_API_KEY")
         self.base_url = base_url or os.getenv("OPENAI_BASE_URL")
         self.request_batch_size = int(request_batch_size) if request_batch_size else self.DEFAULT_REQUEST_BATCH_SIZE
+        self.rate_limit_rps: float | None = float(rate_limit_rps) if rate_limit_rps else None
+        self._rate_lock = _threading.Lock()
+        self._last_request_ts: float = 0.0
         if not self.api_key:
             raise ValueError("OpenAI API key is required")
 
@@ -68,6 +73,7 @@ class OpenAIEmbeddingFunction(EmbeddingFunction):
             "model_name": self.model_name,
             "base_url": self.base_url,
             "request_batch_size": self.request_batch_size,
+            "rate_limit_rps": self.rate_limit_rps,
         }
 
     @staticmethod
@@ -77,10 +83,32 @@ class OpenAIEmbeddingFunction(EmbeddingFunction):
             api_key=config.get("api_key"),
             base_url=config.get("base_url"),
             request_batch_size=config.get("request_batch_size"),
+            rate_limit_rps=config.get("rate_limit_rps"),
         )
 
+    def _wait_for_rate_limit(self) -> None:
+        """Sleep as needed so successive HTTP requests stay under `rate_limit_rps`.
+
+        Applied at the individual HTTP request level (including each
+        sub-batch) so rate-limited providers see a steady request cadence
+        regardless of how many inputs the caller passed. The lock keeps
+        parallel threads (e.g. the pre-search fire-and-forget sync)
+        honest.
+        """
+        rps = self.rate_limit_rps
+        if not rps or rps <= 0:
+            return
+        import time as _t
+        with self._rate_lock:
+            now = _t.monotonic()
+            min_interval = 1.0 / rps
+            wait = min_interval - (now - self._last_request_ts)
+            if wait > 0:
+                _t.sleep(wait)
+            self._last_request_ts = _t.monotonic()
+
     def __call__(self, input: Documents) -> Embeddings:
-        """Generate embeddings, sub-batching when needed.
+        """Generate embeddings, sub-batching + rate-limiting each request.
 
         Upstream callers (e.g. `_process_item_batch`) may pass more items
         than the provider accepts in a single POST. We split into
@@ -91,6 +119,7 @@ class OpenAIEmbeddingFunction(EmbeddingFunction):
             return []
         batch_size = self.request_batch_size or self.DEFAULT_REQUEST_BATCH_SIZE
         if len(input) <= batch_size:
+            self._wait_for_rate_limit()
             response = self.client.embeddings.create(
                 model=self.model_name,
                 input=input,
@@ -99,6 +128,7 @@ class OpenAIEmbeddingFunction(EmbeddingFunction):
         vecs: list = []
         for i in range(0, len(input), batch_size):
             sub = input[i:i + batch_size]
+            self._wait_for_rate_limit()
             response = self.client.embeddings.create(
                 model=self.model_name,
                 input=sub,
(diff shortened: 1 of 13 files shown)


54yyyu:
hey, #260 was just merged and this PR has some merge conflicts now. since this builds on #260, could you rebase against main when you get a chance?

54yyyu:
This is high-value work — the per-chunk embedding directly fixes the long-document truncation problem (#290), and the dedupe-then-rerank + provider sub-batching are needed for OpenAI-compatible backends. It now conflicts only because it was stacked on #260, which has since merged. Could you rebase onto current `main`? The diff should collapse to just the chunking/throttle commits, and I'll do a focused re-review of the chunk-id handling in `_enrich_search_results` and the throttle/retry path. Thanks!

54yyyu:
Thanks for the great work here, @SipengXie2024! The ingest engine in this PR is the part we want to keep: diff-driven incremental sync, gap-fill resume, embedding throttling and sub-batching, and the annotation self-heal.

We landed #350's passage chunking as the canonical schema (the `#`-style ids with char and page offsets), since the page-level provenance is what the citation and grounding work builds on. So rather than have you redo all of it, I'm going to rebase your engine commits onto the new main and drop the token-based `__` chunking layer, keeping your incremental-sync and throttle work intact. The one real overlap is `OpenAIEmbeddingFunction.__call__`, where I'll keep your batching and throttle loop and add `encoding_format="float"` from #348. I'll push it as a follow-up branch and tag you for review!

54yyyu:
Landed the portable part of this in #356 (now on main): the OpenAI embedding sub-batching and per-request rate-limit throttle, configurable via `embedding_config.request_batch_size` and `rate_limit_rps`. Thanks @SipengXie2024!

We took a different route on the rest: passage chunking is now #350's schema (the `#`-style ids with char/page provenance), the incremental-sync and gap-fill landed earlier on main, and we kept embedding annotation text rather than excluding it. So I'm closing this since its unique value is in now. Really appreciate the work here!

PR #263, opened 2026-04-25
feat: add standalone zotero-cli entry point

bkmzhmtd:
## Summary

Adds `zotero-cli` as a second script alongside `zotero-mcp`. It exposes the full Zotero toolset directly from the terminal — no AI assistant or MCP client required.

**Motivation:** MCP tool schemas consume thousands of tokens per context. For agentic pipelines that already have shell access (e.g. Claude Code, computer-use agents), a CLI entry point gives access to the same Zotero functionality at a fraction of the token cost. It is also useful for humans writing shell scripts, cron jobs, and one-off queries.

## Changes

**`src/zotero_mcp/cli_standalone.py`** (new) — standalone CLI module. Commands: `search` (`s`), `get` (`g`), `annotations` (`ann`), `notes` (`n`), `add`, `collections` (`coll`), `tags`, `edit`, `duplicates`, `db`, `library`, `outline`, `config`. Short aliases for interactive use; `-v` flag for verbose progress output; reads from stdin when `--text -` is passed to `notes create`.

**`src/zotero_mcp/_context.py`** (new) — imports `fastmcp.Context` with a lightweight no-op fallback when fastmcp is not installed. This lets tool modules be imported in CLI mode without the full MCP stack, without any conditional logic in the tools themselves.

**`src/zotero_mcp/__init__.py`** — wraps `from .server import mcp` in `try/except ImportError` so the package is importable in environments where fastmcp is not installed.

**`src/zotero_mcp/tools/*.py`** — all six tool modules now import `Context` from `zotero_mcp._context` instead of directly from `fastmcp`.

**`src/zotero_mcp/setup_helper.py`** — `find_claude_config()` gains a `verbose: bool = False` parameter so it doesn't print to stdout when called programmatically from `cli_standalone`.

**`pyproject.toml`** — registers `zotero-cli = "zotero_mcp.cli_standalone:main"`.

**`README.md`** — new *CLI Mode* section with a quick-reference command table and a CLI-vs-MCP comparison.

## Test plan

- [x] `pytest tests/test_cli_standalone.py` — 34 tests pass (CLIContext, parser aliases, `main()` dispatch, `cmd_search`, `cmd_notes`, `cmd_edit`)
- [x] Full suite — no regressions in tests that don't require fastmcp
- [ ] `zotero-cli --help` smoke test after install
- [ ] `zotero-cli search "test"` with local Zotero running

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,11 @@
 - Web API for cloud library access
 - Hybrid mode: read from local Zotero, write via web API
 
+### ⌨️ Standalone CLI (`zotero-cli`)
+- Search, browse, and edit your library directly from the terminal — no AI assistant required
+- Ideal for scripting, automation, and quick lookups
+- Short aliases (`s`, `g`, `ann`, `coll`) for interactive use
+
 ## 🚀 Quick Install
 
 > **New to the command line?** Try the community-built [Zotero MCP Setup](https://github.com/ehawkin/zotero-mcp-setup) — includes a macOS GUI installer (DMG), one-click install scripts for Mac/Windows, and a step-by-step guide. No Terminal experience needed.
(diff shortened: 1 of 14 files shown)


danmackinlay:
@bkmzhmtd There is a sense in which we might thing of a CLI as a "separate" project. What do others think?
FWIW I would be happy to have the CLI included in this MCP

rbpasker:
There will likely be other integrations besides MCP and CLI, so it would be good if they all had the same functionality and semantics, just different ways of invoking

danmackinlay:
right, I agree with that. maybe I feel like we have a naming problem, in that case. `zotero-mcp` would then be too narrow.

rbpasker:
@danmackinlay agreed.

54yyyu:
Thank you guys for all of the feedback! I've already started working on a project that will be more generalized than a mcp (and cli) - please stay tuned! 

This repo will continue serve as a mcp for those who only want it as a connector. :)

titaniumbones:
as a user, I'd love to have independent CLI entrypoints.  At present I use my own homegrown scripts precisely because mcp is so expensive, but zotero-mcp's featureset is richer and better-maintained, so I would love to switch. I'm sure that's true for many others as well.  I hope we can keep a single ecosystem of interrelated tools rather than a fracture between zotero-mcp and whatever the new project is.

54yyyu:
thanks for contributing!

PR #268, opened 2026-04-30
feat: add zotero_get_attachment_path tool

ajdavis:
## Summary

- Adds a new MCP tool `zotero_get_attachment_path(item_key)` that returns the local filesystem path(s) of a Zotero item's attachments. Local mode only.
- Exposes a public `LocalZoteroReader.get_attachment_paths(parent_key)` wrapping the existing `_iter_parent_attachments` + `_resolve_attachment_path` helpers.
- Factors the repeated "open `~/.config/zotero-mcp/config.json` + parse + guard" pattern into `_load_zotero_mcp_config()` in `tools/_helpers.py`, used by both `get_item_fulltext` and the new tool.

## Why

`zotero_get_item_fulltext` is page-capped (default 10 pages), which is fine for papers but truncates books. With this tool, a caller can resolve the attachment's filesystem path and read large PDFs directly (e.g. via Claude Code's PDF reader, which handles 20-page chunks).

Returns all attachments for the item with key, content type, raw Zotero path, resolved local path, and an existence flag --- so books with multiple files (e.g. a snapshot + the PDF) are visible.

## Test plan

- [x] New unit tests in `tests/test_local_db.py::TestGetAttachmentPaths` cover resolved paths, missing files, unknown items, and multiple attachments.
- [x] Existing `tests/test_local_db.py`, `tests/test_fulltext_local_mode.py`, `tests/test_pdf_outline.py`, `tests/test_shared_helpers.py` all pass (77 passed, 1 skipped).
- [x] Smoke-tested live against a local Zotero DB: returns `/Users/.../Zotero/storage/<KEY>/<filename>` for a real book item.

### DIFF
--- a/src/zotero_mcp/local_db.py
+++ b/src/zotero_mcp/local_db.py
@@ -627,6 +627,28 @@ class LocalZoteroReader:
     def extract_fulltext_for_item(self, item_id: int) -> tuple[str, str] | None:
         return self._extract_fulltext_for_item(item_id)
 
+    def get_attachment_paths(self, parent_key: str) -> list[dict]:
+        """Return resolved filesystem paths for a parent item's attachments.
+
+        Each entry has: ``key`` (attachment key), ``content_type``, ``zotero_path``
+        (the raw stored path like ``storage:foo.pdf``), ``resolved_path`` (a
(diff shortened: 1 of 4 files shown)


54yyyu:
thanks for contributing!

PR #275, opened 2026-05-06
fix(arxiv): honor attach_mode='linked_url' in _add_by_arxiv

### LINKED ISSUE #274: _add_by_arxiv ignores attach_mode, breaking WebDAV file sync users
## Problem

When `add_by_url` receives an arXiv URL, it dispatches to the internal `_add_by_arxiv` helper (`src/zotero_mcp/tools/write.py` ~L571). This helper does not accept the `attach_mode` argument that `add_by_url` already takes, and the PDF attachment step is hardcoded to `write_zot.attachment_both()` (~L693), which uploads the PDF binary to Zotero's official cloud storage via the Web API.

## Impact

For users who sync Zotero attachments via **WebDAV** (third-party storage such as Box, Nutstore, or any custom WebDAV — i.e. anything other than Zotero's official paid storage):

- Item metadata is created correctly through the Zotero Web API
- The PDF binary is uploaded to zotero.org's official storage path
- The local Zotero client is configured to pull files from WebDAV, so it never sees the binary
- Result: the attachment shows up in the library but appears empty, the user thinks "the PDF wasn't added"

The DOI path (`add_by_doi` → `_attach_pdf_with_fallback`) already respects `attach_mode` correctly. The arXiv path simply does not forward the parameter.

## Reproduce

```python
zotero_add_by_url(
    url="https://arxiv.org/abs/2401.00001",
    attach_mode="linked_url",   # expects: linked-URL attachment, no upload
)
```

Actual: PDF is downloaded and uploaded to zotero.org; the resulting attachment has `linkMode=imported_file`.

Expected: a `linkMode=linked_url` attachment pointing to the arXiv PDF URL; no binary upload.

## Environment

- zotero-mcp-server `0.3.0` (current `main`)
- Python 3.10 / 3.11
- File sync: WebDAV

A fix with regression tests is up at #275.

chengzhag:
## Problem

`_add_by_arxiv` (called by `add_by_url` for arXiv URLs) previously ignored the `attach_mode` argument and unconditionally used `write_zot.attachment_both()`, uploading PDF binaries to Zotero's official cloud storage via the Web API.

This breaks **WebDAV**-based file sync users (third-party storage like Box, Nutstore, custom WebDAV — anything other than Zotero's official paid storage): the metadata is created, but the binary lives in a storage path the local client never pulls from, so the PDF is invisible to the user.

The DOI path (`_attach_pdf_with_fallback`) already respects `attach_mode` correctly — this PR brings the arXiv path in line.

Fixes #274

## Changes

- Add `attach_mode` parameter to `_add_by_arxiv` (default `"auto"`, preserving existing behavior).
- Forward `attach_mode` from `add_by_url` to `_add_by_arxiv`.
- When `attach_mode == "linked_url"`, route to `_helpers._attach_pdf_linked_url` to create a linked-URL attachment pointing to the arXiv PDF URL — no binary download or upload.

Backward compatible — callers that don't pass `attach_mode` keep the same upload behavior as before.

## Test plan

- [x] Added `TestArxivAttachMode` to `tests/test_add_by_url.py` with two cases that mock `_attach_pdf_linked_url` and `attachment_both` to verify routing in both `linked_url` and default modes.
- [x] `pytest tests/` — 437 passed locally.
- [x] `ruff check` — no new warnings introduced.
- [x] End-to-end: added an arXiv item with `attach_mode="linked_url"` against a real Zotero account, confirmed attachment `linkMode=linked_url` and zero zotero.org storage usage delta.

### DIFF
--- a/src/zotero_mcp/tools/write.py
+++ b/src/zotero_mcp/tools/write.py
@@ -568,7 +568,8 @@ def add_by_url(
         # arXiv URL routing
         arxiv_id = _helpers._normalize_arxiv_id(url)
         if arxiv_id:
-            return _add_by_arxiv(arxiv_id, collections, tags, write_zot, ctx)
+            return _add_by_arxiv(arxiv_id, collections, tags, write_zot, ctx,
+                                 attach_mode=attach_mode)
 
         # Generic webpage
         ctx.info(f"Creating webpage item for: {url}")
@@ -599,7 +600,7 @@ def add_by_url(
         return f"Error adding by URL: {e}"
 
 
-def _add_by_arxiv(arxiv_id, collections, tags, write_zot, ctx):
+def _add_by_arxiv(arxiv_id, collections, tags, write_zot, ctx, attach_mode="auto"):
     """Add an arXiv paper by ID. Internal helper for add_by_url."""
     ctx.info(f"Fetching arXiv metadata for: {arxiv_id}")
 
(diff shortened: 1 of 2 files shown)


chengzhag:
Pushed 4cb5ac4 to address the Copilot review comment on `tests/test_add_by_url.py`.

The previous test only asserted that `attachment_both` was not invoked. The PR description claims `linked_url` mode performs no binary download either, but a regression that still fetched the PDF URL via `requests.get` (without uploading) would have slipped through. The strengthened test now also asserts:

- `requests.get` is invoked exactly once
- the single call targets `export.arxiv.org` (the metadata endpoint)
- no call contains `arxiv.org/pdf/` in its URL argument

This enforces the full "no binary download or upload" contract.

The other Copilot suggestion (reusing `_helpers._download_and_attach_pdf` in the binary-attach branch) is intentionally not addressed in this PR — the duplication predates this change, and switching to that helper would introduce additional behavior (Content-Type / minimum-size checks) that's outside the scope of this `attach_mode` bug fix. Happy to do it as a separate refactor PR if you'd prefer.

54yyyu:
Thanks for contributing to zotero-mcp! Great fix for WebDAV users.

PR #279, opened 2026-05-11
fix: upload attachments to WebDAV in zotero_add_from_file

### LINKED ISSUE #278: zotero_add_from_file silently fails for WebDAV-storage accounts
## Summary

When the running Zotero account uses **WebDAV file storage** (rather than Zotero's hosted S3 storage), `zotero_add_from_file` reports `File attached: <name>` to the caller but the file bytes never reach the user's WebDAV. The attachment item is created in Zotero with `md5` / `mtime` / `filename` set, but the actual `<KEY>.zip` and `<KEY>.prop` files are never written to the user's WebDAV server.

The result for the user:

- `zotero_get_item_fulltext` → `Attachment <KEY> was not found in WebDAV storage` (after #249).
- Opening the attachment in Zotero Desktop → `File Not Found` pointing at `~/Zotero/storage/<KEY>/<file>`.
- The `<KEY>.zip` blob is genuinely absent from the WebDAV server (verifiable with `curl PROPFIND`).

## Root cause

`zotero_add_from_file` calls `pyzotero.Zotero.attachment_both(...)` which delegates to `Zupload.upload()`. `Zupload`'s file-upload pipeline targets Zotero's `/items/<KEY>/file` endpoint, which authorizes uploads against **Zotero Storage (S3)** only — not WebDAV. For WebDAV-storage accounts, either:

1. the `_get_auth` response says `{"exists": 1}` and pyzotero short-circuits (no bytes uploaded anywhere), or
2. the `_upload_file` step succeeds against an S3 URL the user's library doesn't read from.

In both cases no exception is raised, so `zotero_add_from_file` proceeds to its happy path and reports success. The attachment item exists, but the bytes are inaccessible to every Zotero client the user actually uses.

## Reproduce

1. Configure an account that uses WebDAV for file sync (Preferences → Sync → File Syncing → WebDAV).
2. Set `ZOTERO_WEBDAV_URL` / `_USERNAME` / `_PASSWORD` for the MCP server.
3. Call `zotero_add_from_file(file_path='/tmp/some.pdf', collections=[...])`.
4. Call `zotero_get_item_fulltext` on the returned item key → `Attachment <K> was not found in WebDAV storage`.
5. `curl -u $USR:$PSW -X PROPFIND -H 'Depth: 1' "$URL" | grep '<KEY>.zip'` → empty.

## Proposed fix

When `webdav.is_webdav_configured()` is true (i.e. after #249 lands), perform the upload directly via WebDAV PUT in addition to the pyzotero metadata write:

- `PUT <base>/<KEY>.zip` — file content wrapped in a single-member zip under its basename
- `PUT <base>/<KEY>.prop` — XML sidecar `<properties version="1"><mtime>…</mtime><hash>…</hash></properties>`
- Order: `.zip` first; Zotero treats an orphan `.prop` as corruption.
- The md5/mtime in the `.prop` must match what pyzotero wrote to the attachment item metadata.

Draft PR (depends on #249): linked above.

## Notes

- The maintenance burden is small; the fix is ~50 lines in `webdav.py` plus a ~20-line patch to `zotero_add_from_file`.
- The PR keeps pyzotero's `attachment_both` call so the attachment item metadata still lands; the WebDAV PUT is purely additive when WebDAV is configured.
- Existing Zotero-Storage workflows are unaffected — the new path is gated on `is_webdav_configured()`.

rbpasker:
**Depends on #249** — this PR is stacked on the WebDAV attachment fallback branch. Once #249 lands, rebase against `main`.

Fixes #278.

## Summary

When `ZOTERO_WEBDAV_*` env vars are configured, `zotero_add_from_file` now performs a direct WebDAV PUT of the file content (in addition to pyzotero's metadata write) so the attachment is actually retrievable. Without this, the call reports "File attached" but the bytes never reach the user's WebDAV — see #278 for the failure mode.

## Changes

- `webdav.py`: add `upload_attachment_to_webdav(key, file_path)` that writes `<KEY>.zip` (file under basename) and `<KEY>.prop` (XML sidecar with `mtime` + `hash`) via WebDAV PUT, in that order. Computes md5 in-stream; returns `(md5_hex, mtime_ms)` so callers can keep Zotero metadata consistent.
- `tools/write.py`: after `attachment_both` returns, if `is_webdav_configured()`, pull the new attachment key out of pyzotero's response shape and call the new upload helper. WebDAV failures surface in the tool output as a warning rather than aborting the call — the attachment item already exists in Zotero and the upload can be retried by the user separately.
- `tests/test_webdav.py`: 4 new tests covering PUT ordering, zip contents (basename, not full path), URL-escaping of attachment keys, and the `WebDAVNotConfiguredError` branch.

## Test plan

- [x] Unit tests: `pytest tests/test_webdav.py -v` — 8/8 passing (4 existing + 4 new)
- [x] Full suite: `pytest -q` — 445/445 passing on this branch
- [ ] Manual end-to-end on a real WebDAV account (Fastmail in my case): pending a real reproduction once #249 lands at upstream

## Behavior matrix

| Storage backend | `is_webdav_configured()` | Behavior |
|---|---|---|
| Zotero Storage (cloud) | False | unchanged — pyzotero's `attachment_both` handles upload as before |
| WebDAV-only | True | pyzotero writes the attachment item metadata; new WebDAV PUT writes `<KEY>.zip` + `<KEY>.prop` |
| WebDAV configured but upload fails | True | attachment item exists; tool output includes a `WARNING: WebDAV upload failed — <reason>` notice; user can retry |

## Notes for review

- The `.prop` XML is minimal (`mtime` + `hash`). Zotero accepts this; richer property files (compression flag, etc.) aren't required for sync.
- md5 is computed by streaming the file (8KB chunks) — same approach pyzotero uses in `_get_auth`.
- The `.zip` stores the file under its basename (not full path), matching what `_extract_archive` in #249 expects on the download side.
- No new dependencies — `zipfile` and `hashlib` are stdlib.

### DIFF
--- a/README.md
+++ b/README.md
@@ -281,6 +281,9 @@ zotero-mcp setup --no-local --api-key YOUR_API_KEY --library-id YOUR_LIBRARY_ID
 - `ZOTERO_API_KEY`: Your Zotero API key (for web API)
 - `ZOTERO_LIBRARY_ID`: Your Zotero library ID (for web API)
 - `ZOTERO_LIBRARY_TYPE`: The type of library (user or group, default: user)
+- `ZOTERO_WEBDAV_URL`: Optional WebDAV folder URL for direct attachment downloads in remote mode
+- `ZOTERO_WEBDAV_USERNAME`: Optional WebDAV username
+- `ZOTERO_WEBDAV_PASSWORD`: Optional WebDAV password
 
 **Semantic Search:**
 - `ZOTERO_EMBEDDING_MODEL`: Embedding model to use (default, openai, gemini)
@@ -354,7 +357,7 @@ The first time you use PDF annotation features, the necessary tools will be auto
 - `zotero_search_by_tag`: Search your library using custom tag filters
 
 ### 📚 Content Tools
-- `zotero_get_item_metadata`: Get detailed metadata (supports BibTeX export via `format="bibtex"`)
+- `zotero_get_item_metadata`: Get detailed metadata (supports `format="markdown"`, `format="json"` for complete raw Zotero metadata, and `format="bibtex"`)
 - `zotero_get_item_fulltext`: Get full text content
 - `zotero_get_item_children`: Get attachments and notes
 
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -185,7 +185,7 @@ zotero-mcp serve --transport sse --host localhost --port 8000
 When connected to Claude Desktop or another MCP client, you'll have access to these tools:
 
 - **zotero_search_items**: Search your library by title, creator, or content
-- **zotero_get_item_metadata**: Get detailed information about a specific item
+- **zotero_get_item_metadata**: Get detailed information about a specific item, including complete raw metadata via `format="json"`
 - **zotero_get_item_fulltext**: Get the full text content of an item
 - **zotero_get_collections**: List all collections in your library
 - **zotero_get_collection_items**: Get all items in a specific collection
--- a/src/zotero_mcp/cli.py
+++ b/src/zotero_mcp/cli.py
@@ -31,7 +31,18 @@ def obfuscate_config_for_display(config):
         return config
 
     obfuscated = config.copy()
-    sensitive_keys = ["ZOTERO_API_KEY", "ZOTERO_LIBRARY_ID", "API_KEY", "LIBRARY_ID"]
+    sensitive_keys = [
+        "ZOTERO_API_KEY",
+        "ZOTERO_LIBRARY_ID",
+        "ZOTERO_WEBDAV_URL",
+        "ZOTERO_WEBDAV_USERNAME",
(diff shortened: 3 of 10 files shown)


rbpasker:
Live-verified end-to-end on a real Fastmail-backed Zotero library — the PR's manual-test checkbox now passes. Details posted on #278.

Short version: `zotero_add_from_file` reports `(uploaded to WebDAV as <KEY>.zip)`, a direct WebDAV `PROPFIND` confirms both `<KEY>.zip` and `<KEY>.prop` exist, and after a Zotero Desktop sync the file lands in local storage and `zotero_get_item_fulltext` returns 121K characters of extracted PDF text — i.e. the WebDAV write-side and the existing read cascade close the loop.

54yyyu:
thanks for contributing!

PR #280, opened 2026-05-11
Paper fetching via proxy with browser cookie auth

juan-cobos:
Hi, first of all, this is a really cool project! 🙂

While integrating LLMs into my Zotero workflow, I wanted to give them direct access to papers referenced by URLs. This allows the model to retrieve content such as equations, figures, and text directly from the source, without relying on PDFs or similar.

I put together a working example that retrieves papers requiring authentication by reusing browser cookies and rewriting URLs through a proxy scheme, similar to how the Zotero Connector handles access.

This is a fairly naive proof of concept that works well for my specific use case, and there are probably cleaner or more robust ways to implement it.

I’d love to hear your thoughts!


54yyyu:
Thanks for the proof of concept — useful capability. Holding for a few things before it can ship in a published package:
1. **Remove the hardcoded `insb.bib.cnrs.fr` default** — that's an institution-specific proxy; it should have no default (error if unconfigured).
2. **DOI claim**: the tool description says it 'resolves DOIs automatically' but there's no DOI-resolution code — either implement it or drop the claim (a bare DOI currently rewrites into a malformed host).
3. **Robustness**: add a request timeout and a response size cap to `fetch()` (currently an unbounded `session.get`), and add tests.
4. **Security/consent**: reading the live browser cookie jar (incl. OS-keychain-decrypted Chrome/Edge cookies) is a meaningful new surface — please add a user-facing consent/security note in the docs.

It's correctly opt-in (separate `proxy` extra, lazy import, dormant without config), so once the above are addressed I'm happy to take another look.

54yyyu (changes requested):
Thanks for putting this together! A few things to address before this can merge:

- It reads cookies from the user's browser and forwards them, so this should be clearly opt-in and documented as such, with the proxy domain required in config rather than defaulted.
- The hardcoded default proxy domain (`insb.bib.cnrs.fr`) should come out.
- There's no timeout on the request, so it can hang indefinitely. Please add one.
- The tool description says it resolves DOIs, but the code doesn't do that yet, so either add resolution or update the description.
- It needs tests for the new fetch and cookie-handling paths.

Happy to take another look once those are in!

juan-cobos:
@54yyyu   Thanks for the thorough feedback! I've addressed the points in a minimal, focused way.

  There's still room to expand the README into a fuller consent/security note and to add proper DOI resolution later, happy to follow up with those in a separate pass.

  Looking forward to your thoughts!
Juan

54yyyu:
Thanks, and sorry for the long silence. Two things keep me from merging this as-is: pulling auth cookies out of local browser profiles is a sensitive capability to ship enabled in a general-purpose MCP server, and the proxy flow adds a fairly large surface that only some institutional users need. If you're up for it, I'd consider a slimmed version: proxy URL rewriting as an opt-in config, with cookie sourcing left to an explicit user-supplied cookie file rather than reading browser profiles. Otherwise I'll likely close this in a future cleanup, with credit, and link it from an issue for institutional-access support.

54yyyu:
Thanks for this, and sorry for the slow reply.

The use case is real: getting at the HTML source rather than the PDF gives the model equations, tables and structure that PDF extraction mangles, and #391 and #378 are circling the same problem from different directions.

I don't want to take the cookie-reading part into core, though. Reading the user's browser cookie jar means the MCP server holds credentials for every site they're logged into, not just their institutional proxy, and it's reachable by whatever model is driving the tools. That's a meaningful escalation in what a compromised or merely over-eager agent can do with this server, and it isn't something I can put behind a config flag and call handled. The EZproxy URL rewriting on its own is fine; it's the credential source that I'm not willing to own.

Two directions that would work:

- A companion package that owns the cookie access and exposes its own MCP tool. Users who want it install it deliberately, and the trust decision stays theirs. Happy to link it from the README.
- Or, in core, a fetch path that takes an explicitly configured proxy plus credentials the user supplies directly, with no browser cookie access at all. Narrower, but it covers the institutional-access case for most people.

Closing this PR since the current shape isn't one I'll merge, but genuinely interested if you want to take either direction. Thanks for writing it up and for the clear framing about it being a proof of concept.

PR #281, opened 2026-05-11
feat: add `place` field to zotero_update_item

rbpasker:
## Summary

Add `place` (publication place) to the set of writable fields on `zotero_update_item`. Today `zotero_update_item` exposes title, creators, date, publisher, publication_title, volume, issue, pages, DOI, ISBN, ISSN, url, language, abstract, short_title, edition, book_title, extra, item_type, and tag/collection helpers — but not `place`, which is a standard Zotero top-level field on `book`, `bookSection`, `thesis`, `manuscript`, `report`, `conferencePaper`, and several other parent item types. As a result, a common book/chapter metadata fix ("publisher city was concatenated into publisher, move it to place") can't currently be done via the MCP and has to be patched up in Zotero Desktop by hand.

## Changes

- `tools/write.py`:
  - Add `place: str | None = None` to the `update_item` signature parallel to `publisher` / `edition`.
  - Add `"place": "place"` to `_UPDATE_ITEM_API_TO_PARAM` so the skip-warning message uses the snake_case param name. (Zotero's API key is already `place` lowercase, so the mapping is identity — but the registration keeps the warning machinery consistent.)
  - Route `place` through the existing `field_updates` if-chain.
- `tests/test_update_item.py`:
  - `test_update_place_on_book` — round-trip on a book fixture.
  - `test_update_place_on_book_section` — same for bookSection.
  - `test_update_place_skipped_on_journal_article` — passing `place=` on a journalArticle (which doesn't carry the field) produces the same skip warning the rest of the per-itemtype fields emit.

## Test plan

- [x] Unit: 3 new tests for the place field, all passing.
- [x] Full suite: 438/438 passing locally.
- [ ] Manual round-trip against a real Zotero library — pending live verification on a book item.

## Notes

Mirrors the convention #195 established when ten other named fields were exposed. No behavioural changes for existing parameters; `place` is purely additive.

### DIFF
--- a/src/zotero_mcp/tools/write.py
+++ b/src/zotero_mcp/tools/write.py
@@ -723,6 +723,7 @@ _UPDATE_ITEM_API_TO_PARAM = {
     "issue": "issue",
     "pages": "pages",
     "publisher": "publisher",
+    "place": "place",
     "ISSN": "issn",
     "language": "language",
     "shortTitle": "short_title",
@@ -760,6 +761,7 @@ def update_item(
     issue: str | None = None,
     pages: str | None = None,
     publisher: str | None = None,
+    place: str | None = None,
     issn: str | None = None,
     language: str | None = None,
     short_title: str | None = None,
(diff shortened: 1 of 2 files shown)


54yyyu:
Thanks for contributing to zotero-mcp!

PR #300, opened 2026-05-22
fix(local_db): extract fulltext from .txt/.vtt/.srt-style attachments (#265)

### LINKED ISSUE #265: `zotero_get_item_fulltext` returns 404 in local mode for non-PDF/HTML attachments
**Target:** <https://github.com/54yyyu/zotero-mcp/issues>
**Filed by:** scaltagi

---

## Title

`zotero_get_item_fulltext` returns 404 in local mode for non-PDF/HTML attachments (`.txt`, `.vtt`, `.pptx`, `.docx`, `.epub`, video, …) even when the attachment is active and on-disk

### Summary

In local mode, `zotero_get_item_fulltext` silently fails with a 404 against `file://localhost/...` for any attachment whose `contentType` is not `application/pdf` or `text/html*`, even when the file is present in `~/Zotero/storage/{KEY}/` and the attachment is active (`data.deleted = 0`). PDF and HTML attachments work as expected.

The 404 is misleading: the file exists, the metadata is correct, the web API serves the bytes fine, but the local-mode extractor refuses to consider the attachment, then falls back to a code path that cannot work over the local API.

### Affected version

- `zotero-mcp` `0.3.0` (installed via `uv` from PyPI)
- Zotero desktop with local API at `http://localhost:23119`
- Library type: `group` (also reproducible against `user` libraries based on the code path)
- OS: Ubuntu 24.04 (WSL2); Zotero storage on Windows host accessed via `/mnt/c/Users/<user>/Zotero`

### Reproduction

```python
# Minimal repro — pyzotero against the local API, same code path zotero-mcp uses internally
from pyzotero import zotero

z_local = zotero.Zotero(library_id, "group", api_key, local=True)
z_local.file("AAAAAAAA")   # AAAAAAAA = a .txt or .vtt attachment key — 404
z_local.file("BBBBBBBB")   # BBBBBBBB = a .pdf attachment key      — also 404 (!)

z_web = zotero.Zotero(library_id, "group", api_key)  # local=False
z_web.file("AAAAAAAA")     # OK — returns full bytes
z_web.file("BBBBBBBB")     # OK — returns full bytes
```

Or end-to-end through the MCP tool:

```text
mcp__zotero__zotero_get_item_fulltext(item_key="<txt or vtt key>")
→ "Error accessing attachment: \nCode: 404\nURL: file://localhost/C:/Users/.../storage/<KEY>/<filename>\nMethod: GET\nResponse: "

mcp__zotero__zotero_get_item_fulltext(item_key="<pdf key>")
→ Full text returned successfully
```

### Expected behavior

For a non-PDF/HTML attachment whose file is present on disk, `zotero_get_item_fulltext` should return the file's text content (or, for binary formats, a clear "unsupported content type" message — never a misleading 404).

### Actual behavior

Returns 404 with the `file://localhost/...` URL, identical to the response for a missing/deleted attachment. There is no signal that the file exists and is readable.

### Root cause

The local-mode extractor at [`zotero_mcp/local_db.py:_extract_fulltext_for_item()`](https://github.com/54yyyu/zotero-mcp/blob/main/src/zotero_mcp/local_db.py) gates candidate-collection on content type and only accepts PDF or HTML:

```python
# zotero_mcp/local_db.py — line numbers from installed v0.3.0
def _extract_fulltext_for_item(self, item_id: int) -> tuple[str, str] | None:
    best_pdf = None
    best_html = None
    for key, path, ctype in self._iter_parent_attachments(item_id):
        resolved = self._resolve_attachment_path(key, path or "")
        if not resolved or not resolved.exists():
            continue
        if ctype == "application/pdf" and best_pdf is None:
            best_pdf = resolved
        elif (ctype or "").startswith("text/html") and best_html is None:
            best_html = resolved
    target = best_pdf or best_html
    if not target:
        return None                          # <-- non-PDF/HTML attachments fall out here
    ...
```

When `target` is `None`, the caller in [`zotero_mcp/tools/retrieval.py:get_item_fulltext()`](https://github.com/54yyyu/zotero-mcp/blob/main/src/zotero_mcp/tools/retrieval.py) falls through to:

1. `zot.fulltext_item()` — Zotero's pre-indexed text. Empty for `.vtt`, `.txt`, presentations, etc.
2. `zot.dump()` → `zot.file()` — uses the local API. **Zotero's local API never returns raw bytes**; it returns a `file://localhost/...` URL pointer. pyzotero then tries to HTTP-GET that URL and returns 404.

PDFs do not hit this path because they are caught earlier by the local-storage extractor. Non-PDF/HTML attachments are not, so they always reach the broken `zot.dump()` fallback.

The unreachable code: a few lines above the gate, [`_extract_text_from_file()`](https://github.com/54yyyu/zotero-mcp/blob/main/src/zotero_mcp/local_db.py) (lines 341–352 in v0.3.0) already has a generic `file_path.read_text(errors="ignore")` fallback. The candidate-collection gate just never adds non-PDF/HTML files to the candidate list, so the generic fallback never runs.

### Side effects

The same `_extract_fulltext_for_item` populates ChromaDB during `zotero-mcp update-db`, so any non-PDF/HTML attachment in a zotero-mcp semantic-search index has **no fulltext indexed** — only metadata. This silently degrades `zotero_semantic_search` quality for libraries that contain transcripts, slide decks, plaintext notes, etc.

### Suggested fix

Extend the candidate-collection loop to add a third bucket (or a single generic bucket) for any attachment `_extract_text_from_file()` can already handle:

```python
def _extract_fulltext_for_item(self, item_id: int) -> tuple[str, str] | None:
    best_pdf = None
    best_html = None
    best_other = None
    for key, path, ctype in self._iter_parent_attachments(item_id):
        resolved = self._resolve_attachment_path(key, path or "")
        if not resolved or not resolved.exists():
            continue
        if ctype == "application/pdf" and best_pdf is None:
            best_pdf = resolved
        elif (ctype or "").startswith("text/html") and best_html is None:
            best_html = resolved
        elif best_other is None and (ctype or "").startswith("text/"):
            # text/plain, text/vtt, text/markdown, text/csv, ...
            best_other = resolved
    target = best_pdf or best_html or best_other
    if not target:
        return None
    text = self._extract_text_from_file(target)
    ...
    source = (
        "pdf" if target.suffix.lower() == ".pdf"
        else "html" if target.suffix.lower() in {".html", ".htm"}
        else "file"
    )
    return (text, source)
```

Conservative: this only opens `text/*` content. If you'd like to also support `.pptx`, `.docx`, `.epub`, `.vtt` properly, those need real extractors (not `read_text`) — that is a larger change and could be deferred to a follow-up issue.

### Workaround

For consumers of zotero-mcp that maintain their own document index, route fulltext for non-PDF/HTML attachments through their own extraction pipeline rather than calling `zotero_get_item_fulltext`. Until the upstream fix lands, the 404 cannot be distinguished from a real "file is gone" condition without comparing against `attachment.contentType`.

### Impact

In our library we have hundreds of attachments at risk. Anyone using zotero-mcp on a library that contains course transcripts, lecture decks, conference talks, or Markdown notes will see the same silent 404 pattern.

### Related

- pyzotero `Zotero(local=True).file()` is the canonical example of why the local API can't be the fallback path: <https://github.com/urschrei/pyzotero/blob/main/src/pyzotero/zotero.py>

brianckeegan:
## Summary

#265: ``zotero_get_item_fulltext`` returned 404 in local mode for any attachment whose content type wasn't ``application/pdf`` or ``text/html*`` — even when the file was right there on disk. Same code path also silently dropped the attachment from semantic indexing.

Root cause: ``_extract_fulltext_for_item`` in [src/zotero_mcp/local_db.py](src/zotero_mcp/local_db.py:376) only collected PDF and HTML candidates. For non-PDF/HTML attachments ``target`` was ``None``, the caller in ``tools/retrieval.py`` fell through to ``zot.dump()``/``zot.file()`` — but pyzotero's local-API mode returns a ``file://localhost/...`` URL pointer rather than raw bytes, and the subsequent HTTP GET 404s. The misleading 404 looked identical to a missing attachment, masking that the file existed and was readable.

The ``_extract_text_from_file`` generic ``read_text(errors=\"ignore\")`` fallback already existed (lines 363–367) — the gate just never let non-PDF/HTML attachments reach it.

## Fix

Add a third "any textual attachment" bucket gated by a small allowlist on a new ``_is_extractable_attachment`` classmethod:

- Any ``text/*`` MIME type, plus a curated set (``text/vtt``, ``application/json``, ``application/xml``, …).
- Known textual extensions: ``.txt``, ``.vtt``, ``.srt``, ``.sbv``, ``.md``, ``.markdown``, ``.rst``, ``.csv``, ``.tsv``, ``.json``, ``.xml``, ``.log``, ``.text``.

Binary office formats (``.docx``, ``.pptx``, ``.xlsx``) and media (``.mp4`` etc.) are **not** accepted — the existing ``read_text(errors=\"ignore\")`` fallback would return garbage for those and pollute the semantic index. Those formats need format-specific extractors (markitdown / ebooklib etc.); could be a follow-up.

Preference order is unchanged: PDF first, then HTML, then any textual attachment.

## Test plan

- [x] ``tests/test_fulltext_non_pdf_html.py`` — 11 new tests. Covers the classifier (text/plain, text/vtt, .srt extension, unknown ``text/*`` subtype, .docx rejected, video rejected). End-to-end: solo ``.vtt`` yields fulltext (the #265 main repro), PDF still wins when both exist, ``.docx``-only item still returns None.
- [x] ``tests/test_local_db.py`` (15 tests) continues to pass.

Fixes #265.

### DIFF
--- a/src/zotero_mcp/local_db.py
+++ b/src/zotero_mcp/local_db.py
@@ -5,16 +5,15 @@ Provides direct SQLite access to Zotero's local database for faster semantic sea
 when running in local mode.
 """
 
-import os
-import sqlite3
-import platform
 import logging
+import os
+import platform
+import sqlite3
+from dataclasses import dataclass
 from pathlib import Path
 from typing import Any
-from dataclasses import dataclass
-from urllib.parse import urlparse, unquote
 
-from .utils import is_local_mode, _normalize_for_search
+from .utils import _normalize_for_search, is_local_mode
 
 logger = logging.getLogger(__name__)
 
@@ -231,7 +230,7 @@ class LocalZoteroReader:
 
         # Linked file as URL: 'file:///path/to/file.pdf'
         if zotero_path.startswith("file://"):
-            from urllib.parse import urlparse, unquote
+            from urllib.parse import unquote, urlparse
             parsed = urlparse(zotero_path)
             decoded_path = unquote(parsed.path or "")
             # file:///C:/... on Windows
@@ -353,6 +352,36 @@ class LocalZoteroReader:
         except Exception:
             return ""
 
+    # Extensions / MIME types we know we can read as plain text. Used by
+    # ``_is_extractable_attachment`` to gate non-PDF/HTML attachments into
+    # the fulltext extractor. Binary formats (.docx, .pptx, .epub, video,
+    # etc.) are intentionally excluded — ``read_text`` returns garbage for
+    # those and we don't want to pollute the semantic index with it.
+    _TEXTUAL_SUFFIXES = frozenset({
+        ".txt", ".vtt", ".srt", ".sbv", ".md", ".markdown", ".rst",
+        ".csv", ".tsv", ".json", ".xml", ".log", ".text",
+    })
+    _TEXTUAL_CONTENT_TYPES = frozenset({
+        "text/plain", "text/vtt", "text/markdown", "text/csv",
+        "text/tab-separated-values", "text/srt", "application/json",
+        "application/xml", "text/xml",
+    })
+
+    @classmethod
+    def _is_extractable_attachment(cls, file_path: Path, ctype: str | None) -> bool:
+        """Return True when ``_extract_text_from_file`` can return useful text.
+
+        PDF and HTML are handled by their dedicated extractors elsewhere. For
+        anything else, accept the attachment iff its MIME type or extension
+        is in the textual allowlist — never accept arbitrary binaries.
+        """
+        normalized_ctype = (ctype or "").lower()
+        if normalized_ctype.startswith("text/"):
+            return True
+        if normalized_ctype in cls._TEXTUAL_CONTENT_TYPES:
+            return True
+        return file_path.suffix.lower() in cls._TEXTUAL_SUFFIXES
+
     def _extract_text_from_file(self, file_path: Path) -> str:
         """Extract text content from a file based on extension, with fallbacks."""
         suffix = file_path.suffix.lower()
@@ -376,11 +405,18 @@ class LocalZoteroReader:
     def _extract_fulltext_for_item(self, item_id: int) -> tuple[str, str] | None:
         """Attempt to extract fulltext and source from the item's best attachment.
 
-        Preference: use PDF when available; fall back to HTML when no PDF exists.
-        Returns (text, source) where source is 'pdf' or 'html'.
+        Preference: PDF first, then HTML, then any textual attachment
+        (transcript .vtt, plain .txt, captions, etc. — see
+        ``_TEXTUAL_SUFFIXES``). Returns ``(text, source)`` where ``source``
+        is ``"pdf"``, ``"html"``, or ``"file"``.
+
+        Non-PDF/HTML attachments previously fell out of the candidate
+        collection entirely, so transcripts, plaintext notes, and similar
+        attachments were never indexed (#265).
         """
         best_pdf = None
         best_html = None
+        best_other = None
         for key, path, ctype in self._iter_parent_attachments(item_id):
             resolved = self._resolve_attachment_path(key, path or "")
             if not resolved or not resolved.exists():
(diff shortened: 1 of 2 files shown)


54yyyu:
Merged manually after resolving conflicts with #301. Thanks!

PR #307, opened 2026-05-25
feat(semantic): throttle embeddings via --burst/--pause + 429 retry-with-backoff

hendryman:
## Summary

Adds two complementary ways to keep `zotero-mcp update-db` from tripping
embedding API rate limits — proactive pacing and reactive retry. Both
default to off so behavior is unchanged unless opted in.

Related to #152 (which asked for the inverse — higher-tier throughput).
This PR shares part of the same vocabulary (configurable batch/pause)
but the goal is rate-limit safety, not concurrency. Happy to fold or
split as the maintainer prefers.

## Problem

Gemini `gemini-embedding-001` free-tier limits are 100 RPM / 30k TPM
(Tier 1: 1500 RPM / 1M TPM). With `--fulltext`, a single 25-item outer
batch ships ~100-200k tokens in one `embed_content` call — guaranteed
TPM violation on free tier, and under bursty load Tier 1 trips too. The
existing end-of-run retry uses `time.sleep(1)` which is too short to
clear a Gemini quota window (60s).

### 1. Pacing (`--burst` / `--pause`)

New flags on `update-db`:

```
zotero-mcp update-db --fulltext --burst 500 --pause 10
```

Counter accumulates items processed since the last pause; when it
crosses `--burst`, sleep `--pause` seconds. Skips the final pause when
no items remain.

Config fallback under `semantic_search.throttle`:

```json
{
  "semantic_search": {
    "throttle": { "burst_items": 500, "pause_seconds": 10 }
  }
}
```

CLI args override config. Both `--burst` and `--pause` (or both config
keys) must be set to engage; passing one alone logs a warning and runs
without pacing.

### 2. Reactive retry-with-backoff

`OpenAIEmbeddingFunction.__call__` and `GeminiEmbeddingFunction.__call__`
/ `embed_query` are wrapped in `_with_retry`. Schedule: 30s → 60s → 120s,
3 attempts max. First wait is 30s (not 1s) because Gemini's TPM/RPM
windows are 60s.

Detector is intentionally narrow to avoid false positives:
- `getattr(exc, 'code', None) == 429`
- `getattr(exc, 'status_code', None) == 429`
- `'RESOURCE_EXHAUSTED' in str(exc)` or `'429' in str(exc)`
- `isinstance(exc, openai.RateLimitError)` (if openai SDK importable)

HuggingFace embeddings are untouched (local model, no API).

## Tests

New `tests/test_embedding_throttle.py` covers:

- `_is_rate_limit_error` recognizes code/status_code/canonical substring
  shapes; rejects arbitrary "rate limit" text without canonical markers.
- `_with_retry` recovers from a synthetic 429 within `max_attempts`,
  re-raises non-rate-limit errors immediately, gives up after exhausting
  attempts.
- Burst/pause loop pauses exactly 2× for 60 items at burst=25 (no final
  pause), exactly 1× for 50 items, and not at all when the flags aren't
  passed.

All 30 existing semantic tests (`test_semantic_stats.py`,
`test_semantic_search_quality.py`) still pass.

## Test plan

- [x] `pytest tests/test_embedding_throttle.py` — 7 new tests pass
- [x] `pytest tests/test_semantic_stats.py tests/test_semantic_search_quality.py` — 30 existing tests pass
- [x] `ruff check` on touched files — no new violations introduced
- [x] Verified end-to-end on real library (~Tier 1 Gemini key) with `--burst 500 --pause 10` — no more 429s during a `--fulltext` reindex
- [x] `zotero-mcp update-db --help` shows the new flags with documentation

## Notes for maintainer

- The `_with_retry` helper has no upper-bound timeout on individual sleeps
  beyond the max_attempts cap (so worst case: 30+60+120 = 210s extra wait
  before giving up). Happy to add a `max_total_sleep` if desired.
- If you'd prefer a single `--throttle "burst:pause"` flag over two flags,
  I can refactor.
- The detector swallows ImportError on `openai`; if you'd rather hard-require
  it, easy change.

### DIFF
--- a/src/zotero_mcp/chroma_client.py
+++ b/src/zotero_mcp/chroma_client.py
@@ -6,10 +6,12 @@ for semantic search over Zotero libraries.
 """
 
 import json
+import logging
 import os
+import time
+from collections.abc import Callable
 from pathlib import Path
 from typing import Any
-import logging
 
 try:
     import chromadb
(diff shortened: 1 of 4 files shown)


54yyyu:
Thanks — the throttle design (off by default, `--burst`/`--pause`, end-of-run pause skip) and the 429 backoff are solid. One blocker before merge: `_is_rate_limit_error` matches `"429" in msg`, which is a false-positive magnet — any error string that happens to contain `429` (a token count, byte offset, item key, etc.) would trigger the full 30/60/120s retry chain on a doomed call. Please narrow it to HTTP status checks (response `status_code`/`.code`) plus provider sentinels like `RESOURCE_EXHAUSTED`, and drop the bare substring match. Happy to merge once that's tightened.

54yyyu:
Superseded by #356, which adds per-request rate-limit throttling () plus sub-batching for provider input caps. Thanks for raising this @hendryman!

54yyyu:
Superseded by #356, which adds per-request rate-limit throttling (the `rate_limit_rps` embedding config) plus sub-batching for provider input caps. Thanks for raising this @hendryman!

PR #245, opened 2026-04-21
Expose native Zotero fields on read and write paths (#236, #238, #240, #234)

rbpasker:
Consolidates four related "field not exposed" fixes into a single PR. Each issue is a separate commit on this branch — squash-merge or rebase-merge both work.

## Issues addressed

| Issue | Commit | What it does |
|---|---|---|
| [#236](https://github.com/54yyyu/zotero-mcp/issues/236) | `fix: render ISBN and ISSN in markdown metadata output` | The markdown formatter silently dropped ISBN (and ISSN) while BibTeX emitted them. Agents concluded these identifiers were missing and risked overwriting valid stored values. |
| [#238](https://github.com/54yyyu/zotero-mcp/issues/238) | `fix: expose publication place on all read paths` | The `place` field was absent from markdown, BibTeX (`address = {...}`), and the connector `fetch.metadata` dict. Reverification workflows couldn't determine whether users had populated it. |
| [#240](https://github.com/54yyyu/zotero-mcp/issues/240) | `fix: add access_date and place to zotero_update_item` | `zotero_update_item` didn't accept `accessDate` (needed for CSL-compliant webpage closure) or `place` (round-trip companion to the #238 read fix). |
| [#234](https://github.com/54yyyu/zotero-mcp/issues/234) | `feat: expose item_type in zotero_update_item for type migration` | Type-corruption from Zotero Connector/DOI misclassification (book landing pages saved as `journalArticle`) had no programmatic repair path. |

## Read-side changes (`client.py`, `connectors.py`)

- **Markdown** (`format_item_metadata`):
  - Emit `**ISBN:** ...`, `**ISSN:** ...` alongside the existing `**DOI:** ...` line.
  - Replace the book-only `**Publisher:** X, Y` inline concatenation with independent `**Publisher:** X` / `**Place:** Y` lines that apply to any item type with those fields.
  - Add `**Book:** ...` and `**Pages:** ...` for `bookSection` so chapter metadata is surfaced.
- **BibTeX** (`generate_bibtex`): add `place → address` and `bookTitle → booktitle` to the field mappings.
- **Connector `fetch`** (`tools/connectors.py`): add `place`, `publisher`, `isbn`, `issn` to the returned metadata dict.

## Write-side changes (`tools/write.py`)

- `zotero_update_item` gains three new optional parameters:
  - `access_date` → native `accessDate` (webpage/document/etc.)
  - `place` → native `place` (book/bookSection/thesis)
[...]

Issue #240, opened 2026-04-21
Bug: zotero-mcp: zotero_update_item does not expose native accessDate field


rbpasker:
zotero-mcp: zotero_update_item does not expose native accessDate field

Date observed:  2026-04-21
MCP server:     zotero-mcp
Severity:       medium (functionally blocks webpage-type closure without a CSL-extra workaround)

SUMMARY
-------
The `zotero_update_item` tool's parameter schema omits Zotero's native `accessDate`
field. Webpage, document, and several other item types expose `accessDate` in the
Zotero data model and in the Info pane, but the MCP does not accept it as a write
target. This is structurally analogous to BUG-2 (place-field not readable / not
writable) but applies to a field the CSL specification explicitly requires for
accurate bibliography rendering of web resources.

OBSERVED PARAMETER SCHEMA (current)
-----------------------------------
abstract, add_tags, book_title, collection_names, collections, creators, date,
doi, edition, extra, isbn, issn, issue, item_key, language, pages,
publication_title, publisher, remove_tags, short_title, tags, title, url, volume

MISSING
-------
accessDate (native field on webpage, document, artwork, audioRecording, blogPost,
            case, computerProgram, conferencePaper, dictionaryEntry,
            encyclopediaArticle, film, forumPost, hearing, instantMessage,
            interview, letter, magazineArticle, manuscript, map, newspaperArticle,
            podcast, presentation, radioBroadcast, report, statute, tvBroadcast,
            thesis, videoRecording, webpage — per Zotero type map 2026)

REPRODUCTION
------------
1. Create or retrieve a webpage-type Zotero item, e.g. DSCL5RKL.
2. Attempt: zotero_update_item(item_key="DSCL5RKL", accessDate="2026-04-21")
3. Observed: parameter validation error (accessDate not in schema) OR silent drop.
4. Workaround used: zotero_update_item(item_key="DSCL5RKL", extra="Accessed: 2026-04-21")
   — writes to the Extra field; CSL processors honor `Accessed:` prefix at render
   time (CSL specification, date-variables §accessed). Native Info-pane Accessed
   field remains empty; user must migrate manually if desired.

FIX CANDIDATES
--------------
(a) Add `access_date` to the `zotero_update_item` parameter schema, validated per
    Zotero's date-field rules (ISO-8601 preferred; partial granularity allowed).
    Route to the native `accessDate` API field for item types that support it;
    reject for types that do not.
[...]

Issue #238, opened 2026-04-21
Bug: place field not exposed on any read path


rbpasker:
Component: zotero-mcp (MCP server)

Tools: zotero_get_item_metadata (markdown + bibtex), mcp__zotero__fetch

Observed: 2026-04-21, three bookSection items (XXXXXXX...) plus earlier inventory sweep

Symptom: the 'place' / 'publication place' field of a Zotero item
  is absent from every MCP read response. Markdown renderer omits it.
  BibTeX renderer omits it (no address = {...} line). fetch.metadata
  exposes only itemType, date, key, doi, authors, tags, zotero_url,
  web_url, source.

Companion defect: zotero_update_item does not accept a place
  parameter either (documented; this report covers the read side).

Impact: reverification workflows cannot determine whether a user has
  populated the place field via Zotero desktop.

Reproduction: populate place on any item via Zotero desktop, then
  call zotero_get_item_metadata (either format) or
  mcp__zotero__fetch; observe place absent from response.

Fix candidates: (a) add address = {...} to bibtex renderer for
  book/bookSection/thesis types; (b) add **Place:** line to markdown
  renderer; (c) include place in fetch.metadata dict; (d) add place
  parameter to zotero_update_item for round-trip parity.

54yyyu:
fixed in #245

PR #321, opened 2026-06-08
feat: add citation_key field to zotero_update_item (#320)

### LINKED ISSUE #320: feat: add `citation_key` field to zotero_update_item
## Summary

`zotero_update_item` doesn't accept `citation_key` (Zotero's native `data.citationKey` field). The Zotero cloud API accepts it in PATCH bodies; this is a structural gap in the MCP, not a Zotero limitation.

## Why this matters

When items are added programmatically (any add path), BetterBibTeX pins the citekey from whatever metadata exists at the moment of add — often incomplete. By the time the workflow has populated full metadata, the citekey is stuck on a malformed value, for example:

- `doeOldPaperdraftpdf2024` — filename fragment + capitalised author (happens when metadata is empty and BBT falls back to filename)
- `IntroductionTopicAnalysis` — no author, no year (incomplete-metadata pin before creators are populated)

In BBT 9.x there is **no programmatic refresh path** to fix these after the fact:

- BBT JSON-RPC has no mutation methods on citekeys (tracked upstream at retorquere/zotero-better-bibtex#3522)
- Zotero local API at `:23119` returns 501 on item PATCH
- BBT's Extra-field `Citation Key:` override is not honoured in 9.x for items with a set `data.citationKey`

The **cloud API does accept `data.citationKey` in PATCH bodies** — exposing `citation_key` here gives users the only working programmatic mutation path.

## Proposed change

Add `citation_key: Optional[str] = None` to the `update_item` signature in `src/zotero_mcp/tools/write.py` (around line 1207). When provided, include it in the PATCH body as `data.citationKey`.

## Precedent

Direct shape precedent: [PR #281](https://github.com/54yyyu/zotero-mcp/pull/281) (*"feat: add `place` field to zotero_update_item"*, 94 lines, 2 files). Also #240 (accessDate exposure) and #238 (place field), both resolved in PR #245.

PR with the implementation is attached below.

ahharvey:
Closes #320.

## Summary

Exposes Zotero's native `data.citationKey` on `zotero_update_item` via a new `citation_key` parameter, matching the shape of the place / accessDate / ISBN field additions in PR #245 / #281.

## Why

When items are added programmatically (any add path), BetterBibTeX pins the citekey from whatever metadata exists at the moment of add — often incomplete. By the time the workflow has populated full metadata, the citekey is stuck on a malformed value, and BBT 9.x exposes no programmatic refresh:

- BBT JSON-RPC has no mutation methods on citekeys (tracked upstream at retorquere/zotero-better-bibtex#3522)
- Zotero local API at `:23119` returns 501 on item PATCH
- BBT's Extra-field `Citation Key:` override is not honoured in 9.x for items with a set `data.citationKey`

The Zotero cloud API does accept `data.citationKey` in PATCH bodies, so exposing it here gives users the only working programmatic mutation path for malformed pinned keys.

## Change

`src/zotero_mcp/tools/write.py`:

- Add `citation_key: Annotated[str | None, Field(...)] = None` parameter to `update_item` signature (Annotated to surface the BBT context in the field description, matching the `place=` pattern)
- Wire it into `field_updates` as `data.citationKey`
- Register in `_UPDATE_ITEM_API_TO_PARAM`
- Update tool description and docstring

`tests/test_update_item.py`:

- Add `citationKey` to `_make_item` fixture (representative of Zotero 9.x items, which populate it natively)
- Add `test_update_citation_key` covering the round-trip

## Tests

- 56/56 pass in `tests/test_update_item.py` (including the new test)
- Full suite: 809 pass, 2 pre-existing failures in `tests/test_place_field_reads.py` confirmed on upstream `main` (BBT-integration assertion mismatch, unrelated to this change)

## Diff size

+41/-5 across 2 files. Comparable to:
- #240 (accessDate exposure, in PR #245)
- #238 (place field, in PR #245)
- #281 (place field expansion, +94/-1)

### DIFF
--- a/src/zotero_mcp/tools/write.py
+++ b/src/zotero_mcp/tools/write.py
@@ -1172,6 +1172,7 @@ _UPDATE_ITEM_API_TO_PARAM = {
     "edition": "edition",
     "ISBN": "isbn",
     "bookTitle": "book_title",
+    "citationKey": "citation_key",
 }
 
 
@@ -1192,7 +1193,8 @@ _UPDATE_ITEM_API_TO_PARAM = {
         "item_key: 8-character Zotero item key of the item to update. "
         "Editable fields include: title, creators, date, publisher, place, "
         "publication_title, volume, issue, pages, DOI, ISBN, ISSN, url, "
-        "language, abstract, short_title, edition, book_title, extra, item_type. "
+        "language, abstract, short_title, edition, book_title, extra, "
+        "citation_key, item_type. "
         "To migrate an item across types (e.g., journalArticle → book), pass item_type "
         "with a valid Zotero item-type vocabulary value; overlapping fields are preserved "
         "and type-specific fields that do not map to the target type are dropped. "
@@ -1234,6 +1236,10 @@ def update_item(
     edition: str | None = None,
     isbn: str | None = None,
     book_title: str | None = None,
+    citation_key: Annotated[
+        str | None,
+        Field(description="BetterBibTeX / Zotero native citation key. Writes to data.citationKey. Useful when BBT auto-pinned the key from incomplete metadata and the programmatic refresh path is blocked (see https://github.com/retorquere/zotero-better-bibtex/issues/3522)."),
+    ] = None,
     item_type: str | None = None,
     *,
     ctx: Context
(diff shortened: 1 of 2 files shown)


54yyyu:
Thanks, this is the right approach and complements #319 (read side, now merged). One issue before merge: the field-write loop only sets `citationKey` when it's already present on the fetched item (`if field in data`), so passing `citation_key` to an item that doesn't yet have one silently no-ops and emits a misleading 'not valid for item type' warning — which is exactly #320's case. The test currently masks this because `_make_item` was edited to always include `citationKey`. Please special-case adding `citationKey` even when absent (and adjust the test to seed an item without it). Will merge once that's handled.

ahharvey:
Pushed c7fa913 — thanks for catching this, the `_make_item` default was exactly what was hiding it.

- Added an `elif field == "citationKey":` branch so absence routes to add rather than the skip-as-invalid warning
- Reworked `_make_item` to omit `citationKey` by default
- Split the existing test into `test_add_citation_key_when_absent` and `test_update_citation_key_when_present` so both paths are explicitly covered

All 57 tests in `test_update_item.py` green locally. Happy to adjust further if anything looks off.

PR #325, opened 2026-06-08
Update README with Zotero configuration instructions

raffaelemancuso:
Added instructions for configuring Zotero local API in README.


raffaelemancuso:
please merge this as it contains important setup information

54yyyu (changes requested):
Thanks for the docs improvement! Two things to fix before merging:

- The screenshot is embedded via a `private-user-images.githubusercontent.com` URL with a signed token that expires, so it will 404 later. Could you commit the image into the repo (e.g. under `docs/`) and reference it by relative path?
- The JSON examples use unquoted values like `"ZOTERO_LOCAL": false` and a bare `YOUR_LIBRARY_ID`, which isn't valid JSON for the config. Could you quote them as strings to match how the MCP clients parse the config?

Once those are sorted I'm happy to merge!

54yyyu (changes requested):
Thanks, the 'Configure Zotero' section is genuinely useful: the local API toggle is the most common setup failure and a screenshot helps. Two blockers before this can merge:

1. **The screenshot will break.** The `` points at a `private-user-images.githubusercontent.com` URL with an expiring JWT (it is already dead). Please commit the image to the repo (e.g. `docs/assets/zotero-local-api.png`) and reference it with a relative path.

2. **The unquoted-values advice is incorrect for Claude Desktop.** In the MCP config schema, `env` values are strings; `"ZOTERO_LOCAL": false` and a bare numeric `ZOTERO_LIBRARY_ID` make the config invalid for the client this section documents. The existing quoted form (`"ZOTERO_LOCAL": "true"`) is correct. If a different client rejected quoted values for you, let me know which one and we can document that separately, but the main example must stay quoted.

The note that shell environment variables override config values is a fair addition and can stay. If you drop the quoting changes and commit the screenshot, I'll merge the rest.

raffaelemancuso:
Done!

54yyyu (changes requested):
Thanks for the quick turnaround. The quoting fix is good, all values are strings now. But a few problems remain, two of them new in this commit:

1. **The screenshot file is not in the PR.** The README now references `./docs/zotero-local-api.png`, but the only changed file is README.md, so the image 404s immediately. Please `git add` the png itself under `docs/`.

2. **The config file paths are incorrect.** Claude Desktop reads `claude_desktop_config.json` (the file the original text named), not `~/.claude.json`. `~/.claude.json` is Claude Code's config, and Claude Code MCP servers do not go in `~/.claude/settings.json`. For Claude Code the simplest instruction is:

   ```bash
   claude mcp add zotero -- zotero-mcp serve
   ```

   Please restore `claude_desktop_config.json` for the Desktop section and use `claude mcp add` (or `~/.claude.json`) for Claude Code.

3. **Write mode should keep `ZOTERO_LOCAL: "true"`.** With an API key and library ID set, the server runs in hybrid mode: fast local reads, web API writes. Setting `ZOTERO_LOCAL: "false"` routes all reads through the web API too, which is slower and contradicts the new 'Configure Zotero' section saying the local API must be enabled. Please revert the example and the write-mode bullet to `"true"`.

The 'Configure Zotero' section itself and the shell-env-override note are good to go. Happy to merge once these three are sorted.

raffaelemancuso:
Done!

54yyyu:
Merged, thanks for the contribution and the revisions. The screenshot and config paths were all good; I pushed one small commit to your branch before merging that reverts the write-mode bullet to keeping ZOTERO_LOCAL="true", since adding API credentials on top of local mode gives hybrid mode (fast local reads, web API writes) rather than requiring a switch to the web API for everything.

PR #332, opened 2026-06-10
feat(docker): publish multi-flavor GHCR images and document container runtime/persistence

mcree:
## Summary
- Add GitHub Actions workflow to build and publish multi-arch Docker images to GHCR for `core` and `all` flavors.
- Add container entrypoint support for both MCP server mode and standalone CLI mode.
- Expand Docker docs (including README) with container env var guidance and ChromaDB persistence instructions.

### Docker image/runtime
- Updated `Dockerfile` to support build-time flavor selection via `INSTALL_EXTRAS`.
- Added `docker/entrypoint.sh`:
  - default server behavior: `zotero-mcp serve --transport stdio` (or `ZOTERO_TRANSPORT` override)
  - CLI mode via `ZOTERO_APP=cli` using `zotero-cli`

### CI/CD publishing
- Added `.github/workflows/docker.yml`:
  - triggers on PR, `main`, and `v*` tags
  - builds `core` and `all` flavor images
  - publishes to `ghcr.io/<owner>/<repo>`
  - tag strategy includes semver, sha, `latest`, and flavor-suffixed tags (`-core`, `-all`)
  - unsuffixed tags map to `all`

### Documentation
- Updated `README.md` Docker section with:
  - container runtime env vars (`ZOTERO_APP`, `ZOTERO_TRANSPORT`)
  - note that standard Zotero MCP env vars are supported in containers
  - ChromaDB/config persistence mount path (`/home/app/.config/zotero-mcp`)
  - persistent volume usage examples
- Added/expanded `docs/docker-images.md` with:
  - flavor/tag behavior
  - env var reference for container usage
  - ChromaDB persistence behavior and examples

## Why
This introduces a clear and supported Docker distribution story for different dependency footprints (`core` vs `all`), enables both MCP-server and CLI container workflows, and prevents accidental loss of semantic index data by documenting persistent ChromaDB storage.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,5 +1,8 @@
 name: Build and publish Docker images
 
+env:
+  FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: true
+
 on:
   push:
     branches:
@@ -27,17 +30,17 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@v6
 
       - name: Set up QEMU
-        uses: docker/setup-qemu-action@v3
+        uses: docker/setup-qemu-action@v4
 
       - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v3
+        uses: docker/setup-buildx-action@v4
 
       - name: Log in to GHCR
         if: github.event_name != 'pull_request'
-        uses: docker/login-action@v3
+        uses: docker/login-action@v4
         with:
           registry: ghcr.io
           username: ${{ github.actor }}
@@ -51,7 +54,7 @@ jobs:
 
       - name: Docker metadata
         id: meta
-        uses: docker/metadata-action@v5
+        uses: docker/metadata-action@v6
         with:
           images: ${{ steps.image.outputs.name }}
           tags: |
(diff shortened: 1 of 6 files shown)


mcree:
any issues with this?

54yyyu:
Thanks for the patience on this one, and apologies for the slow review. Publishing GHCR images needs decisions on my side (package visibility, release automation wiring, which flavors to commit to supporting), so I don't want to merge the workflow half-blind. I'll take a proper pass at this for an upcoming release cycle. If you're still running these images yourself, a note on which flavor combinations you actually use in practice would help me trim the matrix.

mcree:
> Thanks for the patience on this one, and apologies for the slow review. Publishing GHCR images needs decisions on my side (package visibility, release automation wiring, which flavors to commit to supporting), so I don't want to merge the workflow half-blind. I'll take a proper pass at this for an upcoming release cycle. If you're still running these images yourself, a note on which flavor combinations you actually use in practice would help me trim the matrix.

Thanks. I personally use the [pdf] flavor, but switching to the [all] image would make little practical difference for me, since the main impact is image size. The only factor currently preventing me from using [all] is the significantly longer build time.

I would therefore propose keeping the [core] and [all] flavors as presented in this PR. I agree that supporting every possible flavor combination would create too much maintenance overhead. The current implementation supports a reasonable path forward, though I could also adjust it to generate combinations from a configurable flavor list.

I also do not see image-release automation itself as a major concern, since the project already provides - and presumably maintains - a Dockerfile. However, I am not sure what you mean by “package visibility” in this context. Could you clarify that point?

The larger concern, in my view, would be integration testing of the published images, particularly for configurations that depend on external vector databases.

PR #333, opened 2026-06-10
fix(semantic): hold sync watermark when local sqlite snapshot lags the API (#292)

### LINKED ISSUE #292: bug: immutable SQLite 读不到 WAL 时仍推进 last_update，可能导致新添加文献被增量更新永久跳过
## 问题概述

当前 local mode 读取 Zotero sqlite 时使用了：

```python
sqlite3.connect(f"file:{self.db_path}?immutable=1", uri=True)
```

这个做法可以绕过 Zotero 正在运行时的锁，但有一个副作用：如果 Zotero sqlite 处于 WAL 模式且新 item 仍在 `zotero.sqlite-wal` 中，`immutable=1` 读者可能读不到 WAL 中的最新变更。

这会和 `update_config.last_update` 的增量过滤组合成一个严重漏项问题：

1. 用户刚在 Zotero 添加新文献。
2. `update-db --fulltext` 启动，但 immutable sqlite 读不到 WAL 中的新 item。
3. 本次 update 没有索引新 item，却仍然把 `last_update` 更新为当前时间。
4. 之后 WAL checkpoint 后，新 item 在 sqlite 主库可见。
5. 但这些 item 的 `dateAdded/dateModified` 早于已经推进的 `last_update`。
6. 下一次增量更新执行 `modified_since=last_update`，这些新 item 被永久跳过，除非 force rebuild 或手动修复。

## 本地复现观察

- Zotero 中最新两篇 item key：`LQX5DUU2`, `C5HBQLCC`
- 第一次 `update-db --fulltext` 后，Chroma `last_update` 已推进，但这两个 key 不在 Chroma 中。
- 对 sqlite 做 checkpoint 后，外部只读查询才能看到 itemID `27841/27842`。
- 由于 `last_update` 已经晚于这两篇的 `dateModified`，正常增量更新仍不会捕获它们。

## 影响

这是增量索引的数据完整性问题：用户会认为“更新完成”，但刚添加的文献可能没有进入语义库，而且之后的普通增量更新也不会补回来。

## 建议修复方案

可以考虑组合修复：

### 1. 不要在可能漏读 WAL 的情况下推进 last_update

`last_update` 应该表示“已成功索引到这个时间点前所有可见变更”，而不是“命令运行结束时间”。

如果 local sqlite 使用 `immutable=1`，建议至少使用保守水位线，例如：

```text
last_update = max(dateModified/dateAdded of successfully processed items)
```

而不是 `datetime.now()`。

### 2. 增量更新使用 overlap window

即使保留当前逻辑，也建议每次增量往前回看一段时间，例如 1-24 小时：

```text
modified_since = last_update - overlap_window
```

然后依靠 Chroma upsert 去重。这可以避免 WAL checkpoint、文件同步、Zotero 写入延迟造成漏项。

### 3. 对 local Zotero sqlite 尽量读取 WAL 或触发安全 checkpoint

如果 Zotero 正在运行，不建议直接写 sqlite。但可以：

- 避免 `immutable=1` 作为唯一读取模式，尝试只读 `mode=ro` 并允许读取 WAL；
- 或文档中明确说明 Zotero 运行时 immutable read 可能读不到 WAL；
- 或通过 Zotero local API / plugin 侧触发 checkpoint / sync 后再索引。

### 4. 增量更新结束后校验最近 item

比如读取最近 N 个 Zotero item key，与 Chroma existing ids 比较。如果最近 item 不在 Chroma 中，不应静默成功推进 `last_update`。

## 相关 issue

可能与 #286 中 Windows 下 `update-db --fulltext` 增量更新问题相关，但这里的核心是 `immutable=1 + WAL + last_update` 造成的漏项。

lots-o:
## Summary

Fixes #292. The `update-db --fulltext` path reads `zotero.sqlite` with `immutable=1` to bypass Zotero's lock — but an immutable connection cannot see rows still sitting in an un-checkpointed WAL file. The API-derived `last_sync_version` captured before the scan *does* cover those rows (the running Zotero serves them from memory), so promoting it unconditionally after the scan makes every later incremental update filter them out with `since=watermark` — newly added items get skipped **permanently** unless the user force-rebuilds.

## Reproduction

Verified the sqlite-level mechanism against a WAL-mode copy of a real 112-item library, driving the project's actual `LocalZoteroReader`:

```
journal_mode -> wal
WAL size after commit: 37112 bytes
[before checkpoint] LocalZoteroReader sees 112 items, WALTEST292 found = False   ← scan misses it
[before checkpoint] mode=ro read sees WALTEST292 = True                          ← yet it is committed
[after checkpoint]  LocalZoteroReader sees 113 items, WALTEST292 found = True    ← too late: watermark already advanced
```

So during the pre-checkpoint window, a `--fulltext` run indexes everything *except* the new item, then promotes `last_sync_version` past that item's version. One note on scope: current Zotero builds hardcode `const DB_LOCK_EXCLUSIVE = true` (rollback journal + exclusive lock), where this can't happen; the bug hits setups where `zotero.sqlite` runs in WAL mode (e.g. older builds with `dbLockExclusive=false`), which matches the issue report.

## Changes

- add `LocalZoteroReader.get_all_item_keys()` — the full key set visible to the sqlite snapshot (`SELECT key FROM items`, all item types)
- add `ZoteroSemanticSearch._verify_local_snapshot_version()` — after a local-extraction scan, compare the API's `item_versions()` keys against the snapshot's keys. If the API knows keys the snapshot never saw (= un-checkpointed WAL rows), **keep the previous watermark** instead of promoting, so the next update re-covers the gap. API errors also hold the watermark (conservative).
- wire the check into `update_database()` — only on the `extract_fulltext` full-scan path; web-API ingest paths are unaffected (they read from the API directly and have no snapshot-staleness problem)
- the sqlite read itself stays `immutable=1` (it's still the right way to read while Zotero holds its lock); only the watermark promotion gains the guard

## Tests

6 new tests in `tests/test_fulltext_sync_watermark.py` (minimal fake `zotero.sqlite` fixture + pyzotero stub):

- `get_all_item_keys` returns every key
- watermark promoted when the snapshot covers all API keys
- watermark held back when the API sees a key absent from the snapshot
- watermark held back when the API call fails
- end-to-end through `update_database(extract_fulltext=True)`: stale snapshot keeps `last_sync_version` at its previous value; complete snapshot promotes it

The end-to-end stale-snapshot test fails on `main` with `assert 12 == 10` (watermark wrongly promoted), demonstrating the bug.

```
uv run pytest tests/test_fulltext_sync_watermark.py -q
6 passed
```

Full suite: 906 passed; the 5 pre-existing failures (`test_lifespan`, `test_place_field_reads`) fail identically on `main` and are unrelated.

## Notes / follow-up

This addresses the watermark-integrity half of #292. The issue also suggests an overlap window on incremental reads and a WAL-aware read mode; with the version-based `since` watermark now only advancing when the snapshot is provably complete, those become optional hardening rather than required fixes.

### DIFF
--- a/src/zotero_mcp/local_db.py
+++ b/src/zotero_mcp/local_db.py
@@ -658,6 +658,18 @@ class LocalZoteroReader:
         )
         return cursor.fetchone()[0]
 
+    def get_all_item_keys(self) -> set[str]:
+        """
+        Get the keys of every item in the database, regardless of type.
+
+        Used to verify that the sqlite snapshot is not lagging behind the
+        Zotero API (an `immutable=1` read cannot see rows that are still
+        in an un-checkpointed WAL file).
+        """
+        conn = self._get_connection()
+        rows = conn.execute("SELECT key FROM items").fetchall()
+        return {row[0] for row in rows}
+
     def get_items_with_text(self, limit: int | None = None, include_fulltext: bool = False, key_filter: str | None = None) -> list[ZoteroItem]:
         """
         Get all items with their text content for semantic search.
--- a/src/zotero_mcp/semantic_search.py
+++ b/src/zotero_mcp/semantic_search.py
@@ -123,6 +123,9 @@ class ZoteroSemanticSearch:
         self.zotero_client = get_zotero_client()
         self.config_path = config_path
         self.db_path = db_path  # CLI override for Zotero database path
+        # Item keys seen by the most recent local sqlite scan (set by
+        # _get_items_from_local_db); used to verify watermark promotion.
+        self._last_scan_snapshot_keys: set[str] | None = None
 
         # Load update configuration
         self.update_config = self._load_update_config()
@@ -482,6 +485,12 @@ class ZoteroSemanticSearch:
                 pass
 
             with suppress_stdout(), LocalZoteroReader(db_path=zotero_db_path, pdf_max_pages=pdf_max_pages, pdf_timeout=pdf_timeout) as reader:
+                # Capture the snapshot's full key set on the SAME connection
+                # this scan uses. The staleness check after the (potentially
+                # long) extraction must compare against what this scan could
+                # actually see — a fresh read taken later could already
+                # include rows from a WAL checkpoint that landed mid-scan.
+                self._last_scan_snapshot_keys = reader.get_all_item_keys()
                 # Phase 1: fetch metadata only (fast)
                 sys.stderr.write("Scanning local Zotero database for items...\n")
                 local_items = reader.get_items_with_text(limit=limit, include_fulltext=False)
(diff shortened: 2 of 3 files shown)


lots-o:
Pushed a follow-up commit (af3ea62) after re-reviewing the guard for races: the verification originally re-opened a *fresh* immutable connection after the scan. Since PDF extraction can run for minutes, a WAL checkpoint landing mid-scan would make the fresh read look complete even though the scan itself missed the item — reproducing the exact permanent-skip this PR prevents.

The guard now captures the snapshot's key set on the scan's own connection before extraction starts and verifies against that (fresh read kept only as a fallback when no scan capture exists). Two tests added: one drives the mid-scan-checkpoint scenario (fails on the previous commit with `assert 42 is None`), one pins the capture behavior.

PR #348, opened 2026-06-14
fix(semantic-search): reliable OpenRouter/Gemini embeddings, non-blocking status, config-authoritative model

### LINKED ISSUE #3: Notes Management
What a Zotero MCP! It's a great plugin for using Zotero to read references automatically. 
Could you add note management tools, such as adding notes, searching notes, or reading notes?
I deeply understand the hardship of developing tools. If possible, could you share your payment QR code?

trahloff:
## Summary

Three related robustness fixes for semantic search, surfaced while setting up **Gemini embeddings via OpenRouter** (an OpenAI-compatible gateway). Each is independent; happy to split into separate PRs if you'd prefer.

### 1. Deterministic embeddings via `encoding_format="float"`
The `openai` SDK negotiates **base64** by default. OpenRouter's Gemini embedding providers (`google/gemini-embedding-001`, `google/gemini-embedding-2`) don't return that reliably, so the SDK raises `ValueError: No embedding data received` — intermittently for `-001` (~1 in 3 calls) and *every* time for `-2`. The raw HTTP endpoint returns valid 3072-dim vectors, so it's purely the encoding negotiation.

Setting `encoding_format="float"` explicitly in `OpenAIEmbeddingFunction.__call__` fixes it: **8/8 → 100% reliable** in testing, including batches. Harmless for native OpenAI (returns float, just slightly larger over the wire) and makes behavior deterministic across any OpenAI-compatible backend.

### 2. `db-status` / `zotero_get_search_database_status` no longer loads a model or holds the API lock
The status tool went through `create_semantic_search()`, which constructs the embedding function. For the **default** backend that downloads the ~80MB ONNX MiniLM model on first use — turning a read-only row count into a multi-minute (or network-blocked, indefinite) hang. Worse, the tool held the shared `@with_zotero_api_lock`, so a slow status read wedged every other Zotero tool until the client timed out.

This adds `read_collection_status()` — a lightweight reader that opens the persisted collection and reads `.count()` **without** building any embedding function (uses a no-op EF so ChromaDB doesn't reconstruct the persisted one) — plus module-level `load_update_config()` / `should_update()` helpers (pure config, no I/O). The tool now uses these and **drops the API lock** (it never touches the Zotero API).

### 3. `config.json` is authoritative over a stale `ZOTERO_EMBEDDING_MODEL` env var
`create_chroma_client` let the env var unconditionally override `config.json`'s model. Claude Desktop passes its server `env` block on every launch (and can rewrite that file itself), so a leftover `ZOTERO_EMBEDDING_MODEL=default` there silently downgraded an explicitly configured Gemini/OpenAI setup back to the local model — which then failed every query with an opaque `Collection expecting embedding with dimension of 3072, got 384`.

The env var now only fills the gap when `config.json` leaves the model at the `"default"` placeholder (the normal Claude Desktop case), so the env path still works while an explicit config wins. This is consistent with the existing "config.json wins, env fills gaps" rule already applied to `embedding_config`.

## Testing
- Full suite: **897 passed**; the only failures are pre-existing `test_place_field_reads` cases unrelated to these changes (present on a clean tree).
- Manually verified end-to-end against a real library: rebuilt a 447-doc / 3072-dim index on `google/gemini-embedding-001` via OpenRouter with 0 errors, and confirmed semantic queries return relevant results — including under a simulated stale `ZOTERO_EMBEDDING_MODEL=default` env (fix #3).

## Notes
- No new dependencies. No public API changes.
- Behavior change worth calling out: fix #3 changes env-vs-file precedence for `ZOTERO_EMBEDDING_MODEL`. Only affects setups that set a non-default model in `config.json` *and* a different one via the env var — previously the env won, now the file does.

### DIFF
--- a/src/zotero_mcp/chroma_client.py
+++ b/src/zotero_mcp/chroma_client.py
@@ -73,10 +73,19 @@ class OpenAIEmbeddingFunction(EmbeddingFunction):
         )
 
     def __call__(self, input: Documents) -> Embeddings:
-        """Generate embeddings using OpenAI API."""
+        """Generate embeddings using the OpenAI-compatible API.
+
+        ``encoding_format="float"`` is set explicitly. The OpenAI SDK otherwise
+        negotiates base64 by default, which OpenRouter's Gemini embedding
+        providers (e.g. ``google/gemini-embedding-001``) do not return reliably —
+        the SDK then raises "No embedding data received" intermittently. Forcing
+        float makes every OpenAI-compatible backend, native OpenAI included,
+        respond deterministically.
+        """
         response = self.client.embeddings.create(
             model=self.model_name,
-            input=input
+            input=input,
+            encoding_format="float",
         )
         return [data.embedding for data in response.data]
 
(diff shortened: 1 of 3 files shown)


trahloff:
Heads-up: I've opened #350 (passage-level retrieval, OpenAlex citation-graph discovery, annotation synthesis + bibliography export, MCP prompts & resources) which is **stacked on top of this PR** — its first commit is the reliability work here. Suggested order: merge this one first, then #350 rebases down to just the feature commit. Happy to reorganize if you'd prefer them combined or split differently.

PR #353, opened 2026-06-20
fix(scite): send bare list payload to POST /papers batch endpoint

### LINKED ISSUE #352: get_papers_batch POSTs {"dois":[...]} to /papers, but API expects a bare list → scite_check_retractions always fails
## Summary

`scite_check_retractions` (and any tool that calls `get_papers_batch`) always reports **"Could not reach Scite API"**, even when `api.scite.ai` is fully reachable and the single-item `scite_enrich_item` works fine.

## Root cause

In `zotero_mcp/scite_client.py`, `get_papers_batch` POSTs to `/papers` wrapping the DOI list in a dict:

```python
resp = requests.post(
    f"{_BASE}/papers",
    json={"dois": dois[:500]},   # ← dict payload
    ...
)
```

But the Scite `/papers` batch endpoint expects a **bare JSON array** — exactly like the sibling `/tallies` batch endpoint that `get_tallies_batch` already calls correctly with `json=dois[:500]`. The dict payload makes Scite return HTTP 400:

```
{"message":"1 validation error:
  {'type': 'list_type', 'loc': ('body',),
   'msg': 'Input should be a valid list',
   'input': {'dois': ['10.1038/s41586-021-03819-2']}}"}
```

Because the response is non-200, `get_papers_batch` returns `{}`, and the caller surfaces it as "Could not reach Scite API".

## Impact

- `scite_check_retractions` (collection / tag / recent retraction scanning) is **completely broken**.
- Any code path relying on `get_papers_batch` is affected.
- `scite_enrich_item` is unaffected (uses GET `/papers/{doi}`).

## Reproduce

```bash
# Current behavior (dict) → 400
curl -X POST -H "Content-Type: application/json" \
  -d '{"dois":["10.1038/s41586-021-03819-2"]}' \
  https://api.scite.ai/papers

# Expected (bare list) → 200 {"papers": {...}}
curl -X POST -H "Content-Type: application/json" \
  -d '["10.1038/s41586-021-03819-2"]' \
  https://api.scite.ai/papers
```

## Suggested fix

```python
# zotero_mcp/scite_client.py — get_papers_batch
json=dois[:500],   # was: json={"dois": dois[:500]}
```

Verified locally: with the fix, `get_papers_batch` returns the expected `{"papers": {doi: {...}}}` map including `editorialNotices`.

## Environment

- zotero-mcp-server 0.5.0
- macOS 15, Python 3.13.2

Fangziyang0910:
## Summary

`scite_check_retractions` (and any tool using `get_papers_batch`) always reports **"Could not reach Scite API"**, even though `api.scite.ai` is reachable and the single-item `scite_enrich_item` works fine.

## Root cause

`get_papers_batch` POSTs `{"dois": [...]}` to `/papers`, but the endpoint expects a **bare JSON array** — the same shape the sibling `/tallies` batch endpoint already uses correctly. The dict payload triggers HTTP 400:

```
{"message":"1 validation error: {'type': 'list_type', 'loc': ('body',), 'msg': 'Input should be a valid list', 'input': {'dois': [...]}}"}
```

The non-200 response makes `get_papers_batch` return `{}`, surfaced upstream as "Could not reach Scite API".

## Fix

```diff
- json={"dois": dois[:500]},
+ json=dois[:500],
```

Single-item `scite_enrich_item` is unaffected (it uses GET `/papers/{doi}`).

## Testing

- Added `tests/test_scite_client.py` (4 tests): bare-list payload for `/papers`, non-200 handling, empty-input short-circuit, and a regression guard for the `/tallies` endpoint.
- All 4 new tests pass.
- Full suite: **900 passed, 3 failed** — the 3 failures are pre-existing and unrelated (`test_pdf_cascade` needs network; `test_place_field_reads` is a bibtexparser version issue). Confirmed by re-running them with this change stashed.

Fixes #352

### DIFF
--- a/src/zotero_mcp/scite_client.py
+++ b/src/zotero_mcp/scite_client.py
@@ -105,7 +105,7 @@ def get_papers_batch(dois: list[str]) -> dict[str, dict]:
     try:
         resp = requests.post(
             f"{_BASE}/papers",
-            json={"dois": dois[:500]},
+            json=dois[:500],
             headers=_HEADERS,
             timeout=_TIMEOUT,
         )
--- a/tests/test_scite_client.py
+++ b/tests/test_scite_client.py
@@ -1,0 +1,61 @@
+"""Tests for the Scite public API client."""
+
+from unittest.mock import MagicMock, patch
+
+from zotero_mcp import scite_client
+
+
+def _ok_response(payload):
+    resp = MagicMock()
+    resp.status_code = 200
+    resp.json.return_value = payload
+    return resp
+
+
+def test_get_papers_batch_sends_bare_list_payload():
+    """get_papers_batch must POST a bare JSON list, not {"dois": [...]}.
+
+    The Scite /papers batch endpoint returns HTTP 400 for a dict-wrapped
+    payload, which surfaced upstream as "Could not reach Scite API" from
+    scite_check_retractions. Regression test for
+    https://github.com/54yyyu/zotero-mcp/issues/352
+    """
+    with patch.object(
+        scite_client.requests,
+        "post",
+        return_value=_ok_response({"papers": {"10.1038/x": {"title": "T"}}}),
+    ) as mock_post:
+        result = scite_client.get_papers_batch(["10.1038/x"])
+
+    _, kwargs = mock_post.call_args
+    # Bare list, NOT {"dois": [...]}
+    assert kwargs["json"] == ["10.1038/x"]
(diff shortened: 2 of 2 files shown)


54yyyu:
Superseded by #331, which makes the same POST /papers body-shape fix and also adds case-insensitive DOI matching so uppercase DOIs don't slip past the retraction check. Thanks for the fix here, closing in favor of that one.

PR #361, opened 2026-06-23
attach via WebDAV only when configured, skip cloud /file upload

lukas-blecher:
Skip Zotero cloud file upload when WebDAV is configured

`attachment_both` creates an attachment item and uploads the file bytes to Zotero Storage in one call. For users with File Syncing set to WebDAV, this is the wrong target. It also fails with HTTP 413 once the free-tier 300 MB quota is full.

This PR adds `_webdav_first_attach`, called before attachment_both at every PDF/file attach site (`zotero_add_by_doi`, `zotero_add_by_url`, `zotero_add_from_file`, `_add_by_arxiv`). When WebDAV is configured it:

1. Creates the attachment shell via create_items (no file upload to cloud)
2. PUTs the bytes directly to WebDAV as <key>.zip + <key>.prop
3. Returns a user-facing suffix so the caller can report success or a warning

When WebDAV is not configured it returns None and the caller falls through to the existing attachment_both + `_maybe_upload_to_webdav` path unchanged.

if the WebDAV PUT fails after the shell is created, the attachment item is left in Zotero with no file bytes. Cleanup on failure is not implemented right now.

### DIFF
--- a/src/zotero_mcp/tools/_helpers.py
+++ b/src/zotero_mcp/tools/_helpers.py
@@ -805,11 +805,13 @@ def _download_and_attach_pdf(write_zot, item_key, pdf_url, doi, ctx):
                 ctx.info("Downloaded file too small, likely not a real PDF")
                 return None
 
+            suffix = _webdav_first_attach(write_zot, filename, filepath, item_key, ctx)
+            if suffix is not None:
+                return suffix
             attach_result = write_zot.attachment_both(
                 [(filename, filepath)],
                 parentid=item_key,
             )
-            # Must run inside the with-block — temp file disappears on exit.
             return _maybe_upload_to_webdav(attach_result, filepath, ctx)
     except Exception as e:
         ctx.info(f"PDF download/attach failed: {e}")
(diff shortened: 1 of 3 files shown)


54yyyu (changes requested):
Thanks, this targets a real problem: attachment_both uploads to Zotero Storage, which is the wrong target for WebDAV-sync users and 413s once the free 300 MB quota fills. The shape is right too, with the unconfigured path falling through to the existing flow untouched. Three things before merge:

1. **Cleanup on failure.** As you note, a failed WebDAV PUT currently strands an attachment shell with no file bytes. Wrap the PUT so that on failure the shell is deleted (`write_zot.delete_item` on the fetched attachment), falling back to the warning suffix only if the delete itself fails. An orphaned no-bytes attachment is confusing in the Zotero UI and breaks sync expectations.

2. **Set contentType on the shell.** The template only sets title/filename/parentItem. Set `contentType` (`application/pdf` at the PDF call sites; guess from extension in `add_from_file`) so the item renders and opens correctly in Zotero.

3. **Tests.** None yet, and this touches four attach sites. Please add coverage in the style of the existing WebDAV tests: (a) WebDAV unconfigured returns None and the caller falls through to attachment_both, (b) configured path creates the shell and PUTs via upload_attachment_to_webdav with the right key, (c) PUT failure triggers cleanup and returns the warning suffix.

With those three addressed I'm happy to merge.

54yyyu:
Thanks, skipping the cloud /file upload when WebDAV is configured is the right behavior, and the 413-at-quota failure mode is real.

The blocker for merging as-is is the acknowledged cleanup gap: if the WebDAV PUT fails after `create_items`, the user is left with an attachment item in Zotero that has no file bytes, and nothing tells them to delete it. Could you wrap the PUT so that on failure the just-created attachment item is deleted (best effort, with the original error still surfaced)? With that, plus a test for the failure path, this looks mergeable.

lukas-blecher:
thanks for the review. lmk if this looks good now

PR #367, opened 2026-07-04
Unify zotero_db_path resolution

AndreiPashkin:
## Problem

Several local-mode tools (`list_libraries`, `validate_library_switch`, `list_feeds`, `get_feed_items`, `get_local_zotero_reader`, `read_pdf`) ignore the configured `zotero_db_path` and rely on auto-detection only. This causes them to read the wrong database when the user's Zotero profile uses a non-default DB location.

## Changes

- Refactor config-parsing to use Pydantic for validation and expose values as a nested dataclass.
- Replace raw-dict `_load_zotero_mcp_config()` in `_helpers.py` with typed `load_config()` returning `ZoteroMcpConfig`
- Add top-level `zotero_db_path` config setting with `resolve_zotero_db_path()` method (checks top-level first, falls back to `semantic_search.zotero_db_path`)
- Make all six inconsistent tool call sites to pass the resolved db_path to `LocalZoteroReader`
- Update `get_item_fulltext` and `get_attachment_path` to use the same resolution
- Replace manual config file parsing in `read_pdf.py` with `load_config()`
- Make the setup wizard and the CLI option respect the new setting.
- Add tests for config parsing edge cases (null values, extra keys, type coercion, malformed JSON) and resolution precedence

## Backward compatibility

- Existing configs with `semantic_search.zotero_db_path` continue to work unchanged
- Extra/unknown keys in config JSON are silently ignored
- null values for nested objects are treated as absent (defaults fill in)
- Non-coercible type errors fall back to all-defaults with a log warning

### DIFF
--- a/src/zotero_mcp/cli.py
+++ b/src/zotero_mcp/cli.py
@@ -122,12 +122,8 @@ def _save_zotero_db_path_to_config(config_path: Path, db_path: str) -> None:
             except Exception:
                 pass
 
-        # Ensure semantic_search section exists
-        if "semantic_search" not in full_config:
-            full_config["semantic_search"] = {}
-
-        # Save the db_path
-        full_config["semantic_search"]["zotero_db_path"] = db_path
+        # Save the db_path at the top level
+        full_config["zotero_db_path"] = db_path
 
         # Write back to file
         with open(config_path, 'w') as f:
--- a/src/zotero_mcp/config.py
+++ b/src/zotero_mcp/config.py
@@ -1,0 +1,124 @@
+"""Typed configuration for zotero-mcp.
+
+The server configuration lives at ``~/.config/zotero-mcp/config.json``.
+This module defines stdlib dataclasses for every section and a single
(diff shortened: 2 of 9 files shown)


54yyyu:
Thanks, this is a solid direction. The typed config layer and unifying db-path resolution across the six inconsistent call sites is something we want.

Two things before this can land:

1. It currently breaks an existing test on main: `setup_semantic_search` now returns a tuple, but `tests/test_openai_batch.py::test_setup_openai_new_config_defaults_to_batch` still indexes the result as a dict (`config["embedding_model"]`). Merging your branch into latest main gives `TypeError: tuple indices must be integers or slices, not str`.
2. #379 just merged and touches the same territory from the other side: `_find_zotero_db()` now implements `ZOTERO_DB_PATH` and reads the data dir from the Zotero profile's prefs.js. Your `resolve_zotero_db_path()` returning None so auto-detect kicks in should compose fine with it, but please rebase onto latest main and double check the precedence story (explicit arg > env var > config > prefs.js > default) ends up coherent.

Happy to merge after a rebase with the test fixed.

PR #370, opened 2026-07-06
Add opt-in collection filter for the semantic search database

JohanVisser97:
## Summary

Adds an **opt-in collection filter** for the semantic search database. When `semantic_search.collection_keys` is set in `config.json`, the database is built from only those collections — resolved recursively, so subcollections count — instead of the whole library:

```json
"semantic_search": {
  "collection_keys": ["ABCD1234"]
}
```

Unset (the default), behaviour is completely unchanged.

## Motivation

Large personal libraries often mix one project's literature with teaching materials, feeds and unrelated topics. For focused semantic search (and to keep embedding costs down), you want the vector database to contain just one collection. I've been running this patch for months to keep a PhD-thesis collection (~535 items) searchable inside a much larger library.

## Implementation

- `LocalZoteroReader.get_items_with_text()` gains an optional `collection_keys` parameter that restricts the SQL query to items in the given collections or any of their subcollections (recursive resolution via `parentCollectionID`).
- `_get_items_from_source()` reads `collection_keys` from the config file and passes it to the scan (with a stderr note so runs are self-explaining).
- `update_database()` skips the API-based incremental path when the filter is active: that path fetches changed items from the whole library and its deletion pass compares stored ids against **all** library keys, both of which would bypass the filter. The local full-scan path applies the filter and skips already-indexed items, so filtered updates stay cheap.

## Test plan

- With `collection_keys` set: full rebuild indexes exactly the items in the collection + subcollections (535 of ~1,300 library items); items outside the collection never enter the database; subsequent `update-db --fulltext` runs report the unchanged items as up to date and only process new ones.
- Without `collection_keys`: scan returns the whole library and the incremental path is used exactly as before.

Related: #369 (bugfixes found while running this setup with chunking enabled).

### DIFF
--- a/src/zotero_mcp/local_db.py
+++ b/src/zotero_mcp/local_db.py
@@ -670,12 +670,15 @@ class LocalZoteroReader:
         rows = conn.execute("SELECT key FROM items").fetchall()
         return {row[0] for row in rows}
 
-    def get_items_with_text(self, limit: int | None = None, include_fulltext: bool = False, key_filter: str | None = None) -> list[ZoteroItem]:
+    def get_items_with_text(self, limit: int | None = None, include_fulltext: bool = False, key_filter: str | None = None, collection_keys: list[str] | None = None) -> list[ZoteroItem]:
         """
         Get all items with their text content for semantic search.
 
         Args:
             limit: Optional limit on number of items to return.
+            collection_keys: Optional list of collection keys; when set, only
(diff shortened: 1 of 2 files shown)


54yyyu:
Merged, thanks! Opt-in and recursive subcollection resolution is exactly the right shape. Small follow-up in 17ee49c: the FakeReader test double from #373 needed the new collection_keys kwarg.

PR #373, opened 2026-07-08
fix: retry failed fulltext extraction when the attachment set changes

josk0:
## Problem

Items that are indexed while they have no readable attachment get `has_fulltext: "failed"` in their ChromaDB metadata, so the update scan doesn't re-try extraction on every run. The only signal that cleared this marker was a change in the parent item's `dateModified`.

But **attaching a file does not bump the parent's `dateModified`** — an attachment is its own child item in Zotero's data model, and the parent row in `zotero.sqlite` is untouched. This is true both for PDFs attached by hand in the Zotero desktop UI and for programmatic attachment via the API.

Net effect: any item first indexed metadata-only was **permanently locked out of full-text indexing**. Attaching a PDF later did nothing; `update-db --fulltext` reported "Total items: 0" (the skipped items are filtered out before stats are counted), and semantic search kept matching those items on their title/author line only. In my library, 913 of ~30k indexed items carried a stale "failed" marker.

## Fix

- The local-mode extraction scan computes each item's attachment-key set (via the existing public `reader.get_fulltext_meta_for_item`, sorted and comma-joined) and persists it in the Chroma metadata as `attachment_keys`.
- A "failed" item is retried when **either** the parent's `dateModified` **or** the attachment set changed. The `dateModified` leg still covers the original "user replaces a bad PDF" case.
- Legacy records (written before `attachment_keys` existed) retry once, then converge: genuinely broken PDFs settle back to "failed" with the set recorded, and are not retried again until their attachments actually change.
- `local_has_fulltext` now reuses the already-computed attachment set (`bool(att_keys)`) instead of issuing a second identical SQLite query per item.
- `get_fulltext_meta_for_item`'s return annotation was wrong (`tuple[str, str] | None`; it returns a list of `[key, path, content_type]` rows) — fixed since the scan now relies on that shape.
- The "extraction previously failed" skip summary now mentions that attaching/replacing the PDF retries the item (previously it only suggested `--force-rebuild`).

## Tests

`tests/test_failed_marker_retry.py` (new, 4 tests):
- failed marker + same date + same attachment set → no retry (skip preserved)
- failed marker + same date + attachment added → retried, fulltext extracted, new set recorded
- failed marker + changed date + same attachment set → retried (pins the pre-existing `dateModified` trigger)
- legacy failed record without `attachment_keys` → retries once

Each test fails on the pre-fix code. Full suite passes (1069 tests).

Verified end-to-end on a real library: after the fix, `ZOTERO_LOCAL=true zotero-mcp update-db --fulltext` retried all 912 legacy "failed" items in ~47s with 0 errors, and semantic search returns deep-passage matches from the newly attached PDFs (`has_fulltext: True, source: pdf`).

## Scope / known limitations

- Local mode (`ZOTERO_LOCAL=true`) only: `attachment_keys` is recorded on the local-scan path. The web-API incremental path (`item_versions(since=V)`) discovers changed top-level items and may have the same lockout for web-mode users — left as a follow-up since it needs separate investigation.
- Metadata-only update runs don't set `attachmentKeys`, so an interleaved metadata-only upsert can drop the field; the item then just retries once on the next fulltext run and re-converges (same pre-existing behavior as `has_fulltext` itself).
- A file replaced behind the *same* attachment key without a `dateModified` bump is still not detected (pre-existing). The fingerprint could later be extended to include the attachment path, but paths can contain arbitrary filename characters, so that needs its own encoding care — kept out of this fix.

### DIFF
--- a/src/zotero_mcp/local_db.py
+++ b/src/zotero_mcp/local_db.py
@@ -776,8 +776,9 @@ class LocalZoteroReader:
 
         return items
 
-    # Public helper to quickly check full text metadata for item
-    def get_fulltext_meta_for_item(self, item_id: int) -> tuple[str, str] | None:
+    # Public helper to quickly check full text metadata for item.
+    # Returns one [key, path, content_type] row per attachment of the item.
+    def get_fulltext_meta_for_item(self, item_id: int) -> list[list[str | None]]:
         return self._get_fulltext_meta_for_item(item_id)
 
     # Public helper to extract fulltext on demand for a specific item
--- a/src/zotero_mcp/semantic_search.py
+++ b/src/zotero_mcp/semantic_search.py
@@ -683,6 +683,12 @@ class ZoteroSemanticSearch:
             # Mark so we don't retry on every incremental update
             metadata["has_fulltext"] = "failed"
 
+        # Record the attachment-key set (local mode only) so update runs can
+        # retry a "failed" item once its attachments change — attaching a file
+        # does not bump the parent's dateModified.
+        if (att_keys := data.get("attachmentKeys")) is not None:
+            metadata["attachment_keys"] = att_keys
+
         # Add tags as a single string
         if tags := data.get("tags"):
             metadata["tags"] = " ".join([tag.get("tag", "") for tag in tags])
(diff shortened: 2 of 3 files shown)


54yyyu:
Merged, thanks! The attachment-set fingerprint is the right trigger, and the test coverage on the legacy-record convergence path is appreciated. Noted the web-API follow-up in the tracker.

Issue #277, opened 2026-05-11
Write mode in local only is not available


00sapo:
When trying, I get:

```
# zotero_zotero_add_from_file [file_path=..., title="my title"]
Cannot perform write operations in local-only mode. Add ZOTERO_API_KEY and ZOTERO_LIBRARY_ID to enable hybrid mode.
```

Even with the env variable setups, it can't write.

My config:

```json
    "zotero": {
      "enabled": false,
      "type": "local",
      "command": ["zotero-mcp"],
      "environment": {
        "ZOTERO_LOCAL": "true",
        "ZOTERO_EMBEDDING_MODEL": "default",
        "ZOTERO_API_KEY": "...",
        "ZOTERO_LIBRARY_ID": "1"
      }
    },
```

rahit:
Can confirm this. Trying to use in claude, hitting blocks on write. Reads are fine.

6801318d8d:
I'm also hitting this and I'm unable to write nor with `zotero-cli` nor with Claude.

How to solve this?

@54yyyu

54yyyu:
For local-mode users: the Zotero local API is read-only, so writes require web-API credentials. Set `ZOTERO_API_KEY` + `ZOTERO_LIBRARY_ID` alongside `ZOTERO_LOCAL=true` to enable hybrid mode (local reads, web-API writes). #308 (merged) adds setup docs for this. A clearer up-front error when write creds are missing is still worth doing — leaving open for that.

LordEliasTM:
Also hit this today. Why is writing via local api not possible?

54yyyu:
Following up, since this has three people on it and my earlier reply explained the constraint without fixing anything.

To restate it plainly: Zotero's local HTTP API is read-only. There are no write endpoints and no attachment upload endpoints, so this isn't something zotero-mcp can enable. Writes need web API credentials, which is what hybrid mode is: `ZOTERO_LOCAL=true` for fast local reads, plus `ZOTERO_API_KEY` and `ZOTERO_LIBRARY_ID` for writes.

[...]

PR #374, opened 2026-07-09
Enhance README with local write operations tip

raffaelemancuso:
Added tip for local write operations using zotero_mcp_server_write.

### DIFF
--- a/README.md
+++ b/README.md
@@ -288,6 +288,8 @@ After installation, either:
    > absolute path instead (run `zotero-mcp setup-info` or `which zotero-mcp` to find it) —
    > GUI apps don't always inherit your shell `PATH`.
 
+   > **Tip:** For _local_ write operations, you can try the [zotero_mcp_server_write](https://github.com/akchan/zotero_mcp_server_write)
+
 #### Usage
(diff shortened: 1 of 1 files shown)


54yyyu:
Thanks for the pointer. I'm hesitant to recommend a third-party server that writes to users' libraries directly in the README without having vetted it, especially since local write support is something we're tracking for zotero-mcp itself (see #277). Leaving this open for now as a reference for others who need local writes today; if the linked server matures and picks up users, happy to revisit.

54yyyu:
Closing this, with the same reasoning as before rather than anything new.

I'm not comfortable putting a recommendation for a third-party server that writes to people's libraries in our README without having vetted it. The README is where users go to decide what to trust, and a write-capable server is the wrong place to be casual about that. Local write support for zotero-mcp itself is tracked in #277, which is where I'd rather the effort go.

The PR staying open for three weeks hasn't served anyone, so I'll close it. It stays visible and linked here for anyone searching for local write options today, and if the linked server matures I'm happy to revisit.

Thanks for taking the time to send it.

PR #375, opened 2026-07-10
docs: add Autohand Code MCP setup

igorcosta:
## What changed

Adds an Autohand Code setup example after the existing Claude client instructions. It uses Zotero MCP's installed `zotero-mcp` entry point with the documented `ZOTERO_LOCAL=true` read-only configuration.

The note points hybrid and web API users back to the existing credential guidance and includes the project-scoped option.

## Verification

- Checked the stdio command against the current Autohand Code MCP CLI
- Ran `git diff --check`

This is a documentation-only change.


54yyyu:
Merged, thanks for the docs!

Issue #68, opened 2025-10-02
Adding individual zotero.sqlite location


ypsilonkah:
Hi,
I want to use fulltext database search (`zotero-mcp update-db --fulltext`) but face the following error:

`Error reading from local database: Zotero database not found at C:\Users\USER\Documents and Settings\USER\Zotero\zotero.sqlite`

This is correct, since at this location, no Zotero DB is saved. How can I set an individual localtion?

xinggue:
I see this issue was post on oct 2. If you still haven't solved it yet:

"Use your custom zotero.sqlite path"
```
zotero-mcp update-db --fulltext --db-path "/Your_custom_path/zotero.sqlite"
```
hopes it works for you😘

PR #379, opened 2026-07-12
fix: auto-detect custom Zotero data directory from profile prefs.js (implements documented ZOTERO_DB_PATH)

danmackinlay:
## Problem

`LocalZoteroReader._find_zotero_db()` claims to "auto-detect the Zotero database location based on OS", but on every OS it only checks the hardcoded default `~/Zotero/zotero.sqlite`. A relocated data directory — a first-class Zotero setting (`extensions.zotero.dataDir`, set via *Preferences → Advanced → Files and Folders*) — is never consulted, so every local-mode feature (`zotero_get_item_fulltext`, `zotero_get_attachment_path`, `read_pdf`, semantic-search indexing) fails for those users with:

```
Zotero database not found at /Users/<user>/Zotero/zotero.sqlite.
Please ensure Zotero is installed and has been run at least once.
```

The message is doubly misleading: Zotero *is* installed and running (the same server happily talks to it over the local API), and the message doesn't mention either remedy. #68 was this exact failure; it was resolved with the manual `--db-path` workaround rather than fixing detection.

Two related defects share the root cause:

1. **`ZOTERO_DB_PATH` is documented but not implemented.** The README lists it under environment variables ("Custom `zotero.sqlite` path"), but nothing in `src/` reads it — setting it silently does nothing.
2. **`_get_base_attachment_path()` reads `prefs.js` from the wrong directory.** It looks for `prefs.js` next to `zotero.sqlite` (the data directory), but Zotero keeps `prefs.js` in the *profile* directory. On standard installs the file is never found, so `extensions.zotero.baseAttachmentPath` is never read and `attachments:` linked files can't resolve.

## Fix

`_find_zotero_db()` now resolves in this order:

1. **`ZOTERO_DB_PATH` env var** — implements the documented interface; raises a targeted error if set but pointing at a missing file.
2. **Data directories declared in Zotero profile `prefs.js`** (`extensions.zotero.dataDir`), searched in the OS profile locations:
   - macOS: `~/Library/Application Support/Zotero/Profiles/*/prefs.js`
   - Windows: `%APPDATA%\Zotero\Zotero\Profiles\*\prefs.js`
   - Linux: `~/.zotero/zotero/*/prefs.js`
   A directory configured in Zotero's own preferences is authoritative, so it is checked before the default (a leftover `~/Zotero` from an old install shouldn't shadow the real library).
3. **The `~/Zotero` default**, plus the legacy XP path on Windows (unchanged behavior for the common case).

If nothing is found, the error now lists every path checked and both remedies (`ZOTERO_DB_PATH` or `zotero-mcp setup`).

`_get_base_attachment_path()` now also checks the profile `prefs.js` files (keeping the data-dir check first for unusual setups), via shared helpers. Preference values are unescaped as JS string literals so Windows paths (`C:\\Users\\...`) parse correctly — the previous regex would have returned the raw escaped string.

Existing precedence is untouched: an explicit `db_path` argument (from `--db-path` or `semantic_search.zotero_db_path` config) still wins over everything, since `_find_zotero_db()` only runs when no path was supplied.

## Testing

- `tests/test_local_db_discovery.py` (9 new tests): env-var override, env-var-points-at-nothing error, default location, prefs.js `dataDir` detection, prefs-beats-stale-default precedence, error-message content, Windows path unescaping, absent-pref handling, `baseAttachmentPath` from profile prefs.
- Full suite: **1074 passed** (no regressions).
- Verified on a real library with a relocated data directory (`~/Documents/Zotero`, macOS): auto-discovery finds the database with no env var and no config — previously an immediate `FileNotFoundError` — and reads 37k items.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -17,3 +17,8 @@ ZOTERO_LIBRARY_TYPE=user
 
 # Google Gemini API Key (可选)
 # GEMINI_API_KEY=your_gemini_key_here
+
+# 自定义 zotero.sqlite 路径 (可选; 本地模式)
+# Custom zotero.sqlite path (optional; local mode). When unset, the database
+# is auto-detected from Zotero's configured data directory, then ~/Zotero.
+# ZOTERO_DB_PATH=/path/to/Zotero/zotero.sqlite
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- **Local database auto-discovery now honors a custom Zotero data directory** — the `extensions.zotero.dataDir` preference is read from the Zotero profile's `prefs.js` (macOS/Windows/Linux), so relocated data directories no longer fail with "Zotero database not found at ~/Zotero/zotero.sqlite" (#68). The `ZOTERO_DB_PATH` environment variable, documented in the README but previously unimplemented, now works as an override, and the not-found error lists every location checked plus how to fix it. `extensions.zotero.baseAttachmentPath` is likewise read from the profile's `prefs.js`, fixing resolution of linked attachments relative to a base directory.
+
 ## [0.6.1] - 2026-07-03
 
 ### Fixed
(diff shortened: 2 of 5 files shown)


AndreiPashkin:
A bit related to #367

danmackinlay:
Good point @AndreiPashkin ; AFAICS we need both, right?

54yyyu:
Merged, thanks! prefs.js discovery plus finally implementing the documented ZOTERO_DB_PATH closes a long-standing gap (#68). I resolved a small CHANGELOG conflict with #380 during the merge.

Issue #212, opened 2026-04-17
Perf: update-db re-embeds unchanged items


lucasgfsvd:
## Problem
On a 2500-item library, `zotero-mcp update-db --fulltext` takes noticeably longer than it should. Looks like it re-embeds items whose content hasn't changed.

## Proposed fix
- Track per-item content hash (title + abstract + fulltext sha) in the ChromaDB metadata
- Skip embedding when hash matches existing record
- `--force-rebuild` already exists for the "re-embed everything" case

## Impact
Turns a 10-minute refresh into seconds for typical add-a-few-papers workflows. Also enables realistic `on_change` auto-update (see companion issue).

Happy to PR.

54yyyu:
fixed in #260 - incremental sync now uses since-based versioning to skip unchanged items

PR #380, opened 2026-07-13
fix(semantic): chunked items are never recognized as indexed, so every fulltext update re-embeds the whole library

finnrogosch:
## Problem

With `semantic_search.chunking` enabled (#350), every fulltext update re-extracts all PDFs and re-embeds the entire library, even when nothing changed. There is no error, just a long CPU-heavy run every time.

Chunked items are indexed **only** under their chunk ids (`<key>#<n>`), never under the bare item key. In my own library, all 17,245 embeddings carry a chunk id and not one is a bare key:

```
sample ids: ['238MSPGL#0', '25DLIMLN#0', '25DLIMLN#1', '25DLIMLN#10', ...]
total: 17245 | chunk ids: 17245 | bare keys: 0
```

The "is this item already indexed?" check in `_get_items_from_local_db` (semantic_search.py:964) looks the item up by its bare key:

```python
existing_metadata = chroma_client.get_document_metadata(it.key)
```

and `get_document_metadata` does an exact-id `collection.get(ids=[doc_id])`. For a chunked item that never matches, so `existing_metadata` is `None`, every item counts as new, `skipped_existing` never increments, and the run re-extracts and re-embeds everything.

The since-based incremental sync from #260 does not save this path. It is gated on `not extract_fulltext` (semantic_search.py:1567):

```python
use_incremental = (
    not force_full_rebuild and not extract_fulltext and limit is None and last_sync_version > 0
)
```

and local-mode auto-update always passes `extract_fulltext=is_local_mode()` (tools/search.py:50). So for a local-mode user with chunking on, this per-item check is the only thing standing between them and a full rebuild, and chunking silently disabled it.

This is not a duplicate of #212: that was closed by #260, which fixed the web-API path. Chunking (48a2200, 2026-06-18) landed after #260 (2026-05-19) and reintroduced the same symptom for the local-fulltext path.

## Fix

Chunk 0 carries the item-level metadata the check reads (`date_modified`, `has_fulltext`, `item_key`), so `get_document_metadata` falls back to it when the bare key misses:

```python
result = self.collection.get(ids=[doc_id, f"{doc_id}#0"], include=["metadatas"])
```

Three lines, one call site (`semantic_search.py:964` is the only caller). Unchunked (item-level) records keep working unchanged, since the bare key is still tried.

Verified against my real ChromaDB:

```
before (bare key):     []                -> MISS, item counted as new
after  (with #0):      ['238MSPGL#0']
  date_modified = 2025-05-07T09:58:54Z
  has_fulltext  = failed
```

## Tests

`tests/test_chunked_metadata_lookup.py` covers the chunked lookup, the unchunked lookup, and the not-indexed case. The chunked test fails on `main` and passes with the fix:

```
# on main
FAILED tests/test_chunked_metadata_lookup.py::test_finds_chunked_item_by_bare_key
1 failed, 2 passed

# with fix
3 passed
```

Related existing suites still pass (`test_fulltext_local_mode`, `test_fulltext_web_mode`, `test_fulltext_sync_watermark`, `test_semantic_stats`): 38 passed.

## Note for affected users

Because the broken check made every run look like a full rebuild, the symptom for anyone on chunking plus local mode is a machine that spins up an embedding run on every sync. Worth a line in the release notes, since the config that triggers it (`chunking.enabled: true`) is the one #350 recommends for grounded retrieval.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- **Chunking no longer forces a full re-extract and re-embed of the whole library on every update.** With `semantic_search.chunking` enabled (#350), items are indexed only under their chunk ids (`<key>#<n>`), but the "already indexed?" check in the local-fulltext path looked each item up by its bare key. That exact-id lookup never matched, so every item counted as new: each fulltext update re-extracted every PDF and re-embedded the entire library, silently and without an error. `get_document_metadata` now falls back to chunk 0, which carries the item-level `date_modified` and `has_fulltext` that the check needs, so unchanged items are skipped again.
+
 ## [0.6.1] - 2026-07-03
 
 ### Fixed
--- a/src/zotero_mcp/chroma_client.py
+++ b/src/zotero_mcp/chroma_client.py
@@ -795,16 +795,22 @@ class ChromaClient:
 
     def get_document_metadata(self, doc_id: str) -> dict[str, Any] | None:
         """
-        Get metadata for a document if it exists.
+        Get metadata for an item if it is indexed.
+
+        With passage chunking enabled, an item is stored only under its chunk
+        ids (``<key>#<n>``) and never under the bare item key, so an exact-id
+        lookup on the key alone misses every chunked item. Chunk 0 carries the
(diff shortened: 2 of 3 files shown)


54yyyu:
Merged, thanks for the exceptionally thorough writeup and the regression test. This will be called out in the 0.6.2 release notes since it hits everyone running chunking in local mode.

PR #381, opened 2026-07-13
fix(annotations): map highlight rects to native PDF user space

28Smiles:
`zotero_create_annotation` and `zotero_create_area_annotation` wrote highlight rectangles in PyMuPDF's CropBox-normalized coordinate space (where `page.rect` is always `(0, 0, w, h)`), but Zotero positions annotations in the PDF's **native user space**, whose origin is the MediaBox lower-left corner. For PDFs whose page box origin is not `(0, 0)` every created highlight was translated by exactly `(MediaBox.x0, MediaBox.y0)`: shifted into the margin horizontally and up/down the page vertically. The text search found the right text and the comment/tags/annotation key were correct; only the drawn rectangle (and the derived `annotationSortIndex`) was wrong.

PDFs with a default `(0, 0)` page box origin were unaffected, which made the bug easy to miss.

### DIFF
--- a/src/zotero_mcp/pdf_utils.py
+++ b/src/zotero_mcp/pdf_utils.py
@@ -219,19 +219,42 @@ def _get_spans_in_range(
 # Coordinate Conversion
 # =============================================================================
 
+def _page_to_pdf_transform(page) -> tuple[float, float, float, float, float, float]:
+    """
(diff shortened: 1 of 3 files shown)


54yyyu:
Merged, thanks! Great catch on the MediaBox origin offset, and the transform-based fix handles rotation too.

Issue #393, opened 2026-07-19
Semantic search's incremental sync watermark (`last_sync_version`) is shared across all libraries, corrupting sync state after `zotero_switch_library`


mronkko:
## Summary
`ZoteroSemanticSearch` stores exactly one `last_sync_version` integer in `config.json` (`semantic_search.last_sync_version`), used to drive since-based incremental updates via pyzotero's `item_versions(since=V)`. But every Zotero library — the personal library and each group — maintains its own independent, monotonically increasing version counter. Because the watermark isn't scoped per library, switching the active library with `zotero_switch_library` and then running `zotero_update_search_database` compares the new library's version counter against a stale watermark left over from a *different* library, and then overwrites that shared watermark with the new library's counter — corrupting sync state for both.

## Component
`src/zotero_mcp/semantic_search.py`:
- Read: `_load_last_sync_version()` (~line 528)
- Write: `_save_update_config()` (~line 546, writes `semantic_search.last_sync_version`)
- Incremental gate: `update_database()` (~lines 1599–1601, 1638, 1647, 1655, 1829)

## Steps to Reproduce
1. In local or web-API mode, run `zotero_update_search_database()` against your personal library. Suppose its version counter is at 50,000 after this sync; `last_sync_version` is saved as `50000`.
2. Call `zotero_switch_library(library_id="<some group>", library_type="group")` to switch the active library to a group whose own version counter is much lower (say `1,200` — an independent counter, unrelated to the personal library's).
3. Run `zotero_update_search_database()` again.

## Actual Behavior
- `last_sync_version` (50000) is loaded and is `> 0`, so `use_incremental = True`.
- `target_sync_version = self.zotero_client.last_modified_version()` returns the **group's** current version (1,200) — `last_modified_version()` hits `self.items(limit=1)`, scoped to whichever library the client is currently configured for.
- `1200 != 50000`, so the "unchanged since last sync" short-circuit doesn't fire.
- `_get_changed_items_from_api(since_version=50000)` calls `item_versions(since=50000)` **on the group library**. Since no item in that group has ever reached version 50000 (its counter tops out at 1,200), this returns an **empty dict** — every item in the group looks "unchanged," even on the group's first-ever sync.
- Result: 0 items are processed/added for the group (reported as a successful no-op update), **and** `_save_update_config(last_sync_version=1200)` overwrites the shared watermark, clobbering the personal library's `50000` with the group's `1200`.
[...]

Issue #163, opened 2026-03-20
Feature request: global search across all libraries (metadata and semantic)


mronkko:
## Summary

Add a `search_all_libraries` parameter (or equivalent mechanism) to both metadata search tools and `zotero_semantic_search`, so that a single query can find items regardless of which library they live in. Alternatively, this could be implemented by setting the current library to "global". 

## Motivation

Currently, all search tools — `zotero_search_items`, `zotero_advanced_search`, and `zotero_semantic_search` — operate on one library at a time. The user must know in advance which library contains the item, switch to it with `zotero_switch_library`, and then search. This makes it easy to miss items that exist in group libraries, and forces unnecessary round-trips when the target library is unknown.

**Concrete example:** An article stored in my group library ("AI in entrepreneurship") is completely invisible to semantic search, even though topically identical articles in the personal library are returned. The only indication that something is missing is prior knowledge of the item's existence.

If Zotero MCP is used to aid in writing process, it would be beneficial if it worked similarly to the insert citation dialog that searchers sources across all libraries.

## Relation to existing issues

This request consolidates and extends two open issues:

- **#18 – Semantic indexing of multiple libraries**: requests that the ChromaDB index cover group libraries in addition to the personal library.
- **#123 – Support searching local group libraries**: requests that keyword/metadata search in local mode include group libraries.

Both address the same root cause — searches are scoped to one library — but only for their respective search modalities. This issue requests a unified solution covering both.

### 1. Global metadata search

Add a `search_all_libraries: bool = False` parameter to `zotero_search_items` and `zotero_advanced_search`. When `True`:

- Iterate over all accessible libraries (user library + all group libraries).
- Merge and deduplicate results.
- Include a `library` field in each result indicating its source.

Alternatively, allow setting the current library to -1, NULL or something else indicating global search scope.

### 2. Global semantic search

Extend `zotero_update_search_database` and `zotero_semantic_search` to support a multi-library index:

- **Option A — unified index**: embed all libraries into a single ChromaDB collection, tagging each document with its `library_id`. Filter by tag when a specific library is active; omit the filter for global search.
- **Option B — per-library collections**: maintain one ChromaDB collection per library, and fan out queries across all collections when global search is requested.

Either way, `zotero_semantic_search` should accept an optional `library_id` filter and default to searching all indexed libraries.

### 3. Result format

Results from a global search should include a `library` field, e.g.:

```{Library: AI in entrepreneurship (groupID=6015547)}```

## Use cases

[...]

PR #396, opened 2026-07-19
Tag semantic search index with group_id for multi-library filtering (#163, phase 1)

mronkko:
## Summary
- ChromaDB documents in the semantic search index are now tagged with a `group_id` field (`0` = personal library, else the Zotero groupID) — local mode via a new `LocalZoteroReader.get_key_group_map()`, web mode via the active `zotero_switch_library` override.
- `zotero_semantic_search` gains a `library_id` parameter (`0`/`"user"` for personal, a group's numeric groupID, or omitted to search every indexed library — the new default) that filters DB-side via a ChromaDB `where` clause, never a Python post-filter.
- Existing collections migrate automatically on the next `zotero_update_search_database` run: a metadata-only backfill (no re-embedding, no doc-id changes), gated by a new `index_schema_version` config key.

This is phase 1 of #163 (global multi-library search). It's deliberately scoped to just tagging + migration + filtering:
- Cross-library result enrichment (fetching a search hit's full item via a client scoped to *its* library, rather than whichever library is currently active) is left for a follow-up — a hit from a group library other than the active one currently degrades to the existing bare-key fallback instead of full detail.
- The incremental-update deletion pass is not yet scoped per library, and the sync watermark is still a single process-wide scalar (#393) — both are pre-existing bugs, independent of this change and not worsened by it, tracked for a later bug-fix pass rather than bundled here.

## Test plan
- [x] `pytest tests/` — 1106 passed
- [x] `ruff check` clean on all touched files
- [x] New tests: `tests/test_semantic_multilibrary.py` (tagging across all three item-producing paths, feed/publications-library exclusion, migration backfill idempotency, DB-side `where`-filter merging) and `LocalZoteroReader.get_key_group_map()` coverage added to `tests/test_local_db.py`
- [x] Manual: built an index, ran `zotero_update_search_database` without `--force-rebuild` — backfill ran, doc count/IDs unchanged, every doc gained `group_id`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- **Multi-library tagging in the semantic search index (#163, phase 1)** — ChromaDB documents are now tagged with a `group_id` field (0 = personal library, else the Zotero groupID — the same `0`/groupID convention already used by `zotero_switch_library`). `zotero_semantic_search` gains a `library_id` parameter (accepting `0`, `"user"`, or a group's numeric groupID; omitted to search every indexed library — the new default) that filters on `group_id` DB-side via a ChromaDB `where` clause, never a Python post-filter. Existing collections are migrated automatically on the next `zotero_update_search_database` run (metadata-only backfill, no re-embedding, no downtime). Note: full item enrichment for a search hit still uses the currently active library's client, so a hit from a group library other than the active one may show limited detail until that capability lands in a follow-up; the incremental-update deletion pass is likewise not yet scoped per library (tracked separately, see #393 and related issues).
+
 ## [0.6.2] - 2026-07-13
 
 ### Added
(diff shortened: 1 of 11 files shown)


mronkko:
This PR starts implementing #163. The first step is to add group_id to the ChromaDB and filter by group in database calls instead of in python code. This is just a bit of plumbing for now, with a huge efficiency gain in semantic searches for users with multiple groups. I have about 20 groups and 25 000 items in these groups. Finding something from a small group is virtually impossible with the current implementation because the search results from the large groups need to be all filtered out from paged database results.

mronkko:
One more thing: Even if the code and the PR comment is mostly done with Claude Code, the fixes themselves are designed and reviewed by me.

mronkko:
The full set of changes is available in this branch [https://github.com/mronkko/zotero-mcp/tree/feat/sqlite-search-backend](https://github.com/mronkko/zotero-mcp/tree/feat/sqlite-search-backend). I will issue a new pull request once this one is merged. The SQLite search backend is one or two orders of magnitude faster than local or Web API, and implements #167 and #163. The branch also contains a number of bug fixes for the previous search implementation (e.g. #393 and erroneous parsing of years from date strings in advanced search)

PR #397, opened 2026-07-19
Optimization & Scaling for Large Libraries: Multicore PDF Text Extraction with Transient Cache, Multitread API Streaming with Rate Limiting & Gemini Batch API

### LINKED ISSUE #390: Optionally use parallel processing for PDF extraction
Indexing large libraries currently takes hours. Even if the embeddings can be generated using parallel processing and batch jobs on the OpenAI API, the text extraction is done sequentially. Text extraction should optionally use parallel processing with all available cores.

mronkko:
# Optimization & Scaling for Large Libraries: Multicore PDF Extraction, Streaming Rate Limiting & Gemini Batch API

Closes #390

## Overview

This PR refactors and scales the full-text extraction and semantic embedding pipeline to handle large, long-standing Zotero research collections efficiently. 

For research libraries accumulated over decades—such as a my 20-year-old collection with 10,000+ PDFs yielding a full-text index in the 50M token range—sequential processing and unpaced API calls result in index operations that take days. This PR addresses those performance barriers directly, implementing multicore PDF extraction, a transient disk cache, multi-threaded API access with header-aware streaming rate limiting, and Gemini Batch API support.

---

### 1. Multicore PDF Extraction & Transient Fulltext Cache (Resolves #390)
* **Multicore PDF Processing**: PDF text extraction now utilizes multiprocessing to process PDFs concurrently across CPU cores instead of running sequentially.
* **Transient Disk Cache (`fulltext_cache.py`)**: Extracted full-text content is cached on disk prior to embedding generation. If an indexing run is interrupted by network failures, API rate limits, or user cancellation, subsequent runs resume instantly from the cached text without re-extracting PDFs. Cached entries are automatically evicted as items are successfully indexed into ChromaDB.

### 2. Embedded Realtime Pipeline & Adaptive Rate Pacing (OpenAI / Remote APIs)
* **Concurrent Streaming Pipeline**: Sub-batches are prepared, embedded, and committed concurrently using a queue-based `ThreadPoolExecutor` architecture (`max_parallel_requests`).
* **Header-Aware Adaptive Rate Limiter**: Pacing dynamically checks OpenAI rate-limit headers (`x-ratelimit-remaining-tokens`) and response status:
  * When capacity headroom is available (>30% tokens remaining), worker threads run at full parallel speed.
  * When token capacity narrows or `429` responses occur, token dispatches pause dynamically without failing the overall run.
* **Expanded HTTP Connection Pooling**: Configured explicit `httpx` connection pool limits (`max_connections=100`) to eliminate TCP/TLS socket acquisition bottlenecks under high thread concurrency.

### 3. Gemini Batch API Integration
* **Asynchronous Batching for Gemini**: Added support for Gemini Batch API submission (`--gemini-batch` / `--batch-provider gemini`), allowing bulk embedding generation for millions of tokens at ~50% API cost.

### 4. Modular Embedding Provider Architecture
* **Unified Provider Hierarchy**: Refactored embedding functions under `RemoteEmbeddingFunction`, standardizing batching, error classification, retry logic, and telemetry logging across OpenAI, Gemini, Voyage, and Ollama providers.
* **Informative Telemetry**: Detailed real-time progress logging reporting worker thread IDs, HTTP round-trip latency, OpenAI server processing durations (`openai-processing-ms`), and rate limiter wait states.

---

## Updated CLI Options for `zotero-mcp update-db`

| Flag / Parameter | Description |
| :--- | :--- |
| **`--extraction-workers N`** | Sets number of parallel CPU worker processes for PDF text extraction during `--fulltext` (default: config `semantic_search.extraction.workers` or 1; capped at CPU count / 8). |
| **`--clear-fulltext-cache`** | Clears the transient extracted-fulltext cache from disk before running. |
| **`-v`, `--verbose`** | Enables detailed real-time telemetry output, including worker thread IDs, HTTP round-trip latency, OpenAI server execution times, and rate-limiter wait states. |
| **`--batch` / `--no-batch`** | Generic CLI flag to enable or disable asynchronous Batch API mode for the configured embedding model (replaces deprecated `--openai-batch`/`--gemini-batch`). |
| **`--batch-provider {openai,gemini}`** | Explicitly selects the Batch API provider to use with `--batch` (default: auto-detected from configured embedding model). |
| **`--batch-max-tokens N`** | Overrides max estimated enqueued tokens per batch job (supports tier presets for OpenAI & Gemini). |
| **`--batch-max-requests N`** | Overrides max requests per uploaded JSONL batch file (default: 50,000). |
| **`--auto-loop`** | Automatically polls batch status, imports completed batches, and submits pending chunks until the full run finishes. |
| **`--batch-poll-interval N`** | Sets interval in seconds between status polls when `--auto-loop` is enabled (default: 60s). |

The PR has been tested and works well with my library.


mronkko:
This needs a bit more integration testing still, but I wanted to create a PR showing that this is coming. I will mark the integration test completed once I have indexed my full library.

54yyyu:
Thanks for putting this together, and for #396 alongside it. #396 is merged (with a small conflict fix against #387's `children()` pagination changes, which touched the same imports in `semantic_search.py`).

This one I can't take as it stands, and there's a concrete blocker: the suite doesn't collect at all.

CI never ran here (first-time-contributor runs sit in `action_required`, which is on me), so nothing caught it. Running it locally on pytest 8.3.3:

```
$ pytest -q --ignore=tests/test_lifespan.py
...
E   ModuleNotFoundError: No module named 'conftest'
!!!!!!!!!!!!!!!!!!! Interrupted: 35 errors during collection !!!!!!!!!!!!!!!!!!!
```

The cause is `consider_namespace_packages = true` in `pyproject.toml`. Your comment explains the problem you were solving (two `conftest.py` files colliding under the same module name), but the cure removes the rootdir entry that makes the repo-wide `from conftest import FakeZotero` convention work, and 35 test modules use it. It may well pass on whichever pytest version you're running; it does not on 8.3.3, and CI installs a plain `pytest`.

Please drop that setting and solve the live-conftest collision another way. Naming the live fixtures module something other than `conftest.py` and importing it explicitly is the least invasive option, since it leaves the existing convention untouched.

Second thing: at +15,377/-1,413 across 58 files this is more than I can review as one unit, and it lands four independent changes that carry quite different risk:

1. multicore PDF extraction
2. the transient fulltext disk cache
3. streaming rate limiting with header-aware pacing
4. Gemini Batch API support

Could you split it? (1) and (2) together resolve #390 and are the ones I most want. They're also self-contained enough to review properly and ship on their own. Rate limiting and the Gemini batch path can follow separately.

Worth knowing before you rebase: quite a lot has landed on main since you opened this (#387 paginates `children()` everywhere, #367 unified config loading into a new `config.py` and changed `setup_semantic_search()` to return a `(config, db_path)` tuple, plus #396). `semantic_search.py` and `chroma_client.py` both moved.

Also flagging that `.github/workflows/release.yml` installs a hand-maintained dependency list with `--no-deps`, so `pytest-httpserver` and anything else new needs adding there or the tests that need it will silently not run in CI.

Happy to review the extraction and cache split as soon as it's up!

mronkko:
I agree that this became too big. I will split it in four sequential pull requests. I am traveling, so the first pull request will coming about two week from now when I am back at the office.

54yyyu:
Closing this umbrella now that the split plan is underway and its first half has landed: #440 (multicore extraction + transient fulltext cache) is merged, which covers items (1) and (2) — the parts flagged earlier as most wanted — and resolves #390.

The remaining two splits stay very welcome as fresh PRs against current main:
- streaming rate limiting with header-aware pacing
- Gemini Batch API support

Two things worth knowing before you cut them: main has moved substantially (#440, #442-context, #444's embeddings-provider package — the embedding functions now live in `zotero_mcp/embeddings/providers/` behind a registry, which is probably where the Gemini batch path wants to hook in), and the rate-limit picture changed under pyzotero 1.13.5 (it now retries 429s internally and raises `TooManyRetriesError` — see the review on #442), so the header-aware pacing design may want to sit on top of that rather than around it.

Thanks again for the work in here — it's been landing piece by piece, which is exactly what was asked.

---

PR #401, opened 2026-07-26
Restrict default Chroma storage permissions

michaelzehetleitner:
## Summary

- create the default `~/.config/zotero-mcp` directory with owner-only permissions on POSIX systems
- tighten existing default config directories to `0700` when semantic search initializes
- keep the hardening best-effort and leave custom persistence directories unchanged
- add a regression test for the directory permissions

## Why

The default ChromaDB directory can contain indexed Zotero metadata and full text. The previous `mkdir` call inherited the process umask, which commonly creates a traversable `0755` directory and exposes that index to other local accounts.

## User impact

On POSIX systems, the next semantic-search initialization makes the default Zotero MCP config directory accessible only to its owner. Windows behavior and explicitly configured persistence directories are unchanged.

## Validation

- `uv run ruff check src/zotero_mcp/chroma_client.py tests/test_security_hardening.py`
- `uv run pytest tests/test_security_hardening.py -q` — 4 passed
- `uv run pytest -q --deselect tests/test_local_db.py::TestResolveAttachmentPath::test_attachments_without_base_path_returns_none` — 1,086 passed, 1 deselected

The deselected test reads the machine's real Zotero attachment preference and expected no configured base path; this machine has one configured.

### DIFF
--- a/src/zotero_mcp/chroma_client.py
+++ b/src/zotero_mcp/chroma_client.py
@@ -27,6 +27,17 @@ from zotero_mcp.utils import suppress_stdout
 logger = logging.getLogger(__name__)
 
 
+def _restrict_directory_permissions(path: Path) -> None:
+    """Best-effort owner-only permissions for sensitive local storage."""
+    if os.name != "posix":
+        return
+
+    try:
+        os.chmod(path, 0o700)
+    except OSError as exc:
+        logger.warning("Could not restrict permissions on %s: %s", path, exc)
+
+
 @register_embedding_function
 class OpenAIEmbeddingFunction(EmbeddingFunction):
     """Custom OpenAI embedding function for ChromaDB.
@@ -465,7 +476,8 @@ class ChromaClient:
         if persist_directory is None:
             # Use user's config directory by default
             config_dir = Path.home() / ".config" / "zotero-mcp"
-            config_dir.mkdir(parents=True, exist_ok=True)
+            config_dir.mkdir(parents=True, exist_ok=True, mode=0o700)
+            _restrict_directory_permissions(config_dir)
             persist_directory = str(config_dir / "chroma_db")
 
         self.persist_directory = persist_directory
--- a/tests/test_security_hardening.py
+++ b/tests/test_security_hardening.py
@@ -38,6 +38,20 @@ def test_restrict_file_permissions_swallows_errors():
     _restrict_file_permissions("/nonexistent/path/to/config.json")
 
 
+def test_restrict_chroma_directory_permissions_sets_owner_only(tmp_path):
+    """Default semantic-search storage is inaccessible to other local users."""
+    if os.name != "posix":
+        pytest.skip("POSIX directory permissions only")
+    from zotero_mcp.chroma_client import _restrict_directory_permissions
+
+    config_dir = tmp_path / "zotero-mcp"
+    config_dir.mkdir(mode=0o755)
+
+    _restrict_directory_permissions(config_dir)
+
+    assert stat.S_IMODE(config_dir.stat().st_mode) == 0o700
+
+
 def test_pdfannots_timeout_returns_empty(monkeypatch):
     """A pdfannots2json timeout is handled gracefully (returns [])."""
     import zotero_mcp.pdfannots_helper as ph


michaelzehetleitner:
This PR was coded by Claude Code and I verified.

54yyyu:
Thanks for this! The reasoning is right: the default Chroma directory holds indexed metadata and full text, and inheriting the umask to 0755 exposes it to other local accounts.

CI is red on all three Python versions, but not for a reason that reflects your change:

```
ERROR tests/test_security_hardening.py::test_restrict_chroma_directory_permissions_sets_owner_only
E   FileNotFoundError: [Errno 2] No such file or directory: '/tmp/pytest-of-runner/pytest-0'
```

It fails at fixture setup, not in the assertion, and it passes locally for me (1084 passed on your branch merged onto main). My read is that tightening a directory inside pytest's tmp root interferes with `tmp_path_factory`'s numbered-directory rotation, which only shows up on a runner where that root is freshly created. Creating the directory under your own `tmp_path` subdirectory rather than anywhere pytest manages should settle it.

One thing worth deciding while you're here: you tighten existing default config directories to 0700 on every semantic search init. That's a silent permission change to a directory the user may have deliberately set otherwise. I'd rather it only applied at creation time, or logged when it changes an existing directory's mode, so nobody has to work out later why their setup changed.

Main has moved a fair bit today (seven PRs merged, `chroma_client.py` among the files touched), so a rebase is worth doing before the next push.

Mark it ready when CI is green and I'll take another look!

54yyyu:
Correcting my earlier comment: the CI failure was not your PR's fault, and my guess about `tmp_path_factory` rotation was wrong.

`_cleanup_path` in `tools/read_pdf.py` deletes the parent directory of a PDF, guarded only by `parent.startswith(tempfile.gettempdir())`. On Linux `gettempdir()` is `/tmp`, so a path like `/tmp/test.pdf` has `/tmp` as its parent, passes the check, and runs `shutil.rmtree("/tmp")`. A test stub in `test_read_pdf.py` returned exactly that path, so every CI run on Linux wiped the runner's temp directory partway through the suite. Whichever tests happened to want a `tmp_path` after that point failed with `FileNotFoundError: /tmp/pytest-of-runner/pytest-0`, which is the error you and I were both looking at.

It never showed up locally on macOS because `gettempdir()` there is under `/var/folders`, so the prefix never matches `/tmp` and the whole branch was dead.

Fixed on main: cleanup now requires the directory to be strictly below the temp root and to carry the `zotero_pdf_` prefix that `mkdtemp` gives the directories we create, plus the caller no longer runs cleanup at all on files resolved from the user's Zotero storage. CI is green again on all three Python versions.

So: rebase on main and your test should pass as written. Sorry for pointing you at your own fixture first.

The substantive question from my earlier comment still stands, if you want to address it in the same push: tightening *existing* directories to 0700 on every semantic search init is a silent permission change to a directory a user may have set deliberately. I'd rather that applied only at creation, or logged when it changes an existing mode.

54yyyu:
Heads up: the red CI on this branch is not your change.

The failure is `FileNotFoundError: /tmp/pytest-of-runner/pytest-0` — the runner's pytest temp root vanishing mid-suite. That was `zotero_read_pdf_pages`' cleanup step deleting the *parent directory* of the PDF it had worked on, guarded only by a `startswith(tempfile.gettempdir())` check that on Linux passes for anything directly under `/tmp`. It wiped the runner's temp dir out from under whatever test ran next, which is why it landed on your `tmp_path`-using test. It was fixed in 0.6.3, released two days after you opened this.

I merged your branch onto current `main` locally and ran the suite: **1358 passed, 41 skipped** with `CI=true`, and `tests/test_security_hardening.py` passes 4/4. So a rebase should be all this needs.

On the change itself — tightening the default `~/.config/zotero-mcp` to `0700` is worth having, and scoping it to the default directory while leaving explicitly configured persistence paths alone is the right call. Happy to review properly once you take it out of draft.

54yyyu:
Nudge on this one, since it's been sitting since the 2nd.

Main has moved a good deal since you last pushed, so a rebase is worth doing before anything else. The CI failure you saw was the `read_pdf` temp-cleanup bug, not your change, and that's been fixed since 0.6.3, so your test should pass as written once you're on current main.

The one substantive question from my earlier review still stands: tightening *existing* directories to 0700 on every semantic search init is a silent permission change to a directory a user may have set deliberately. Either apply it at creation only, or log when it changes an existing mode.

Happy to review properly whenever you take it out of draft. No rush if you're busy, just didn't want it going stale.

54yyyu:
@michaelzehetleitner thank you for this, and sorry it sat so long. I've taken it over so it can ship in 0.12.0: 69425cf on `maint/0.12.0` (#544), with you as co-author.

Two changes from your version, both from the review question that was still open:

- **Creation only, not every run.** A directory the code creates is `0700`. An existing directory is left alone, since its mode may be deliberate, and if other users can read it you get one warning with the `chmod 700` to run.
- **Every path that can create the directory, not just the Chroma client.** `~/.config/zotero-mcp` is also created by the setup wizard, `update-db`, local write authorization and the update lock, and whichever runs first decides the mode. They all go through one `utils.ensure_private_dir` now.

The regression tests are in `tests/test_security_hardening.py`. Closing this PR in favour of that commit.

PR #409, opened 2026-07-29
fix: resolve update-item fields through the item type's base-field schema (#402)

### LINKED ISSUE #402: zotero_update_item silently drops base-field-renamed fields (e.g. statute title) — worth making field routing schema-aware?
## What happens

`zotero_update_item` can't set the title of item types whose title field isn't literally `title`. I hit this cataloguing a regulation as a `statute`. Passing `title=` leaves the item unnamed — the value is silently dropped with `Skipped (not valid for item type 'statute'): title`. The same hits every base-field rename: `title` → `nameOfAct` (statute) / `caseName` (case) / `subject` (email), and beyond title, `date` → `dateEnacted`, `publisher` → `institution`, and so on. The only workaround is to mis-type the item as `document` (which has a literal `title`) — trading correct typing for a settable name.

## Root cause

The field-write loop maps each param to a fixed API key (e.g. [`field_updates["title"] = title`](https://github.com/54yyyu/zotero-mcp/blob/566b396/src/zotero_mcp/tools/write.py#L1834)) and gates each write on [`if field in data`](https://github.com/54yyyu/zotero-mcp/blob/566b396/src/zotero_mcp/tools/write.py#L1877) — "is this key already on the fetched item?" as a proxy for "is it valid for the type?". That proxy fails whenever a type renames a base field, because the param name (`title`) never appears in `data` (which has `nameOfAct`).

## Recurring issue

This is the same gate flagged in #320: *"the loop only sets `citationKey` when it's already present … passing it to an item without one silently no-ops."* We fixed that with an `elif field == "citationKey":` special-case (#321). `nameOfAct` is the second instance of the same root cause — and another special-case (or a small whitelist map) is just a second patch on the same spot; it won't prevent the next renamed field or a future item type from dropping writes.

## Towards a durable fix

Make routing base-field-aware: resolve each generic param to the type's actual field via Zotero's base-field schema, and validate against the type's field set (from the template) rather than instance presence. Two possible shapes:

- **Runtime** — fetch + cache Zotero's schema (it carries `baseField`), build per-type maps. Adapts to new types automatically.
- **Vendored** — a base-field table generated from the schema, no runtime fetch; regenerate when Zotero changes.

(pyzotero 1.13 exposes templates but not `baseField`, so either shape needs the schema.)

This is bigger than the per-field additions so far and it touches the core write path, so I wanted to check before building it. **@54yyyu, are you open to a change of this size, is there a shape you'd prefer, or a reason to keep the presence-based gate?** I'm happy to draft a PR against whichever direction you'd take.

ahharvey:
Closes #402.

Takes the schema-aware route: `zotero_update_item` resolves each generic parameter to the item type's actual field and validates against that type's declared field set, instead of gating on whether the key already exists on the fetched item.

What changes, and why:

- **Resolution + validation via the schema** — `title`→`nameOfAct` on a statute, `caseName` on a case, `date`→`dateEnacted`; a valid-but-absent field is now written, not skipped. The old `if field in data` gate used instance-presence as a proxy for type-validity, which breaks on every base-field rename.
- **`citationKey` special-case removed** — it's a real schema field on every regular item type, so it validates like any other. Retires the #321 branch instead of adding a second one beside it.
- **Base-field map is vendored, not just fetched** (`data/zotero_basefields.json`, a trimmed per-type slice of `/schema`, regenerated by `scripts/gen_basefield_map.py`). #402 proposed falling back to today's behaviour when `/schema` is unreachable — but that falls back to the bug, and strands the local-only/offline users who can't reach `api.zotero.org`. The vendored floor is always correct for known types; the legacy presence-gate now only applies to a type absent from the map entirely. That's more than #402 outlined — happy to drop to pure-runtime-with-today's-fallback if you'd prefer.
- **Resolver lives in a new `schema.py`, not `write.py`** — the `schema-refresh` CLI and the read-side callers (the `data.get("title")` sites that render a statute as "Untitled") would otherwise have to import the write tool. `write.py` consumes `resolve_field`/`valid_fields`; reads can reuse `base_field_of` later.
- **Read-side migration deferred** — the 50-plus display sites that assume `title` are a separate behaviour from the write path and out of scope for #402; the inverse is exposed now so that follow-up is mechanical.
- **Runtime refresh** at startup (backgrounded, non-fatal) and via `zotero-mcp schema-refresh` — a 7-day TTL-gated conditional GET, so an unchanged schema is a bodyless 304 and new item types are picked up without touching the per-request path. Zotero has no user- or library-defined item types, so the global schema is authoritative for every library.
- **Left out on purpose** — a weekly CI regen job + a release-time freshness gate. Both touch the release workflow, so they're a separate PR if you want them.

Tests cover schema resolution/validation and the refresh paths (TTL gate, 304, 200-with-new-type, non-JSON/500, offline fallback), routing through `update_item` for statute/case/bookSection, and the unknown-type presence-gate fallback. Two existing tests that asserted `place`-on-journalArticle and `accessDate`-on-book were skipped are flipped — both are valid per the schema and pyzotero's own `item_template`, so they encoded the presence-gate bug.

Backwards compatible. Happy to split commits, scope down, or move the module boundary.


ahharvey:
Two things I deliberately kept out of this PR. Both are self-contained follow-ups — happy to open either if you want them.

**1. Read-side base-field resolution.** `schema.py` already exposes `base_field_of` (the inverse: `nameOfAct`→`title`), unused for now. A follow-up would switch the display reads that assume a literal `title` over to it, so a statute/case/email renders its real name instead of "Untitled". The sites are `client.py`, `utils.py`, `tools/connectors.py`, `tools/annotations.py`, `semantic_search.py`, `cli.py`, and `better_bibtex_client.py`. Mechanical, but a different behaviour from the write path, so better reviewed on its own.

**2. Keeping the vendored map current in CI.** A scheduled GitHub Action (weekly) that runs `scripts/gen_basefield_map.py` and opens a PR when Zotero bumps the schema version, plus a release-time check that fails the build if the vendored map is behind. Keeps the offline floor fresh without anyone remembering to regenerate it. It touches the release workflow, so I left it out here.

PR #417, opened 2026-08-01
Add opt-in SQLite metadata search backend (#167)

mronkko:
## Summary

Adds an opt-in **direct-SQL metadata search backend** for `zotero_search_items` and `zotero_advanced_search`, closing #167. Set `ZOTERO_SEARCH_BACKEND=sqlite` (local mode only) to route both tools through direct SQL against `zotero.sqlite` instead of pyzotero:

- `zotero_search_items`: one SQL query with an `OR`'d `LIKE` across title/creator/year (plus abstract/tags/notes in `everything` mode) replaces the pyzotero variant-search loop.
- `zotero_advanced_search`: real SQL `WHERE`/`EXISTS` clauses replace the existing client-side path, which today has **no server-side condition support at all** — it pages the *entire* library 100 items at a time via `zot.items(start=start, limit=100)` (no `q`, no `itemType`, no field filter of any kind) and only evaluates conditions in Python, per item, after every single one has been fetched over the API.

Falls back automatically and transparently to the existing pyzotero path — the tested, shipped behavior — whenever the SQL backend is unavailable (not local mode, unreadable DB) or hits a condition/operator/`itemType` expression this v1 doesn't cover (an unsupported field, a boolean `itemType`/`tag` expression, an unrecognized `group_id`). Default behavior (`ZOTERO_SEARCH_BACKEND` unset) is completely unchanged.

Condition support (all ten existing operators): `title`, `creator`, `year`, `date`, `dateAdded`, `dateModified`, `itemType`, `tag`, `publicationTitle`, `abstractNote`, `DOI`, `collection` (new — recursive subcollection resolution, reusing the pattern `get_items_with_text` already uses for `collection_keys`).

Built on #396 (Phase A, **merged**) — this PR's `local_db.py` changes build directly on that PR's `get_key_group_map()` / library-scoping additions, and this branch is rebased onto current `main` to sit cleanly on top of it.

## How the opt-in actually works

Nothing changes unless `ZOTERO_SEARCH_BACKEND=sqlite` is set explicitly. `get_search_backend()` (`utils.py`) reads that one env var and returns `"sqlite"` only for that exact literal value; unset — or anything else — returns `"api"`, the existing pyzotero-based behavior every deployment already runs. Two more gates layer on top, both there to make the opt-in *safe* rather than to restrict it further:

1. **Local mode is required.** `get_local_zotero_reader()` returns `None` immediately unless `ZOTERO_LOCAL` is set — the SQL backend reads `zotero.sqlite` directly off disk, so it's meaningless in web-API mode.
2. **The DB must actually be readable.** If `LocalZoteroReader()` can't find/open `zotero.sqlite` (`FileNotFoundError`), that also returns `None`.

Every one of those checks fails *soft*. At each call site (`_search_items_via_backend` for `zotero_search_items`, and the equivalent block inside `advanced_search`), a `None`/exception from the SQL path falls straight through to the existing pyzotero code, unchanged — same for any individual query the v1 SQL translator doesn't understand (a boolean `tag`/`itemType` expression, an unrecognized field, an unrecognized `group_id`).

Net effect: with the env var unset, the code path is byte-identical to today — the `get_search_backend() == "sqlite"` check short-circuits false and the SQL block never runs at all. With it set, you get the SQL path for whatever it covers, and silently the existing path for whatever it doesn't. There's no scenario where opting in can produce a worse or broken result versus not opting in — only faster/more-correct, or identical.

## Tested against real data, not just synthetic fixtures

Beyond the unit tests (33 new cases against a synthetic sqlite fixture, plus routing/fallback tests), I ran this against my own real, ~55,000-item Zotero library across all three ways a search can actually execute — the local Zotero desktop API, the Zotero web API, and the new SQLite backend — using the real shipped tools end-to-end (not mocks).

| Scenario | Local Zotero API | Web Zotero API | SQLite backend (this PR) |
|---|---|---|---|
| `search_items`, author "X" (multiple real items) | 17.2s → 155 found | 6.9s → 100 found | **0.82s → 100 found** |
| `search_items`, author "Y" (exactly one real item) | 6.6s → 1 found | 1.5s → 1 found | **0.83s → 1 found** |
| `advanced_search`, creator contains "X" AND date between two real years | 27.4s → 86 found ⚠️ | 467.6s (~7.8 min) → 86 found ⚠️ | **0.20s → 111 found ✓** |

For the two `search_items` scenarios, all three backends found the same items (small count differences come from an existing, harmless quirk: the pyzotero path doesn't cap its *total* merged result to `limit`, only per-query-variant — documented in code).

For `advanced_search`, the SQL backend's 111 was verified against an independently hand-written SQL query built directly in the test (not reusing any of this PR's code) — exact match. The old path's 86 is **not** used as a correctness reference: chasing down why it disagreed surfaced a real, pre-existing bug, unrelated to this PR's own logic, and fixed as part of it (see below) once understood.

**Root cause, and why this PR's numbers are the correct ones:** Zotero stores the `date` field internally as a multipart string in `itemDataValues.value` — `"<ISO YYYY-MM-DD, '00' for unknown parts> <original user-typed text>"`, e.g. `"2016-10-01 October 1, 2016"` — but the pyzotero/web-API JSON only ever exposes the display half (`"October 1, 2016"`). The *existing* `advanced_search`'s `year` condition does `data.get("date")[:4]`, i.e. slices the *display* string — which is only actually a year when the user's original text happened to start with one. For most real dates ("October 1, 2016", "07/2021", ...) it silently takes garbage and the comparison becomes essentially arbitrary. This PR's SQL backend reads the *raw* multipart value directly, matching Zotero's own canonical year extraction exactly (confirmed against Zotero's own source: `item.js`'s `getField('date', true, true).substr(0, 4)` and `searchConditions.js`'s `SUBSTR(value, 1, 4)`).

While fixing that, I found (and fixed, in this PR) a related bug in my own new code: the `date` field's *own* condition (not `year`) needs to split by operator the same way Zotero's `search.js` does — `SUBSTR(value, 1, 10)` (the ISO portion) for range operators (`isBefore`/`isAfter`/`isGreaterThan`/`isLessThan`), the display text for everything else (`is`/`contains`/`beginsWith`/etc.). Both are covered by new regression tests against a fixture using real multipart-formatted dates.

The existing `advanced_search`'s buggy `year` extraction is a pre-existing issue on `main` today, unrelated to this PR's own correctness, and is being tracked separately for a fix to the old path.

## Also tested

- Real local-API/SQL-backend parity for `advanced_search`, bounded to one real collection (exact key-set equality).
- Real inert-in-web-mode behavior: `ZOTERO_SEARCH_BACKEND=sqlite` forced with no local mode falls through to the real web API correctly (not a mock).
- Real fallback on an unsupported condition/tag filter, using the real shipped tools.


mronkko:
This is a second of a set of four PRs. This PR fixes a bug in date processing which I discovered during testing and introduces SQLite search backed, which is substantially faster than API backends. All three backends are tested to produce identical search results (after the date bug fix). After this is merged, the next PR will introduce global search features that allow extending all current search functionality to cover all libraries the user has access to. The fourt PR will contain fix to a number of bugs I found in the code while working on the search backend.

54yyyu:
Reviewed and verified locally. Not merging yet — one behavioural finding below, and the underlying call is the maintainer's, for a reason I'll state at the end.

## What I verified

- **Merges clean onto current `main`** (which has since taken #409, #411, #414, #415). Full suite: **1436 passed**, no conflicts in `local_db.py` or `tools/search.py`.
- **SQL is properly parameterized.** Every value is bound with `?`; the only interpolated fragment is the column expression, and that comes from a `_SIMPLE_FIELD_SQL` / `_YEAR_FIELD_SQL` lookup gated on `resolved in _SIMPLE_FIELD_SQL`, so it can't carry user input. No injection surface I can find.
- **The opt-in gate is strict.** `get_search_backend()` returns `sqlite` only for that exact literal; unset, empty, and anything else return `api`. Confirmed the default path is unreachable without the env var.
- **Your multipart-date finding is correct**, and it's a good catch. `data["date"][:4]` on the display half is indeed arbitrary for anything not typed year-first, and `SUBSTR(value, 1, 4)` on the raw value is what Zotero itself does.

## The finding: the two backends disagree on `collection`

Your SQL path resolves collections **recursively** — `_collection_scope_ids` walks `parentCollectionID` and includes every descendant.

I fixed the API path's `collection` condition yesterday in 67932ef (#418: membership was read from a `data["collection"]` that doesn't exist, so the condition matched nothing on any backend). That fix is **direct membership only**, matching Zotero's own "Collection is X" with the subcollections box unchecked.

So after this merges, `advanced_search(conditions=[{"field":"collection","operation":"is","value":"X"}])` returns a different set depending on whether `ZOTERO_SEARCH_BACKEND=sqlite` is exported — same tool, same arguments, silently different answers. That's the failure mode the #418 reporter specifically objected to, and the automatic-fallback design makes it worse rather than better: a query that falls back mid-session changes semantics without saying so.

Not asking you to drop the recursive behaviour — it's arguably the more useful one. But the two paths need to agree. Either:
- make the SQL path direct-membership to match, and propose recursion separately as an explicit `includeSubcollections` option on both paths; or
- make the API path recursive too, and land both together.

I'd take the first — smaller, and it keeps the opt-in genuinely behaviour-neutral, which is the property the rest of this PR works hard to preserve.

## Why I'm not merging it myself

Everything above is addressable. The part that isn't mine to decide is that this adds a **second, permanently maintained implementation** of the two most-used tools in the server. Every future change to search semantics has to land twice and stay in sync — and the divergence above is that cost showing up before the PR has even merged. That's an architecture call for @54yyyu, not a review outcome, so I'm flagging it rather than deciding it.

The measurements are genuinely compelling (27s → 0.20s on `advanced_search`, and the existing path really does page the entire library to filter in Python), so I don't think the answer is no. Flagging for a maintainer decision on the maintenance commitment, with the collection-semantics fix as the concrete blocker either way.

mronkko:
I amended the PR with a couple of changes:
1) Fixed a bug where author title year search matches notes
2) Implemented live tests, that run test against live Zotero APIs and database to ensure the backends produce identical results
3) The collections functionality has been harmonized: All backends default to non-recursive collections in search. Recursive collections are still implemented in the sqlite backend, but not exposed to users.

Once this is merged, I will post the next PR. It will introduce global search (https://github.com/54yyyu/zotero-mcp/issues/163) and recursive collection search as user-facing features. Thereafter comes a fourth PR with various small bug fixes.

mronkko:
Moved the tip commit ("Let `zotero_add_item` add multiple DOIs/URLs/ISBNs in one call") off this branch — it doesn't belong in a SQLite-search-backend PR. It now lives on [`feat/batch-doi-add`](https://github.com/mronkko/zotero-mcp/compare/main...mronkko:zotero-mcp:feat/batch-doi-add) in my fork and will land separately. This branch is otherwise unchanged; apologies for the force-push noise.

54yyyu:
Status update after re-review on current main (post #436/#440/#444/#445):

- **The collection-semantics blocker from the earlier review is resolved.** All backends now default to direct membership for the `collection` condition; recursion stays implemented in the SQLite backend but unexposed, matching the plan of proposing it as an explicit feature on both paths later. Verified in the diff, not just the description.
- **Merges clean onto current main** — real test merge auto-resolves, and the merged tree runs **1674 passed** with only the known environmental failures. SQL remains fully parameterized (interpolated fragments are whitelist-derived only), the reader stays `immutable=1` read-only with per-query connections, and the opt-in gate is strict: without `ZOTERO_SEARCH_BACKEND=sqlite` the pyzotero path is byte-for-byte the old behavior, except for one deliberate parity fix worth a changelog line — `titleCreatorYear` results now exclude standalone notes on *every* backend (notes only ever matched via content, contradicting the mode's semantics).
- Minor opt-in-only divergences noted for the record, none blocking: `%`/`_` in a `contains` value act as LIKE wildcards under sqlite but literals on the API path, and SQL `LIKE` is ASCII-case-insensitive so accented-name matching differs slightly.

What remains is the question already flagged as the maintainer's call, and this comment deliberately doesn't decide it: merging adds a second, permanently maintained implementation of the two most-used tools, and every future search-semantics change lands twice. The live parity tests the author added (`ZOTERO_MCP_LIVE_TESTS=1`) are a real mitigation for exactly that risk. Technically this is ready; @54yyyu, the maintenance-commitment decision is yours.

---

mronkko:
I will change the architecture a bit.

mronkko:
Pushed `dc3b994`, which addresses the maintenance question rather than arguing around it.

## The short version

You were right that two implementations drift, and I under-rated how much. The two "minor opt-in-only divergences" you noted for the record were not minor, and there was a third nobody had found. So instead of asking you to accept the sync burden, the condition semantics now have **one definition** that both backends consume, and a parity suite in CI asserts they agree.

## The divergence that mattered

`tools/search.py`'s `_compare` folded *both* sides of a comparison through `_normalize_for_search().lower()`. The SQL path folded neither and relied on SQLite's ASCII-only `LIKE`.

On my real library (57,641 items):

```
creator contains "muller"   ->   4 items via SQL,  15 via the API
```

Every accented spelling was missing. In a library with Ambühl, Aygün, Andersén, Ångström, Bigné-Alcañiz, that is not an edge case — and it fails *silently*, returning a short answer rather than an error. Opting in was not "faster or identical" as the description claimed; it was quietly lossy.

`%` was worse than described too. It was not merely "acting as a wildcard": `title contains "%"` matched essentially the entire library. It now returns the one item with a literal `%`.

## What changed

`search_semantics.py` holds the operator table, the field aliases, the normalization function, and the comparator. `tools/search.py` delegates to it; `local_db.py` builds SQL from it. Crucially the *same* `normalize()` is registered on the SQLite connection as `zsearch_norm`, so SQL compares the identical folded string rather than an approximation:

```sql
zsearch_norm(<expr>) LIKE ? ESCAPE '\'
```

Two decisions worth flagging, both argued in comments:

- **Range operators are not wrapped.** `compare()` normalizes before its numeric parse, but the values reaching an ordering comparison are ASCII digits and ISO date prefixes where normalization is the identity. Skipping it is behaviour-preserving and avoids a Python callback per row. Pinned by test.
- **`search_items_sql`'s free-text path keeps `_generate_search_variants` and only gains escaping.** Its parity target is different — it does fuzzy recall, including umlaut *expansion* (`Müller` → `Mueller`) that `normalize()` deliberately does not do. Folding there as well would over-match relative to the pyzotero path.

## The third divergence, found by the parity suite

`_scalar_condition` guarded negated operators with `IS NOT NULL`, so an item with no `publicationTitle` failed `publicationTitle doesNotContain "Nature"` — while the API path matched it. The "a missing value satisfies nothing" rule belongs only to genuinely multi-valued fields, where `_extract_values` really does return `[]`; for scalars and collections the API path deliberately yields `[""]`, and documents that choice explicitly for `collection`. The SQL side had generalised the rule too far. Fixed, and the API path is untouched.

## Why this answers the maintenance question

`tests/test_search_parity_offline.py` runs **177 cases** through the real `advanced_search` tool against each backend and asserts the returned key sets are equal. There are no hand-written expected results — the assertion is that the backends *agree* — so adding a case extends coverage of both paths at once and cannot go stale.

It runs in CI. That is the part that matters: the live parity suite you credited as a mitigation cannot run in CI, and CI is exactly where a future divergence would be introduced. `tests/_search_corpus.py` declares the corpus once and materialises it both as a synthetic `zotero.sqlite` and as pyzotero dicts, so the comparison can't degenerate into comparing two different libraries. `test_sql_search_backend.py` now shares that schema too.

The deliberate `year`/`date` difference on multipart dates is **asserted rather than skipped**, so if either path changes, a test says so.

It is also worth saying the duplication is asymmetric: the pyzotero path is mandatory for web mode and cannot be removed, the SQL path is local-only, and in local mode pyzotero is itself reading the same `zotero.sqlite`. This PR does not add a second way to do something that previously had one — it makes the local case stop going through an HTTP server to reach a file it can open.

## Numbers, honestly

I re-measured on the current branch rather than reusing the figure in the description. Creator + date range, warm cache, best of 5, 57,641 items: **75 ms**, against 51 ms before normalization. So the fix costs ~25 ms and the earlier `0.20s` in the description was an older measurement, not a regression. Changelog updated to the current number.

Full suite: **1971 passed, 13 skipped**. The 78 existing search tests pass **unmodified** — the pyzotero path is a pure extraction with no behaviour change, which is why nothing here needs a `Fixed` or `Changed` entry.

## Deliberately not done

- **The silent per-query fallback is unchanged.** Making it loud — an explicit backend-selection gate with an enumerated coverage boundary — is a real improvement and I'd rather propose it separately than grow this PR.
- **Collection scoping stays direct-membership.** Recursion remains implemented and unexposed, as agreed.

Both are yours to call, as is the merge itself. Happy to split any of the above out if you'd prefer smaller pieces.

PR #429, opened 2026-08-04
fix(local_db): exclude trashed attachments and prefer PDFs in _iter_parent_attachments

### LINKED ISSUE #427: get_item_fulltext can return a trashed HTML snapshot's ft-cache instead of the live PDF (local mode)
## Summary

`LocalZoteroReader._iter_parent_attachments` ([src/zotero_mcp/local_db.py](https://github.com/54yyyu/zotero-mcp/blob/main/src/zotero_mcp/local_db.py)) neither excludes trashed attachments nor orders its rows. As a result, `zotero_get_item_fulltext` (local/hybrid mode) can return the text of a **trashed HTML snapshot** instead of the live PDF's content.

## Reproduction

1. Save a paper via the browser connector — you get a parent item with an HTML snapshot child. Zotero indexes it, writing `storage/<KEY>/.zotero-ft-cache`.
2. Attach the actual PDF, and move the HTML snapshot to the trash (a very common cleanup — the snapshot is just the publisher's landing page).
3. Call `zotero_get_item_fulltext` on the parent.

**Expected:** the PDF's text.
**Actual (when the PDF can't be parsed/resolved and the ft-cache fallback kicks in):** the trashed snapshot's `.zotero-ft-cache` — i.e. the landing page HTML text, not the paper.

Two independent problems in the same query:

1. **No `deletedItems` filter** — trashed attachments still participate both in the on-disk candidate scan and in the `.zotero-ft-cache` fallback loop of `_extract_fulltext_for_item`. Every other item-level query in `local_db.py` already excludes `deletedItems`; this iterator is the one exception.
2. **No `ORDER BY`** — the ft-cache fallback loop takes the first attachment with a cache in undefined SQLite row order, so even among live attachments an HTML snapshot's flat cache can win over the PDF's.

## Fix

`LEFT JOIN deletedItems` + `WHERE d.itemID IS NULL`, and `ORDER BY (ia.contentType = 'application/pdf') DESC, ia.itemID` so PDFs are yielded first. PR incoming with regression tests (trashed snapshot + live PDF, only-trashed attachments, `NULL` contentType, trashed parent).

feima3333:
Fixes #427.

## Problem

`_iter_parent_attachments` neither filters `deletedItems` nor orders its rows, so in local/hybrid mode `zotero_get_item_fulltext` can return a **trashed HTML snapshot's** `.zotero-ft-cache` (the publisher landing page text) instead of the live PDF's content.

Real-world sequence: save via browser connector (HTML snapshot child gets created and indexed) → attach the real PDF → trash the snapshot. If the PDF can't be parsed or resolved and the ft-cache fallback in `_extract_fulltext_for_item` kicks in, the trashed snapshot's cache wins — it still sits in `storage/<KEY>/` because emptying the trash hasn't happened yet.

## Change

One query, two guards:

- `LEFT JOIN deletedItems d … WHERE d.itemID IS NULL` — trashed attachment rows no longer participate in either the on-disk candidate scan or the ft-cache fallback loop. Every other item-level query in `local_db.py` already excludes `deletedItems`; this iterator was the one exception.
- `ORDER BY (ia.contentType = 'application/pdf') DESC, ia.itemID` — PDFs are yielded first, so when several live attachments have a ft-cache the PDF's flat cache is preferred over an HTML snapshot's, deterministically.

Scope note: only the attachment's *own* trash state is checked. Parent-trash inheritance is deliberately out of scope — callers reach this iterator through live-item paths that already exclude trashed parents (covered by a test documenting that behavior).

## Tests

New `TestIterParentAttachments` in `tests/test_local_db.py`, against a real temp sqlite DB:

- trashed HTML snapshot + live PDF → only the PDF is yielded
- PDF yielded before other types regardless of insertion order
- only trashed attachments → nothing yielded
- `NULL` contentType neither crashes the `ORDER BY` nor outranks a PDF
- trashed parent + live attachment → attachment still yielded (scope note above)

Without the fix, the first four fail; with it, all pass. `tests/test_local_db.py`, `test_local_db_discovery.py`, `test_fulltext_local_mode.py`, `test_fulltext_non_pdf_html.py`, `test_fulltext_display_cap.py` all green (57 passed).

### DIFF
--- a/src/zotero_mcp/local_db.py
+++ b/src/zotero_mcp/local_db.py
@@ -302,7 +302,14 @@ class LocalZoteroReader:
         return None
 
     def _iter_parent_attachments(self, parent_item_id: int):
-        """Yield tuples (attachment_key, path, content_type) for a parent item."""
+        """Yield tuples (attachment_key, path, content_type) for a parent item.
+
+        Explicitly trashed attachment rows (the attachment itself is in
+        deletedItems; parent-trash inheritance is not checked) are excluded,
+        and PDFs are yielded before other content types — otherwise a leftover
+        HTML snapshot's .zotero-ft-cache would win over the PDF's in
+        _extract_fulltext_for_item.
+        """
         conn = self._get_connection()
         query = (
             """
@@ -313,7 +320,9 @@ class LocalZoteroReader:
                    att.key as attachmentKey
             FROM itemAttachments ia
             JOIN items att ON att.itemID = ia.itemID
-            WHERE ia.parentItemID = ?
+            LEFT JOIN deletedItems d ON d.itemID = ia.itemID
+            WHERE ia.parentItemID = ? AND d.itemID IS NULL
+            ORDER BY (ia.contentType = 'application/pdf') DESC, ia.itemID
             """
         )
         for row in conn.execute(query, (parent_item_id,)):
--- a/tests/test_local_db.py
+++ b/tests/test_local_db.py
@@ -437,6 +437,111 @@ def test_get_key_group_map_includes_trashed_items_with_their_library(tmp_path):
     assert "DELETEDKEY1" not in result.excluded_keys
 
 
+# ---------------------------------------------------------------------------
+# _iter_parent_attachments: trashed attachments excluded, PDFs first
+# ---------------------------------------------------------------------------
+
+
+def _create_attachment_db(db_path: Path) -> None:
+    conn = sqlite3.connect(db_path)
+    conn.executescript(
+        """
+        CREATE TABLE items (
+            itemID INTEGER PRIMARY KEY,
+            key TEXT,
+            libraryID INT
+        );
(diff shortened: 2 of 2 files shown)


54yyyu:
Merged and released in 0.9.1. Thanks for this one, and for the ft-cache write-up in #427 that went with it. The reasoning about `_iter_parent_attachments` being the one item-level query without a `deletedItems` guard was right, and the regression tests made it easy to confirm: reverting just the `local_db.py` hunk fails four of the five.

One follow-up I'd take a PR on if you're interested. The PDF-first `ORDER BY` is deterministic, which is the important part, but it hardcodes a preference that `attachment_priority` made configurable in 0.8.0. The primary path routes through `pick_by_priority`; the ft-cache fallback now doesn't. Someone running `["markdown", "pdf", "html"]` gets different orderings from the two paths. Routing the fallback through the same chooser would close that.

Issue #403, opened 2026-07-28
`zotero_add_from_file` reports "File attached" without checking the `attachment_both` result — root cause behind the silent-attach failures (#278, #306, #399)


mikemt:
## Summary

`zotero_add_from_file` creates the parent item and files it into collections, then reports `File attached: <name>` — but the attachment never lands (no child attachment item, `numChildren: 0`). I hit this on every file in a large import, with a paid storage plan and plenty of quota, so it's not the 413/quota case from #399. It looks like the same silent-failure class as #278 / #306, but with a reproducible client-side cause.

## Environment

- zotero-mcp-server 0.6.2 (uv tool install; first hit on 0.6.0, retested on 0.6.2 — the attach code in `write.py` is unchanged between them), Python 3.14, macOS 15
- Web API mode (`ZOTERO_API_KEY` set, no `ZOTERO_LOCAL`), unlimited Zotero storage
- pyzotero 1.13.2

## Root cause (two parts)

**1. pyzotero's `attachment_both()` fails client-side**, returning the attachment in its `failure` list without raising. Standalone repro, no MCP involved:

```python
from pyzotero import zotero
z = zotero.Zotero(LIBRARY_ID, "user", API_KEY)
res = z.attachment_both([("paper.pdf", "/abs/path/paper.pdf")], parentid="XXXXXXXX")
# res == {"success": [], "failure": [<attachment payload>], "unchanged": []}
```

The failed payload has the full filesystem path in `filename` and `md5: None`. Verified against a fresh test item on 0.6.2's environment: `failure: 1`, `children: 0`. (I'll file this against pyzotero separately.)

**2. `write.py` never checks that result.** In `zotero_add_from_file` the return value of `attachment_both` goes straight into the success message:

```python
attach_result = write_zot.attachment_both([(display_name, file_path)], parentid=parent_key)
attach_info = f"File attached: {display_name}" + ...
```

So any upload failure — this one, quota 413s, network errors — is reported as success. This is why the failures in #278/#306/#399 are silent.

Separately, in local mode (`ZOTERO_LOCAL=true`) the attach step can never work: `attachment_both` starts with a template fetch that 404s (`http://localhost:23119/api/items/new` → "No endpoint found"), since the local API has no attachment/upload endpoints. That case could fail fast with a clear message instead.

## Working alternative (suggested fix shape)

[...]

PR #430, opened 2026-08-04
chore(deps): require pyzotero >=1.13.3 so attachment_both uploads work first-try

feima3333:
## Problem

pyzotero ≤1.13.2 has a bug in `attachment_both()`: it derives the stored filename from the full filesystem path, the Zotero Web API rejects the upload with **400 "Stored-file filename cannot contain a directory path"**, and pyzotero swallows the error, returning the payload in `failure` instead of raising.

Since #403, zotero-mcp detects this (`_describe_attach_failure`) and recovers via the two-step create+upload fallback in `_attach_and_verify` — so files do land and failures are no longer silent. But with the current `pyzotero>=1.8.0` floor, any environment resolving an old pyzotero pays for it on **every** attach (`add_by_url`, `add_by_doi`, arXiv PDFs, `add_from_file`, …): the primary `attachment_both()` call always fails client-side and every upload takes the fallback's extra API round trips.

Before #403 landed, on a stale install this was user-visible as *all* PDF attach operations silently failing while the tool reported success — which is how it was originally noticed (zotero-mcp 0.4.x + pyzotero 1.13.2 in the wild).

## Change

Raise the floor to `pyzotero>=1.13.3` (current PyPI release), where the filename handling is fixed and the primary `attachment_both()` path succeeds first-try. Comment in `pyproject.toml` records why the floor exists so it doesn't get relaxed accidentally.

The #403 fallback stays — it still covers the local-API 404 case and any future client-side rejection.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,7 +19,12 @@ classifiers = [
     "Operating System :: OS Independent",
 ]
 dependencies = [
-    "pyzotero>=1.8.0",
+    # 1.13.3 fixes attachment_both() putting the full filesystem path into the
+    # stored filename, which the Zotero API rejects with 400 ("Stored-file
+    # filename cannot contain a directory path") while pyzotero swallows the
+    # error — on older versions every upload silently falls back to the
+    # two-step retry in _attach_and_verify (#403).
+    "pyzotero>=1.13.3",
     "python-dotenv>=1.0.0",
     # PDF text extraction. Rust with prebuilt abi3 wheels for every platform
     # we support and no Python dependencies of its own. Pinned exactly while


54yyyu:
Merged and released in 0.9.1.

One correction I made in a follow-up (#434), for the record rather than as a criticism: the comment blamed `attachment_both()`, but that function is byte-identical between 1.13.2 and 1.13.3. I diffed the wheels to check. The actual fix is `Zupload._outgoing` in `_upload.py`, which sends `Path(filename).name` at upload step 0; step 1 (`_get_auth`) already used the basename, which is why it was never the failing half. The conclusion and the floor were both right, so the comment on main now just names the right function.

1.13.3 also attaches the server's rejection reason as an `error` key on `failure` entries. `_describe_attach_failure` still stringifies the whole payload, so there's a small improvement available there now if you want it.

PR #440, opened 2026-08-12
perf: parallel attachment extraction and a transient fulltext cache (#390)

### LINKED ISSUE #390: Optionally use parallel processing for PDF extraction

mronkko:
# Parallel attachment extraction and a transient fulltext cache

Closes #390. First of four PRs splitting #397, as requested — this is the pair you said you most wanted. Rebased onto current `main` (0.9.1) rather than carried over from `feature/batch-semantic-search`; the branch's first commit bundles four features across 19 files, so nothing there was cherry-pickable.

**In this PR:** the transient fulltext cache, and parallel attachment extraction.
**Not in this PR:** the provider refactor, streaming/rate limiting, and Gemini Batch — those follow separately.

The test-collection blocker is gone: `consider_namespace_packages` is not here, and neither is `tests/live/`. The suite collects and passes on this branch (1680 passed, 25 new). Only one existing test changed — one line, noted at the bottom.

---

## First, something that changed under the original PR

You resolved #390 differently on `main` while #397 sat open. Extraction no longer shells out to a pdfminer subprocess per file; `extract.py` now calls pdf-inspector in-process. That invalidates the design in #397, which used a **thread** pool on the explicit assumption that "each PDF already runs in its own subprocess". It doesn't any more, and pdf-inspector holds the GIL while parsing, so that implementation would have shipped almost no benefit.

This PR uses processes instead. Measured on 1000 PDFs sampled at random from a 16,586-PDF library (1429 MB, 10 cores), through the production path `extract_file(p, max_pages=50)`, against the pinned pdf-inspector 0.2.6:

| strategy | time | per PDF | speedup |
| :--- | ---: | ---: | ---: |
| sequential | 147.7 s | 147.7 ms | 1.00× |
| threads ×8 | 131.1 s | 131.1 ms | 1.13× |
| processes ×4 | 37.7 s | 37.7 ms | 3.92× |
| processes ×8 | 25.0 s | 25.0 ms | 5.91× |
| processes ×10 | 24.0 s | 24.0 ms | 6.16× |

Extrapolated across that library, the extraction phase goes from **~41 minutes to ~7**.

The GIL diagnosis is measured, not inferred. A pure-Python counter thread running alongside pdf-inspector parsing gets **18.2%** of its solo throughput — against **50.3%** next to a known GIL-holder (a Python busy loop, which splits the lock evenly) and **99.1%** next to `time.sleep` (which releases it). Amdahl with 18% GIL-free time predicts 1.16×/1.19× for 4 and 8 threads; measured was 1.16×/1.17×.

Worth saying plainly: this is a ~6× improvement to one phase, not the "days to minutes" claim in #397. That framing described the old subprocess-per-PDF world and no longer holds.

---

## Parallel extraction

`LocalZoteroReader.extract_fulltext_for_items()` fans parsing out over a `ProcessPoolExecutor`, controlled by `--extraction-workers N` or `semantic_search.extraction.workers`.

**The default stays 1** — byte-for-byte the existing sequential path. Opting in is a deliberate choice rather than a behaviour change for every user; happy to make it default to something higher if you'd prefer, the numbers above support it.

The division of labour keeps the risky parts in one process: sqlite, attachment resolution and the cache all stay in the parent, and workers receive only `(path, max_pages)`. The sqlite connection is never shared and nothing crosses the process boundary but two strings and an int. Target resolution is deliberately cheap — sqlite plus one `stat()` per attachment — so the pool only ever does parsing.

Two details worth flagging for review:

- The pool takes an `initializer` that re-applies the `zotero_mcp.extract` logger suppression. That suppression is parent-process state and means nothing in a worker, so without it parallel runs print extraction warnings that sequential runs hide. It isn't hypothetical: ~0.4% of real-world PDFs fail to parse (corrupt structure, a JPEG named `.pdf`).
- `_extract_worker` returns `""` rather than raising, so one bad file can't take down a worker.

## Transient fulltext cache

Extracted text is persisted under the config directory, keyed by attachment key + mtime + size, so a replaced file can never serve stale text. The page cap is recorded in the entry's profile, so lowering `pdf_max_pages` misses rather than serving over-long text. Attachment priority deliberately isn't in the profile — entries are keyed per attachment, so a changed priority resolves to a different key and misses naturally.

Entries are evicted per batch as embeddings land in ChromaDB, so the cache tracks what is still un-embedded rather than growing to the size of the library. `purge_stale()` sweeps abandoned entries and orphaned text files; `--clear-fulltext-cache` discards everything.

**It is off by default** and only the indexing path opts in. `zotero_get_item_fulltext` extracts under a *display* page cap, and letting it write would poison the index's cache with truncated text.

One behaviour that isn't obvious but matters: the `.zotero-ft-cache` fallback is cached too. An image-only PDF parses to an empty string and falls through to Zotero's own cache — so without this, every run re-parses exactly the files that are slowest to parse, purely to rediscover they're empty. On a 40-item sample this took a warm re-run from 3.6 s to 0.05 s.

---

## Notes for review

**One existing test changed.** `test_setup_openai_new_config_defaults_to_batch` gains a single `""` in its positional answer list, because setup now asks about worker count alongside the existing page-cap question. Nothing else in the suite was touched.

**Test doubles shaped two decisions.** `FakeReader` in `test_failed_marker_retry.py` implements `extract_fulltext_for_item(item_id)` and nothing else, and several tests build `LocalZoteroReader` / `ZoteroSemanticSearch` without `__init__`. Rather than make the doubles grow methods to accommodate a faster path, the batch call goes through a small documented shim in `semantic_search.py` that falls back to per-item calls, and both classes carry class-level attribute defaults. Say the word if you'd rather the doubles were updated instead.

**Verification.** Beyond the suite, parallel output was checked to be identical to sequential across a real library — same text, same `fulltextSource`, for every item — and the cache was verified to serve reruns by rewriting file contents while restoring mtime and byte count, so a hit returns the original text and a re-parse would return the replacement.

`pytest-httpserver` and the `release.yml` dependency list you flagged aren't needed here; they belong with `tests/live/`, which arrives with the rate-limiting PR.

### DIFF
--- a/src/zotero_mcp/cli.py
+++ b/src/zotero_mcp/cli.py
@@ -358,6 +358,13 @@ def main():
                                  help="Path to semantic search configuration file")
     update_db_parser.add_argument("--db-path",
                                  help="Path to Zotero database file (zotero.sqlite), overrides config")
+    update_db_parser.add_argument("--extraction-workers", type=int, metavar="N",
+                                 help="Parse N attachments in parallel during --fulltext "
+                                      "(default: semantic_search.extraction.workers, or 1 for "
+                                      "sequential). Capped at the CPU count")
+    update_db_parser.add_argument("--clear-fulltext-cache", action="store_true",
+                                 help="Discard the transient extracted-fulltext cache before "
+                                      "running, forcing every attachment to be re-parsed")
     openai_batch_group = update_db_parser.add_mutually_exclusive_group()
     openai_batch_group.add_argument("--openai-batch", dest="openai_batch", action="store_true",
                                    help="Submit OpenAI embeddings through the asynchronous Batch API")
@@ -574,9 +581,19 @@ def main():
             # Save the db_path to config file for future use
             _save_zotero_db_path_to_config(config_path, db_path)
 
+        if getattr(args, "clear_fulltext_cache", False):
+            from zotero_mcp import fulltext_cache
+
+            removed = fulltext_cache.clear_all(config_path=str(config_path))
+            print(f"Cleared transient fulltext cache ({removed} entries)")
+
         try:
             # Create semantic search instance with optional db_path override
-            search = create_semantic_search(str(config_path), db_path=db_path)
+            search = create_semantic_search(
+                str(config_path),
+                db_path=db_path,
+                extraction_workers=getattr(args, "extraction_workers", None),
+            )
             if args.openai_batch is True and search.chroma_client.embedding_model != "openai":
                 print("Error: --openai-batch requires ZOTERO_EMBEDDING_MODEL=openai", file=sys.stderr)
                 sys.exit(1)
--- a/src/zotero_mcp/fulltext_cache.py
+++ b/src/zotero_mcp/fulltext_cache.py
@@ -1,0 +1,266 @@
+"""Transient on-disk cache for extracted attachment fulltext.
+
+``update-db --fulltext`` runs in two phases: extract text from every changed
+attachment, then embed it. Extraction is cheap per file but not free across a
+library — roughly 100 ms per PDF, so a 16,000-item library spends around half
+an hour in phase one. Embedding is the phase that fails: rate limits, token
+limits, network errors, or the user pressing Ctrl-C. When it does, the
+extracted text is lost and the next run pays for phase one all over again.
+
+This module persists that text to disk so a rerun skips straight to embedding.
+Entries are evicted once their item's embedding has actually landed in
+ChromaDB — the cache exists only to survive the gap between the two phases,
+not as a long-lived store.
+
+Layout under ``~/.config/zotero-mcp/fulltext_cache/``:
+
+- ``index.json`` — maps cache keys to entry metadata (source, item_key, ...)
+- ``<hash>.txt`` — the extracted UTF-8 text, one file per entry
+
+Concurrency
+-----------
+Extraction runs in a :class:`~concurrent.futures.ProcessPoolExecutor` (see
+``extraction_workers``), but **only the parent process touches this cache**.
+Workers receive a file path and return text; the parent writes each entry as
+its future completes. Nothing here needs to be process-safe, and a lock in
+this module would not have provided that anyway — each worker is a separate
+interpreter.
+
+Writing incrementally rather than once at the end is what makes the cache
+useful: a run killed part-way through still leaves every completed extraction
+on disk for the next one. :data:`_INDEX_LOCK` guards against the parent
+mutating the index from more than one thread.
+"""
+
+from __future__ import annotations
+
+import hashlib
+import json
+import logging
+import os
+import threading
+from collections.abc import Iterable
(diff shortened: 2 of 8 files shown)


54yyyu:
Merged, thanks — this is the extraction + cache split I asked for on #397, and it's exactly the right scope. Verified locally before merging: process-pool isolation is clean (workers get only `(path, max_pages)`; sqlite, the cache index, and chroma writes all stay in the parent), invalidation keys on `mtime_ns + size + page-cap profile`, eviction only runs after a successful upsert, `workers == 1` is the exact historical sequential path, and the full suite passes with no regressions.

Two non-blocking items worth a fast-follow, since neither affects the default sequential path:

1. **`purge_stale()` is never called from any shipped path.** On the `--openai-batch` flow nothing evicts, so the "transient" cache accumulates the whole library's extracted plaintext until someone runs `--clear-fulltext-cache`, and crash-orphaned `.txt` files linger forever. Wiring `purge_stale()` into the `update-db` entry point (and evicting on batch apply) would make the transient promise hold on that path too.

2. **`BrokenProcessPool` amplification** in `extract_fulltext_for_items`: if one worker dies (say, OOM on a pathological PDF), every remaining pending future raises `BrokenProcessPool`, each is swallowed as empty text, and each of those items gets recorded `has_fulltext="failed"` — which the skip logic then treats as not-worth-retrying until the item changes or `--force-rebuild`. One transient failure can poison a large batch in a single run, where the sequential path only ever loses one item. Catching `BrokenProcessPool` specifically and either re-raising or falling back to in-process extraction for the affected items would cap the blast radius.

Happy to take both as a follow-up here if you'd rather focus on the rate-limiting and Gemini-batch splits.

---

mronkko:
Taking both of these — they'll come as their own PR shortly, so they won't hold up the rate-limiting and Gemini-batch splits.

Planned shape, so it isn't a surprise:

1. **`purge_stale()` wiring.** A best-effort call after a successful `update_database()` in the `update-db` handler, plus eviction on the batch-apply path. `import_openai_batch` lands embeddings via `upsert_embeddings` but never evicts, which is exactly what lets the cache grow to the size of the library on `--openai-batch`; mirroring what `_process_item_batch` already does on the realtime path closes it.

2. **`BrokenProcessPool` containment.** Catch it specifically instead of via the blanket `except Exception`, and on the first occurrence stop calling `.result()` on the remaining futures — drain them through the existing in-process `_extract_fulltext_for_item` instead. That is already the `workers == 1` path and handles the transient cache and the `.zotero-ft-cache` fallback identically, so it needs no new helper, and the blast radius becomes the single item that actually killed the worker rather than everything still pending.

Tests come with both: nothing in the suite induces a `BrokenProcessPool` today, and `purge_stale`'s age-based branch is currently unexercised.

Issue #441, opened 2026-08-12
if_exists dedup never matches by DOI on the Web API (works on local API)


mronkko:
### Summary

`if_exists="skip"` / `"file"` never dedupes by DOI when zotero-mcp talks to the
**Zotero Web API**. Every re-add creates a duplicate. The same code works
correctly against the **local** API, which is likely why this hasn't surfaced.

### Cause

`find_existing_items()` (`tools/_helpers.py`) locates an existing item with a
server-side quick search:

```python
zot.items(q=doi, qmode="everything", itemType="-attachment", limit=50)
```

The web API's quick search only indexes **title, creator, and year**.
`qmode=everything` extends that to attachment full-text content and notes — it
does *not* search arbitrary metadata fields. There is no `qmode` on the web API
equivalent to the desktop app's "All Fields & Tags", so a DOI query can never
match, and `find_existing_items` returns `[]` for an item that plainly exists.

### Reproduction

Group library, 25 items added by DOI, then the identical 25 DOIs re-added with
`if_exists="skip"`. Result: **0 of 25 deduped**, library went 25 → 50 parents.

Probing one item that is definitely present (`10.1002/smj.3597`, stored `DOI`
field verified by direct read), 10+ minutes after it was written:

| `q=` | `qmode=everything` | `qmode=titleCreatorYear` |
|---|---:|---:|
| title words | 2 hits | 2 hits |
| author surname | 2 hits | 2 hits |
| **DOI** | **0 hits** | **0 hits** |
| DOI suffix only | 0 hits | 0 hits |
| publicationTitle | 0 hits | 0 hits |

Title and creator match instantly, so this is not search-index lag.

The equivalent query against the **local** API (`localhost:23119`) behaves
differently. Probing a DOI-bearing item in the local library (a different item —
the test group isn't synced to the desktop app):

| `q=<DOI>` | local API |
|---|---:|
| `qmode=everything` | **1 hit** |
| `qmode=titleCreatorYear` | 0 hits |

So locally `qmode=everything` *does* reach metadata fields, and over the web API
it does not. That makes this transport-specific rather than a bug in the
normalization or comparison logic.

The identifier paths that go through the same helper — `arxiv_id` (matched
against the `url`/`extra` fields), `isbn`, and `url` — look like they are
affected the same way for the same reason, though I only measured DOI.

### Possible directions

The web API has no field-scoped item query, so the quick search can't be fixed
in place. Options I can see:

[...]

PR #442, opened 2026-08-12
fix: three independent write-path correctness fixes

mronkko:
Three independent correctness fixes on the write path. Each is self-contained
and separately revertable; none depends on the others.

### 1. `attach_mode` now means what its description says

`attach_mode="none"` still downloaded and attached OA PDFs. The parameter was
threaded through the add path but never consulted at the point that actually
fetches the PDF, so the only thing it changed was the reported status string.
Callers asking for metadata-only imports got full PDF downloads.

### 2. Batch tag/extra updates retry on HTTP 412

`batch_update_tags` and `batch_update_extra` read an item, mutate it, and PUT it
back. Anything that bumps the item's version in that window — a sync, a
concurrent edit, Zotero itself — makes the PUT fail with 412 and the update is
silently lost for that item. Adds a bounded re-fetch/re-apply/re-send retry,
matching the pattern already used elsewhere in the file.

### 3. A rate-limited read no longer masquerades as data

pyzotero returns the HTTP 429 error body as `bytes` from every read method
instead of raising, so a throttled read looks like a successful one. Iterating
bytes yields ints, so callers get a list of integers where item dicts should be:

```
_helpers.py  find_existing_items
  AttributeError: 'int' object has no attribute 'get'
```

`len()` reports the body's byte count as a plausible-looking result count, so
nothing looks wrong until something dereferences an entry.

The path through pyzotero 1.13.4:

1. `errors.error_handler` special-cases 429 — it records the server-supplied
   backoff and returns instead of raising, documented as "leaving the caller to
   retry".
2. No caller retries. `_post_check` just returns, so the 429 response is handed
   back as if the request had succeeded.
3. `retrieve` then dispatches on the *response* `Content-Type`. Zotero sends
   `text/plain` for error bodies, which matches none of the JSON/Atom/BibTeX
   branches and falls through to `return retrieved.content`.

The fix has two halves, and the second is the one that does the work.

First, `_ZoteroClient`: a pyzotero subclass whose `_retrieve_data` reissues a
429'd read instead of returning it. Retrying is what `error_handler`'s docstring
already assumes — the backoff is recorded by the time it returns, and
`_check_backoff` sleeps for exactly that duration at the start of the next
request, so a bounded loop needs no timing logic of its own. After N attempts it
raises `ZoteroRateLimitedError`.

Second, the three client factories — `get_zotero_client`,
`get_local_zotero_client`, `get_web_zotero_client` — now construct that subclass
instead of `zotero.Zotero`. This reads like incidental tidying, but without it
the fix does nothing: an override only applies to instances of the subclass, so
while the factories keep handing back a plain `zotero.Zotero`, every client in
the process stays unguarded. Those factories are the only places a client is
constructed, which is what lets one change cover every read in the codebase — 87
call sites that would otherwise each need to recognise a `bytes` response on
their own. Nothing else about the client changes.

`find_existing_items` needs more than not-crashing. Every failure there degrades
to "no match" so the caller creates the item — right for a genuinely failed
search, but a throttled search hasn't answered the question, and reading it as
"not present" silently creates duplicates of items that are. It now lets rate
limiting propagate.

Filed upstream as urschrei/pyzotero#352, since any pyzotero consumer is exposed
to this.

`_ZoteroClient` is a workaround, but how it can be retired depends on the shape
of the upstream fix, so it carries a TODO saying so. If pyzotero retries
internally, the class is redundant and deletes cleanly. If pyzotero instead
*raises* on a rate-limited read, deleting it would reintroduce the bug by
another route: `super()._retrieve_data` would raise before the status check,
losing the retry, and `_is_rate_limited` wouldn't recognise pyzotero's own
exception class — dropping it back into the "treat as no match" path that
silently creates duplicates.

`test_rate_limit_guard.py` pins unguarded pyzotero's behaviour directly, so it
starts failing as soon as an installed version changes it, and the dependency
comment in `pyproject.toml` points at the TODO for whoever raises the floor.

### Testing

`1681 passed` on this branch. Each fix ships with regression tests
(`test_v021_fixes.py`, `test_version_conflict_retry.py`, `test_rate_limit_guard.py`).

---

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,21 +19,16 @@ classifiers = [
     "Operating System :: OS Independent",
 ]
 dependencies = [
-    # Below 1.13.3, upload step 0 sends the caller-supplied path as the
-    # stored-file filename; 1.13.3 adds Zupload._outgoing in _upload.py to send
-    # Path(filename).name instead (upstream urschrei/pyzotero#341). The API
-    # rejects the path with 400 ("Stored-file filename cannot contain a
-    # directory path") and pyzotero returns the payload under `failure` rather
-    # than raising, so on older versions every attach pays for the two-step
-    # fallback in _attach_and_verify (#403).
-    #
-    # No upper bound, but note before raising the floor: as of 1.13.4 every
-    # read method returns the HTTP 429 error body as bytes instead of raising
-    # or retrying (upstream urschrei/pyzotero#352), which client._ZoteroClient
-    # works around. When a release fixes that, tests/test_rate_limit_guard.py
-    # will start failing — read the TODO on _ZoteroClient before removing it,
-    # since one plausible upstream fix makes deleting it reintroduce the bug.
-    "pyzotero>=1.13.3",
+    # 1.13.5 retries a rate-limited read internally and raises
+    # TooManyRetriesError once the backoff is exhausted (upstream
+    # urschrei/pyzotero#352). Below it, every read method hands the HTTP 429
+    # error body back as bytes, which callers read as data — dedup searches
+    # most damagingly, where it looks like "no match" and duplicates the item.
+    # _helpers.find_existing_items relies on that exception, so this floor is
+    # load-bearing, not just preferred. (1.13.3 is also required, for sending
+    # Path(filename).name as the stored-file filename — upstream #341, without
+    # which every attach pays for the two-step fallback in _attach_and_verify.)
+    "pyzotero>=1.13.5",
     "python-dotenv>=1.0.0",
     # PDF text extraction. Rust with prebuilt abi3 wheels for every platform
     # we support and no Python dependencies of its own. Pinned exactly while
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,6 +26,13 @@ dependencies = [
     # directory path") and pyzotero returns the payload under `failure` rather
     # than raising, so on older versions every attach pays for the two-step
     # fallback in _attach_and_verify (#403).
+    #
+    # No upper bound, but note before raising the floor: as of 1.13.4 every
+    # read method returns the HTTP 429 error body as bytes instead of raising
+    # or retrying (upstream urschrei/pyzotero#352), which client._ZoteroClient
+    # works around. When a release fixes that, tests/test_rate_limit_guard.py
+    # will start failing — read the TODO on _ZoteroClient before removing it,
+    # since one plausible upstream fix makes deleting it reintroduce the bug.
     "pyzotero>=1.13.3",
     "python-dotenv>=1.0.0",
     # PDF text extraction. Rust with prebuilt abi3 wheels for every platform
--- a/src/zotero_mcp/cli_standalone.py
+++ b/src/zotero_mcp/cli_standalone.py
@@ -708,11 +708,11 @@ def build_parser() -> argparse.ArgumentParser:
     adoi = add_sub.add_parser("doi", help="Add item by DOI")
     adoi.add_argument("doi")
     _add_common_flags(adoi)
-    adoi.add_argument("--attach-mode", choices=["auto", "linked_url", "import_file"], default="auto")
+    adoi.add_argument("--attach-mode", choices=["auto", "linked_url", "import_file", "none", "required"], default="auto")
     aurl = add_sub.add_parser("url", help="Add item by URL")
     aurl.add_argument("url")
     _add_common_flags(aurl)
-    aurl.add_argument("--attach-mode", choices=["auto", "linked_url", "import_file"], default="auto")
+    aurl.add_argument("--attach-mode", choices=["auto", "linked_url", "import_file", "none", "required"], default="auto")
     afil = add_sub.add_parser("file", help="Add item from local file (.pdf/.epub)")
     afil.add_argument("--filepath", required=True)
     afil.add_argument("--title", help="Override title if metadata extraction misses")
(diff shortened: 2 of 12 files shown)


mronkko:
Cross-referencing #441, since the third fix here and that issue both land on
`find_existing_items` and are easier to judge side by side.

They're independent problems and neither blocks the other:

- **This PR** stops a *throttled* search from being mistaken for a successful
  one — pyzotero hands back the 429 body as `bytes`, and the dedup check either
  crashes on it or, once made defensive, reads it as "no match" and creates a
  duplicate.
- **#441** is that the dedup search doesn't work on the Web API *at all*: `q=`
  quick search only indexes title/creator/year, so a DOI query can never match
  and `if_exists` silently duplicates every time. It works against the local
  API, which is likely why it went unnoticed.

So this PR removes one way `find_existing_items` can silently duplicate, but
#441 is the larger one and needs a design decision (client-side DOI index,
saved search, or reporting honestly that `if_exists` can't dedupe over the web
API) rather than a patch. I didn't fold a fix for it in here, since it changes
behaviour well beyond the scope of these three fixes.

Happy to take a direction on #441 and send it separately.

mronkko:
https://github.com/urschrei/pyzotero/issues/352 has been fixed upstream. Once that is released, the workaround/guard against this bug can be removed.

54yyyu:
Reviewed and verified locally. Two of the three fixes are correct and well-tested; the third is broken against the pyzotero this repo actually installs, and it blocks the merge because four of the PR's own tests fail.

## Commit 1 (attach_mode) — correct

`"none"` short-circuits before any OA lookup; `"required"` raises `OaPdfRequiredError` for both "nothing found" and "found but undownloadable", and both call sites handle it (`add_by_doi` surfaces the error, `_create_and_attach` records a failed batch entry). `_add_by_arxiv` now honors `"required"` too. `"import_file"` behavior is unchanged.

## Commit 2 (412 retry) — correct

Bounded at 3 attempts, re-fetches the item per attempt, retries only `PreConditionFailedError`, and both batch call sites keep exhaustion a per-item skip rather than an aborted batch. One non-blocking caveat: the mutate closures capture the pre-retry `current_tags`/`new_extra`, so a retry re-imposes that snapshot — a tag added concurrently between attempts gets overwritten rather than merged. Same last-write-wins window existed before; fine to leave.

## Commit 3 (rate-limit guard) — broken on pyzotero ≥ 1.13.5

The masquerade bug this targets (429 body returned as bytes, urschrei/pyzotero#352) was fixed upstream in **1.13.5**: `_retrieve_data` now retries a 429 internally (3 attempts, honoring the backoff) and raises `pyzotero.errors.TooManyRetriesError` on exhaustion. Our floor is `pyzotero>=1.13.3` with no lockfile, so every fresh install gets 1.13.5+. Consequences, all reproduced here:

1. The `_ZoteroClient` override never fires — `super()._retrieve_data()` raises `TooManyRetriesError` before your `status_code != 429` check sees anything, so `ZoteroRateLimitedError` is dead code. (Your own TODO at the top of the override predicts exactly this hazard.)
2. `_is_rate_limited()` matches only `ZoteroRateLimitedError`, not `TooManyRetriesError` — so a persistently throttled `find_existing_items` still falls into the generic except → "treating as no match" → duplicate created. That's the very bug the commit message says it fixes.
3. Test run on this branch: `tests/test_rate_limit_guard.py` — **4 failed, 3 passed**, including `test_throttled_search_does_not_read_as_no_match` and `test_unguarded_pyzotero_would_have_returned_bytes` (unguarded pyzotero now raises instead of returning bytes).

Suggested fix: raise the floor to `pyzotero>=1.13.5`, drop `_ZoteroClient`/`ZoteroRateLimitedError` entirely, teach `_is_rate_limited` to recognize `TooManyRetriesError` (and arguably `TooManyRequestsError`), and rewrite `test_rate_limit_guard.py` against the upstream exception. That shrinks the commit to a few lines and gets you the dedup protection for real.

Happy to merge as soon as that lands — or if you'd rather, split commits 1+2 out and they can go in immediately. Note #443 stacks on this branch, so it inherits the same four failures until this is fixed.

---

mronkko:
You were right, and the release landed the same day: pyzotero **1.13.5** was tagged
2026-08-12 at 17:10 UTC, about five minutes after my comment above saying the fix
was upstream and waiting on a release. Issue #352 closed at 16:48 by commit
`144a4d97`, "Retry rate-limited requests instead of returning the 429 error body".

So this is the second of the two shapes the `_ZoteroClient` TODO warned about — the
dangerous one — and your diagnosis is exactly right. I reproduced your **4 failed,
3 passed** locally before changing anything; my machine had 1.13.4 installed, which
is the only reason the suite was green here and red for you. `uv.lock` is gitignored,
so nothing pinned it.

Fixed in `825fe1d`, which takes the "delete it" branch rather than the "keep the
retry" one:

- `_ZoteroClient`, `ZoteroRateLimitedError`, `_MAX_RATE_LIMIT_RETRIES` and
  `_is_rate_limited` are all gone; the three factories point back at `zotero.Zotero`.
  Upstream retries `MAX_RETRY_ATTEMPTS = 3` times honoring `_check_backoff()`, so
  ours was the same bound over the same backoff and loses nothing.
- What survives is the invariant that actually mattered, now one ordered `except` in
  `find_existing_items`: a throttled search propagates, every other failure still
  degrades to "no match". No helper, no name-based fallback.
- Floor raised to `pyzotero>=1.13.5`. Load-bearing rather than preferred — below it
  the exception doesn't exist and the dedup guard silently goes back to duplicating.

One detail on your "and arguably `TooManyRequestsError`": I matched both, but it is
unreachable for 429 today. `error_handler`'s 429 branch either records the backoff
and returns (so `_retrieve_data`'s loop retries, then raises `TooManyRetriesError`)
or, with no backoff header, raises `TooManyRetriesError` immediately — either way it
returns or raises before the `ERROR_CODES` lookup that maps 429 to
`TooManyRequestsError`. It costs one tuple entry and is the semantically correct
class, so I left it in.

`test_rate_limit_guard.py` is rewritten against the upstream exception. I kept the
tests that pin *our* behaviour — above all that a throttled `find_existing_items`
raises rather than returning `[]`, which is the one whose regression is a silent
duplicate — added one for the no-backoff-header path, and dropped the two that
pinned pre-1.13.5 pyzotero internals. The bytes-returning tripwire did its job: it
fired, which is how you caught this.

Also in the commit: `test_local_http1_transport.py`'s `Zotero.__init__` spy now sets
`.client`. 1.13.5 adds a `__del__` that reads it, so the spy raised during GC and
surfaced as a `PytestUnraisableExceptionWarning`.

`1680 passed` on the branch under 1.13.5.

Your two non-blocking notes on commits 1 and 2 I've left as they are — agreed on the
412 retry's last-write-wins window being pre-existing and fine.

#443 has been rebased onto none of this yet; I cherry-picked the same commit there so
it isn't blocked meanwhile, and will do the real rebase once this merges.

PR #443, opened 2026-08-12
feat: add multiple DOIs/URLs/ISBNs in one call, with batched CrossRef and Zotero round-trips

mronkko:
Lets `zotero_add_item` take several DOIs/URLs/ISBNs in one call, and makes that
batch path cost a bounded number of API round-trips instead of one set per item.

Adding 244 DOIs previously meant 244 tool calls, each doing its own client
resolution, item-template fetch, CrossRef lookup and `create_items()` POST — and
holding the Zotero API lock across the whole thing.

### What's in it

| | |
|---|---|
| **Multi-value input** | `doi`/`url`/`isbn` accept a list or a comma/newline-separated string; one bad token never fails its neighbours |
| **A2/A3** | client + collection resolution and item-template fetches happen once per call, memoized by item type, not once per token |
| **A4** | `create_items()` POSTs batch at 50, with one bulk collection-membership read per chunk instead of one `item()` GET per item |
| **A5** | CrossRef metadata for a whole batch resolves in one `/works?filter=doi:…` request per 50 DOIs |
| **Lock scope** | the API lock is taken around the Zotero calls that need it, not held across CrossRef lookups and PDF downloads |
| **429 handling** | a rate-limited Zotero read is retried and then raised, rather than returned as a `bytes` error body (urschrei/pyzotero#352) |

### On A5 specifically

Measured, all HTTP 200, nothing missing:

| DOIs | Crossref URL length | time |
|---:|---:|---:|
| 10 | 343 | 2.30s |
| 25 | 807 | 1.75s |
| 40 | 1362 | 3.29s |
| 60 | 2101 | 2.88s |

Four traps, each verified against the live API:

- **`rows` is sent explicitly.** CrossRef pages at 20 by default, so a 50-DOI
  filter silently returns the first 20 and the rest look like they aren't in
  CrossRef. Confirmed: rows omitted returns 20 items while `total-results`
  reports 50.
- **Chunks at 50, not 60.** 60 was measured working, but CrossRef documents no
  URL-length or filter-count ceiling, so this stays off an unmeasured cliff. 50
  also matches the Zotero POST chunk.
- **Returned DOIs are diffed against requested ones, case-insensitively.**
  CrossRef echoes canonical case. The set difference is the not-found list and
  gets per-entry results — this is also what guards the `doi` filter's
  documented-single/undocumented-multi OR semantics, so nothing is ever
  silently dropped.
- **`mailto` is sent unconditionally**, defaulting to the same project noreply
  identity `discovery.py` already sends to OpenAlex, overridable via
  `ZOTERO_MCP_CONTACT_EMAIL`. The single-DOI path was previously in the
  anonymous pool whenever that variable was unset.

### Measured end to end

Live import into a real Zotero group, `attach_mode="none"`:

| | per DOI | notes |
|---|---:|---|
| 0.9.1, sequential | 3.10s | also downloaded PDFs despite `attach_mode="none"` |
| this branch, batched | **0.89s** | 25/25 created, zero 429s, 0 PDFs |

### Testing

`1722 passed`. `test_batch_crossref.py` pins request *shape* rather than timing:
request count, chunk boundaries, the explicit `rows` value, case-insensitive
matching, per-entry not-found reporting, whole-chunk failure reported per DOI,
`mailto` always present, and the bounded 429 retry. The shared CrossRef test
fake pages by `rows` exactly as the real API does, so dropping that param fails
the suite the same way it failed in production.

### Note on overlap

Three commits here — `attach_mode`, the 412 retry, and the 429 guard — are the
same ones in #442, which targets them at `main` on their own because they're
independent of the batching work. If that lands first I'll rebase this branch
onto it; the batching itself is the remaining five commits.

---


54yyyu:
Reviewed the delta on top of #442 (the batch feature itself — the three #442 commits get their review over on #442, and this branch inherits the rate-limit-guard blocker described there).

The architecture here is genuinely good: chunking has no off-by-ones (every index in every 50-item chunk gets exactly one result across create-exception / non-dict / per-index-failed / created cases), the requested-token ↔ pending-list ↔ created-key alignment is sound, `rows=` correctly defeats CrossRef's 20-row default page, per-chunk failures stay isolated, and repeated DOIs collapse to one fetch. The new tests all pass (1692 total on the branch, only known-environmental failures plus the inherited #442 four). Three things block it, all verified by execution rather than just reading:

**1. Unconditional comma-split mangles single identifiers that contain commas.** `_split_multi_value` splits every string on `,` before any validation, and `add_by_url`/`add_by_doi`/`add_by_isbn` call it unconditionally. A single URL like `https://example.com/page?ids=1,2` — previously handled fine — becomes two webpage items: the truncated URL and a junk item with `url="2"`. Commas in query strings are common, and DOIs may legally contain commas too (`_normalize_doi`'s own `\S+` accepts them). `detect_source_type` is careful to classify a comma list as a batch only when *every* token validates — but the adders don't apply that gate, so its care protects nothing. Fix is small: only treat the input as a batch when all comma-tokens validate as the expected identifier type (mirroring `detect_source_type`), otherwise treat it as one identifier.

**2. The arXiv-via-CrossRef OA-PDF fallback is silently dead.** Old `add_by_doi` passed `crossref_metadata=cr` into `_try_attach_oa_pdf`; the shared worker never does — `crossref_metadata` no longer appears anywhere in write.py — and `_try_arxiv_from_crossref(None)` returns None. DOIs whose only OA copy is the arXiv preprint referenced in the CrossRef record now silently fail to attach, and with `attach_mode="required"` they *error* where they used to succeed. The `cr` message is in hand in `_build_one_doi_item_data`; carry it through the pending payload.

**3. A CrossRef 400 collapses the whole batch.** `_fetch_doi_metadata_batch` does `.get("message", {}).get("items")` with no status check; on validation failures CrossRef returns `"message": [...]` (a list), producing `AttributeError: 'list' object has no attribute 'get'` — which only `ValueError` handling doesn't catch, so the entire multi-DOI call dies with one opaque error instead of per-DOI results. That breaks this PR's own "one bad DOI never fails its neighbours" promise.

Minor, take or leave: repeated identical *new* DOIs in one `skip`/`file` batch all get created (dedup runs wholly in phase 1 before any creation); `_format_multi_result` says "Added N" even when some entries failed; and a JSON-object string with `source_type="doi"` raises from `_split_multi_value` outside the try, instead of returning the usual error string.

One more heads-up: #441's fix just landed on the maintenance branch and adds a `title_hint` fallback to `find_existing_items` (the web API's quick search can't see the DOI field, so dedup needs a title-confirmed second search). Your phase-1 dedup here runs before metadata exists — after rebasing, it's worth wiring the post-fetch titles into the dedup pass so batch `if_exists` actually works over the web API, which is this PR's main audience.

With 1–3 fixed (and #442's guard rewritten against pyzotero ≥ 1.13.5), this is a merge.

---

mronkko:
Thanks — all three blockers confirmed by reproduction and fixed, plus all three
minors. Each fix has a test written before it that fails for the stated reason.
`1763 passed`.

The inherited #442 guard is fixed there and cherry-picked here as `80f357b`, so this
branch isn't blocked on that merge; the real rebase follows once #442 lands.

### 1. Comma-split — `b6d0b0c`

Reproduced exactly: `https://example.com/page?ids=1,2` created two items, the second
titled `2`.

**I didn't take the suggested fix as written.** Gating on "all comma-tokens validate"
breaks this PR's other promise — two ISBNs where the second fails its checksum stop
being a batch, and the good one is never added. `test_partial_failure_reports_both`
caught it. A comma can never occur inside a valid ISBN, so there is nothing to
protect there.

So a line splits when **either** every token validates (unambiguous) **or** the line
as a whole is not a valid identifier (nothing to protect, and splitting is the only
reading that can produce anything). It stays whole only when the tokens don't all
validate *and* the line does — which is precisely the `?ids=1,2` case. Newlines still
separate unconditionally, and the rule applies per line, so
`"https://a.com/x?ids=1,2\nhttps://b.com"` is two URLs rather than three.

URL was the one type with no normalizer to gate on; the heuristic is extracted as
`_looks_like_url` and now used by both `detect_source_type` and the split, so they
agree by construction. It also rejects raw whitespace, which is what lets the gate
tell `"https://a.com, not a url"` from a single URL containing a comma.

### 2. arXiv-via-CrossRef — `f77c8ca`

Confirmed dead. `_build_one_doi_item_data` already had `cr` in hand and discarded it;
it now carries it in the pending payload. `_create_and_attach_batch` takes an optional
`crossref_by_doi` map and looks up against the DOI it already re-derives from
`item_data` — keyed by DOI rather than a list parallel to `item_datas`, since a
parallel list is one more thing that has to stay aligned across chunk boundaries. It
stays optional for the bibtex/CSL-JSON callers.

Tested where the regression actually happened: that each entry's own message is
forwarded on both the single and batch paths, and end-to-end that the preprint PDF is
attached when it's the only source with a copy. The existing cascade tests all passed
`crossref_metadata=None` directly, which is why nothing caught it.

### 3. CrossRef 400 — `30d6cf9`

Reproduced: `Error adding by DOI: 'list' object has no attribute 'get'`, whole call
dead. Now checks the status before parsing (as the single-DOI path does) and requires
`message` to be a dict. Both route into the existing per-chunk error path, so chunk
isolation holds — there's a test where chunk 1 400s and chunk 2 still imports.

Fixed locally rather than in `_crossref_get`, since `_fetch_one_doi_metadata`
deliberately handles its own 404 there.

**One existing test changed expectation.** `test_batch_request_failure_reported_per_doi`
simulated a hard 404 and asserted "DOI not found on CrossRef" per entry — which only
passed because `{}` parsed through to an empty item list. That's the same
wrong-answer-vs-reported-failure distinction as the blocker: `/works` answers "no
matches" with 200 and an empty list, so a 404 means the request broke, not that
CrossRef lacks the DOIs. It now asserts the honest error and keeps its real point, one
line per requested DOI. Same reasoning made a non-dict `message` a reported failure
rather than falling through to an empty item list.

### Minors — `11d9432`

**Repeated identifiers.** Collapsed before any work rather than threading a
created-this-call index through phase 3 — that also drops the redundant CrossRef and
Zotero round-trips. Identity is the normalized identifier: DOIs case-insensitively,
ISBN-10 and ISBN-13 of one book collapse via `_normalize_isbn`, URLs exactly after
stripping (no normalizer exists, and inventing trailing-slash or query-order
equivalence would silently drop URLs someone meant to add separately). Unnormalizable
tokens stay separate so each keeps its own error. Output still carries one line per
requested token; a repeat's line names the entry it duplicates instead of implying a
second item exists.

**"Added N".** Now added-of-requested plus a breakdown of reused / repeated / failed.

Worth flagging: that count is derived by matching the prefixes each adder opens a
result block with, because these three return pre-rendered strings — it's the only
success signal the recursive URL/ISBN paths expose. My first cut missed that
`add_by_url`'s webpage branch says "Created webpage item for:" rather than
"Successfully added" and reported a fully successful batch as `0 of 2`. The markers
now live in one place and each adder's wording is pinned by a batch test, but this is
a stopgap; the real fix is the convergence `_format_multi_result`'s docstring already
describes.

**JSON-object raise.** Each `_split_multi_value` call moved inside its adder's `try`.

### Not done

#441's `title_hint` — it isn't on any branch I have, so wiring post-fetch titles into
phase-1 dedup waits until I rebase onto the maintenance branch. Agreed that's what
makes batch `if_exists` actually work over the web API, which is this PR's audience.

54yyyu:
Heads-up before you spend time on this: `main` moved a lot today and the write path is the part that moved. Merged since this branch's base: #442, #477, #479, #480, #466, #449, #450, #451, #417, #464.

Trying it locally gives 9 conflicts in `src/zotero_mcp/tools/write.py` — all of them in the add paths, and the other side of every one is your own work (#479's embedded-citation reader and #480's CrossRef type map landed into the functions this PR restructures). I'd rather you resolve those than have me guess at which shape you intended, since the batch rewrite changes the surrounding control flow.

Could you rebase onto current `main`? Everything else about the PR looks good to me — the CrossRef `rows`/chunk-at-50/case-diff analysis in particular is the kind of thing that usually gets found in production instead of in review. Happy to merge as soon as it's clean.

mronkko:
I will rebase and update the PR. Might take a few days because of teaching responsibilities.

mronkko:
Rebased onto `4029b6b`. Clean merge, and the branch is linear on current `main`.

Four commits dropped themselves — `attach_mode`, the 412 retry, and both 429
commits are byte-identical to what #442 already landed, so `git rebase` skipped
them as previously-applied. Nine of mine replayed, twenty conflict hunks in
`write.py` plus one in the changelog.

### The three that needed a decision rather than a merge

**`add_by_url`'s generic-webpage branch is no longer lock-safe as a whole.**
The lock-narrowing commit wrapped that branch entirely, on the stated grounds
that it contained no third-party network work. #479 put a fetch of an arbitrary
publisher host in the middle of it, so taking my side of that hunk would have
reintroduced exactly the starvation the commit exists to remove — and taking
yours would have dropped the narrowing. The lock now covers only the Zotero
calls either side of the page fetch, and `_add_from_embedded_metadata` is
decorated itself, since it lost its cover when `add_by_url`'s decorator went.

**#480's `type_note` was mis-aligned by git against my `OaPdfRequiredError`
handler**, which would have silently swapped the unmapped-type warning for the
attach-failure message. It threads through the batched path now:
`_crossref_to_item_data` returns it, `_build_one_doi_item_data` carries it in
the payload, `_render_doi_create_result` renders it.

**#477's thin-record fallback had to be given a place in the batch path.**
`supplemental` describes one specific page, so it cannot be handed to a batch;
each thin record in a batch reads its own landing page instead, per DOI, outside
the lock.

### Verification

`2524 passed`, against `2441` on `4029b6b` itself. One failure, on both trees
identically and before any of my commits: `test_issue_455_toc_stdout_pollution`
can't reproduce the `fitz` deprecation notice in my environment, so its
"pollution really is present" precondition fails. Not related to this branch.

Single-DOI output is byte-identical to `main`'s, diffed rather than asserted —
including the unmapped-type case, which is the one `type_note` governs.

Two commits are new.

`test(write): pin the two behaviours the 0.10.0 rebase put in contact` — neither
guarantee above had a test, because before the rebase neither situation existed.
Both were checked against a mutated tree rather than assumed: dropping the batch
fallback fails the thin-record tests, and moving the page fetch back inside the
lock fails the lock one.

`docs(write): stop claiming the version retry guards concurrent adds` — while
justifying the lock decision above I went to check the narrowing's own rationale,
"A6's version-checked retry is what guards concurrent writes, not this lock," and
it doesn't hold for the case the narrowing exposes.
`_update_item_with_version_retry` is wired into `batch_update_tags` and
`batch_update_extra` only, neither on an add path, and it couldn't help if it
were: it catches a stale version on an item that already exists, while two
concurrent `create_items()` POSTs for one DOI yield two new keys and no version
to conflict on. The comment now states the race plainly instead of pointing at a
guard that cannot close it. Whether adds should be serialized at all seems like a
project-level question rather than one for this branch — I'll open a separate
issue for it rather than widen this PR.

### Still not done

#441's `title_hint` — it's on `claude/project-maintenance-7vgy7g`, not on
`main`, so rebasing onto `main` didn't unblock it after all. Wiring post-fetch
titles into phase-1 dedup still needs that to land first. Happy to do it as a
follow-up the moment it does.

mronkko:
Two documentation follow-ups pushed, no behaviour change.

**The changelog entry was describing the branch as it was at the first commit.**
It said each identifier is "added independently via the existing single-item
path", which was true before A2–A5 and not after — that's precisely what those
commits replaced. It now states the round-trip counts instead: 244 DOIs costing
5 CrossRef requests, 5 creation POSTs and at most 11 template fetches, against
244 of each before.

I deliberately did not quote the end-to-end 3.10s → 0.89s figure from the PR
description there. Part of that gain came from the `attach_mode` fix, which
shipped separately in 0.10.0 via #442, so putting it under this entry would
count it twice.

**The concurrency note now points at #486**, which I opened rather than widening
this PR: `_update_item_with_version_retry` is wired into `batch_update_tags` and
`batch_update_extra` only, so the "version-checked retry guards concurrent
writes" rationale doesn't cover the creation path — and structurally can't, since
two concurrent `create_items()` POSTs for one DOI produce two keys with no
version to conflict on. The comment states the race plainly now and defers the
question of whether adds should be serialized at all.

`2524 passed`, unchanged. The one failure is still
`test_issue_455_toc_stdout_pollution`, which fails identically on pristine
`main` in my environment and is unrelated to this branch.

PR #445, opened 2026-08-12
perf: import submodules without loading the MCP server

mronkko:
## Problem

`zotero_mcp/__init__.py` imports the server eagerly:

```python
try:
    from .server import mcp  # noqa: F401
except ImportError:
    pass
```

so *any* `import zotero_mcp.<submodule>` pays for the whole server before the submodule's own code runs — FastMCP, the MCP SDK, pydantic, pyzotero, bibtexparser and unidecode all get pulled in.

Measured on 0.9.1 (`python -X importtime`):

| | cumulative | modules in `sys.modules` |
|---|---|---|
| `from zotero_mcp.schema import valid_fields` | **~1.73 s** | ~1480 |
| `zotero_mcp.schema` itself | 8 µs | — |
| `import pyzotero.zotero`, for scale | ~0.29 s | — |

All of the cost is the package `__init__`. `schema.py` is deliberately stdlib-only — its whole point is that base-field resolution keeps working offline, from the vendored table or the on-disk cache.

This matters for consumers that use the package **as a library** rather than as a server. `zotero_mcp.schema.valid_fields` is the natural way to filter fields before a batch write (the Zotero API fails an entire 50-item POST on one unknown field key), but reaching it today requires the full server stack to be installed and importable.

## Change

**1. `mcp` resolves lazily** via a PEP 562 module `__getattr__`.

`from zotero_mcp import mcp` is unchanged — it just imports the server on first access. It still degrades to `AttributeError` rather than propagating, matching the previous `try/except ImportError` behaviour when optional server deps are absent. Nothing inside the package uses `from zotero_mcp import mcp`; internal callers already go through `zotero_mcp.server`.

After: **~9 ms, 75 modules**, with none of `fastmcp` / `mcp` / `pydantic` / `pyzotero` / `bibtexparser` loaded.

**2. `_normalize_doi` gets a public home.**

It was the only canonical DOI handling the package exposes, but private, in a private module, and sitting behind `requests` + `pyzotero`. A downstream consumer wanting the same canonicalisation the server uses has to write `from zotero_mcp.tools._helpers import _normalize_doi` and pin itself to an implementation detail.

Moved verbatim to a new stdlib-only `zotero_mcp/identifiers.py` as `normalize_doi`. `tools/_helpers._normalize_doi` is kept as an alias, so every existing caller and the existing assertions in `tests/test_shared_helpers.py` are untouched — which is what demonstrates the move is transparent.

Happy to drop this half if you'd rather keep the surface small; part 1 stands alone.

## Tests

- `tests/test_lightweight_imports.py` — asserts the server packages stay out of `sys.modules` for stdlib-only submodules. Runs each import in a **subprocess**, since import cost is only observable on a cold interpreter (by the time the suite runs, `conftest` has already imported half the package). Also covers `from zotero_mcp import mcp` still resolving to the same object, `dir()`, and unknown-attribute behaviour.
- `tests/test_identifiers.py` — covers `normalize_doi` directly, including case preservation (Scite echoes back what it is given) and the trailing-punctuation handling.

Full suite: **1685 passed**. `ruff check` is clean on every file this touches (the repo has 120 pre-existing findings elsewhere, left alone).

## Context

Found while auditing a downstream plugin ([mronkko/claude-academic-research](https://github.com/mronkko/claude-academic-research)) that imports `zotero_mcp.schema.valid_fields` from a `uv run` script with PEP 723 inline dependencies. It was paying the full server import on every invocation of a batch Zotero importer, and declaring the entire server dependency tree to do it.

### DIFF
--- a/src/zotero_mcp/__init__.py
+++ b/src/zotero_mcp/__init__.py
@@ -4,12 +4,52 @@ Zotero MCP - Model Context Protocol server for Zotero
 This module provides tools for AI assistants to interact with Zotero libraries.
 """
 
+from typing import TYPE_CHECKING
+
 from ._version import __version__ as __version__
 
-try:
-    from .server import mcp  # noqa: F401
-except ImportError:
-    pass
+if TYPE_CHECKING:  # pragma: no cover - type checkers only
+    from .server import mcp as mcp
+
+__all__ = ["__version__", "mcp"]
+
+
+def __getattr__(name: str):
+    """Import ``mcp`` lazily (PEP 562).
+
+    ``from .server import mcp`` at module scope made *every* import of any
+    submodule pay for the whole server: FastMCP, the MCP SDK, pydantic,
+    pyzotero, bibtexparser and unidecode all get pulled in before the
+    submodule's own code runs. Measured on 0.9.1, ``from
+    zotero_mcp.schema import valid_fields`` cost ~1.45 s and ~1480 modules,
+    of which ``schema`` itself was 8 microseconds -- it is a stdlib-only
+    module whose whole point is offline field resolution.
+
+    That matters for consumers that use this package as a library rather
+    than as a server: schema lookups, DOI normalisation and the CLI's
+    lighter subcommands should not require the server's dependency tree to
+    be importable, let alone imported.
+
+    ``from zotero_mcp import mcp`` keeps working unchanged, and still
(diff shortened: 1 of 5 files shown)


mronkko:
This might be a a bit of an edge case, but I see no downsides in this change.

PR #461, opened 2026-08-17
fix(scite): render editorial notices using the real /papers field names

Yeyang28:
## Bug

Scite's `/papers` endpoint returns `editorialNotices` entries shaped `{status, date, noticeDoi, doi}`. Live example for `10.1038/s43586-020-00001-2`:

```json
[{"status": "Has correction", "date": "2021-2-3",
  "noticeDoi": "10.1038/s43586-021-00017-2",
  "doi": "10.1038/s43586-020-00001-2"}]
```

All three render sites — `_format_editorial_notices` (used by `scite_enrich_search`), `scite_enrich_item`, and `scite_check_retractions` — read only `type`/`editorialNoticeType` and `sourceDoi`/`source`, none of which exist in the response. So every real notice renders as an empty:

```
**Notice**: https://doi.org/
```

with no notice type and no notice DOI — a user running a retraction screen can't tell whether a flag is a retraction or a routine correction without querying the Scite API by hand (which is how I caught this: a 110-item library screen returned two flagged items with blank notices; both turned out to be benign corrections only after checking `api.scite.ai/papers/<doi>` directly).

This is the same API-contract bug class as the `POST /papers` body shape and lowercased DOI keys fixed in #331/#353 — the existing tests fabricate the `{type, sourceDoi}` shape, so they pass while the live API renders empty.

## Fix

At all three sites: read `status` and `noticeDoi` first (old keys kept as fallbacks so any older/other response shapes still render), include the notice `date`, and print `(no notice DOI)` instead of a dangling `https://doi.org/` when no DOI is present. Output now:

```
**Has Correction** (2021-2-3): https://doi.org/10.1038/s43586-021-00017-2
```

## Tests

- `test_editorial_notices_render_real_api_field_names` — real API shape through the formatter (fails on main).
- `test_check_retractions_renders_real_api_notice_fields` — end-to-end `check_retractions` output contains notice type + notice DOI (fails on main).
- `test_editorial_notices_legacy_shape_still_renders` — `{type, sourceDoi}` fallback guard.

`pytest tests/test_scite.py`: 6 passed (3 pre-existing + 3 new).

### DIFF
--- a/src/zotero_mcp/tools/scite.py
+++ b/src/zotero_mcp/tools/scite.py
@@ -62,10 +62,25 @@ def _format_editorial_notices(notices: list[dict]) -> list[str]:
     """Format editorial notices as warning lines."""
     lines = []
     for notice in notices:
-        ntype = notice.get("type", notice.get("editorialNoticeType", "notice"))
+        # Scite's /papers endpoint returns notices shaped
+        # {status, date, noticeDoi, doi}; older {type, sourceDoi} shapes are
+        # kept as fallbacks.
+        ntype = (
+            notice.get("status")
+            or notice.get("type")
+            or notice.get("editorialNoticeType")
+            or "notice"
+        )
         ntype = ntype.replace("_", " ").title()
-        source_doi = notice.get("sourceDoi", notice.get("source", ""))
-        lines.append(f"**{ntype}**: https://doi.org/{source_doi}")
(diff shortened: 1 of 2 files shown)


54yyyu:
Merged into `main` (commit 141abf3). Thanks for tracking down the real `/papers` field names.

PR #466, opened 2026-08-20
fix: accept Crossref's CSL JSON in csl_json_to_zotero

### LINKED ISSUE #465: csl_json_to_zotero fails on Crossref CSL JSON: arrays crash .strip(), and Crossref type names fall through to document (dropping journal, volume, issue, pages)
`csl_json_to_zotero` cannot consume Crossref's CSL JSON, which is where most CSL JSON in the wild comes from.

Two independent defects in `src/zotero_mcp/citation_import.py`, both firing on the output of Crossref's CSL transform endpoint:

```
https://api.crossref.org/works/<doi>/transform/application/vnd.citationstyles.csl+json
```

That is the payload `zotero-cli add csl-json` users pipe in and what any DOI-driven import reaches for. Neither defect is reachable from a hand-written CSL fixture, which is why the existing suite is green (see "Why the tests miss both" below).

Reproduced against `main` @ f331646 on 2026-08-20, with live Crossref responses.

## Reproduction

```python
import json, urllib.request
from zotero_mcp.citation_import import csl_json_to_zotero
from zotero_mcp.schema import valid_fields

def template(t):
    return {"itemType": t, **{f: "" for f in valid_fields(t)},
            "creators": [], "tags": []}

url = ("https://api.crossref.org/works/10.1016/j.jbusvent.2019.105970"
       "/transform/application/vnd.citationstyles.csl+json")
req = urllib.request.Request(url, headers={"User-Agent": "repro (mailto:you@example.com)"})
csl = json.loads(urllib.request.urlopen(req, timeout=30).read())

print(repr(csl["type"]), repr(csl["ISSN"]))
# 'journal-article' ['0883-9026']

csl_json_to_zotero(csl, template)
# AttributeError: 'list' object has no attribute 'strip'        <- defect 1

csl["ISSN"] = csl["ISSN"][0]                     # patch defect 1 only
item = csl_json_to_zotero(csl, template)
print(item["itemType"])                          # 'document'   <- defect 2
print(item.get("publicationTitle", "<absent>"))  # '<absent>'
print(item.get("volume"), item.get("issue"), item.get("pages"))  # None None None
```

## Defect 1 — `.strip()` on fields Crossref returns as arrays

`citation_import.py:567-568`:

```python
_set_if_in_template(template, "ISBN", (csl.get("ISBN") or "").strip())
_set_if_in_template(template, "ISSN", (csl.get("ISSN") or "").strip())
```

Crossref sends `"ISSN": ["0883-9026"]` and `"ISBN": ["9781607320395"]`. A non-empty list is not falsy, so `(["0883-9026"] or "")` evaluates to the list and `.strip()` raises. Conversion fails outright — so this breaks essentially **every journal article**, since almost all carry an ISSN, and every book or chapter with an ISBN.

Same exposure on every field the function `.strip()`s directly instead of wrapping in `str()`:

| Line | Field | What Crossref sends |
|---|---|---|
| 535 | `title` | string here; **array** from the plain `/works/{doi}` endpoint |
| 536 | `title-short` | string |
| 552 | `container-title` | string here; array from some other CSL producers |
| 557 | `collection-title` | ditto |
| 564 | `publisher` | string |
| 565 | `publisher-place` | string |
| 567 | `ISBN` | **array** |
| 568 | `ISSN` | **array** |
| 569 | `DOI` | string |
| 570 | `URL` | string |
| 571 | `language` | string |
| 572 | `abstract` | string (JATS-wrapped) |

The `str()`-wrapped neighbours (`volume`, `issue`, `page`, `edition`, `collection-number`, `number-of-pages`) don't raise, but they render a list as `"['42']"`, so they are wrong rather than fatal. The split between the two groups looks accidental.

Note that the coercion has to go at the call sites, not inside `_set_if_in_template`: `.strip()` is evaluated in the argument expression, so it raises before `_set_if_in_template` is entered. A patch there looks plausible and fixes nothing.

`bibtex_entry_to_zotero` needs no equivalent — bibtexparser hands it strings.

## Defect 2 — `CSL_TYPE_MAP` doesn't know Crossref's type vocabulary

`CSL_TYPE_MAP` (`citation_import.py:65-100`) keys on CSL-spec type names. Crossref's "CSL JSON" is not quite CSL: it keeps Crossref's own vocabulary — `journal-article` where CSL says `article-journal`, `book-chapter` where CSL says `chapter`.

Twelve of the fourteen types a literature corpus actually contains miss the table and fall through to the `"document"` default:

```
journal-article  proceedings-article  book-chapter   book-part
book-section     reference-entry      monograph      edited-book
reference-book   report-component     dissertation   posted-content
```

`book` and `report` spell the same in both vocabularies and map correctly by luck. Everything else — including every journal article and every book chapter — becomes a `document`.

**The consequence is data loss, not just a wrong label.** `_set_if_in_template` writes a field only when the template already carries it, and a `document` template has no `publicationTitle`, `volume`, `issue` or `pages`. The reproduction above shows the journal name, volume, issue and page range all vanishing from a record that had every one of them. Nothing warns.

## Why the tests miss both

`tests/test_citation_import.py` and `tests/test_add_by_csl_json.py` cover this converter well, and every CSL fixture in them is hand-written in CSL-spec form — `"type": "article-journal"`, `"ISSN": ""` as a string. Not one is a recorded provider response, so the suite has never exercised the shape the function actually meets in production.

Worth fixing alongside the defects: land at least one captured Crossref transform response as a fixture. Two stable DOIs cover both paths — `10.1016/j.jbusvent.2019.105970` (journal article, `ISSN` array, `type: journal-article`, volume/issue/pages) and `10.5876/9781607320395.c008` (book chapter, `ISBN` array, `type: book-chapter`, page range and `container-title`).

## Not a defect

`csl_json_to_zotero` preserving unmapped fields as labelled `extra` lines is unhelpful for a Crossref payload — it carries `indexed`, `license`, `content-domain`, `member`, `score`, `deposited` and a `reference` list that ran to 108 entries on one ordinary article. But that is the caller's decision, not the converter's, and it is right for hand-written CSL. Same for `id`, which Crossref sets to the DOI: the `Citation Key:` line would pin a garbage key in a field Better BibTeX reads, but for genuine CSL carrying a real citekey the current behaviour is correct. No change wanted there.

Found while adopting `csl_json_to_zotero` in [mronkko/claude-academic-research](https://github.com/mronkko/claude-academic-research)'s `import_to_zotero.py`, which currently works around both in a local adapter.

PR follows.

mronkko:
Fixes #465 — `csl_json_to_zotero` could not read Crossref's CSL JSON at all, which is where most CSL JSON in the wild comes from.

## What was wrong

**Arrays crashed the converter.** `ISSN` and `ISBN` were read as `(csl.get("ISSN") or "").strip()`. Crossref sends `["0883-9026"]`, a non-empty list is not falsy, so `.strip()` raised `AttributeError` and the whole conversion failed. That is essentially every journal article, since almost all carry an ISSN, and every book or chapter with an ISBN.

**Crossref's type names weren't in the map.** `CSL_TYPE_MAP` keyed on CSL-spec spellings, but Crossref's `/transform/application/vnd.citationstyles.csl+json` endpoint keeps Crossref's own vocabulary: `journal-article`, not `article-journal`; `book-chapter`, not `chapter`. Twelve of the fourteen types a literature corpus actually contains missed the table and fell through to the `document` default — `book` and `report` coincide between the two vocabularies and mapped correctly by luck.

That second one is data loss, not a mislabelling. `_set_if_in_template` writes a field only when the template already carries it, and a `document` template has no `publicationTitle`, `volume`, `issue` or `pages`. An article that had all four arrived with all four gone, and nothing warned.

## What changed

- **`_csl_text(value)`** — one helper, applied to every scalar CSL read in `csl_json_to_zotero`. Takes the first non-empty element of an array (the further ISSNs are the same journal's other media forms) and stringifies numbers. That also covers the `title` and `container-title` arrays the plain `/works/{doi}` endpoint returns, and integer `volume`/`issue` from other exporters. It goes at the call sites deliberately: `.strip()` is evaluated in the argument expression, so a guard inside `_set_if_in_template` would never be reached.

  It also folds in the `str()`-wrapped neighbours (`volume`, `issue`, `page`, `edition`, `collection-number`, `number-of-pages`), which didn't raise but rendered a list as `"['42']"`. The split between the two groups looked accidental; now there is one rule.

- **Twelve Crossref type aliases** added to `CSL_TYPE_MAP`, alongside the CSL-spec keys rather than replacing them, so both spellings work. Verified against `https://api.crossref.org/types` — no collisions with existing keys.

- **Two recorded Crossref responses** under `tests/fixtures/`, with a README recording their provenance, the fetch date, and the single elision (the article's 108-entry `reference` list, a third of the payload and irrelevant to the converter). `MANIFEST.in` picks them up for the sdist.

## Verification

Live Crossref payloads, before and after:

| | `10.1016/j.jbusvent.2019.105970` | `10.5876/9781607320395.c008` |
|---|---|---|
| before | `AttributeError` | `AttributeError` |
| before, arrays patched out by hand | `document`, journal/volume/issue/pages gone | `document`, book title and pages gone |
| after — `itemType` | `journalArticle` | `bookSection` |
| after — container | `Journal of Business Venturing` | `The Archaeology of Class War` |
| after — volume / issue / pages | `35` / `1` / `105970` | — / — / `161-185` |
| after — ISSN / ISBN | `0883-9026` / — | — / `9781607320395` |
| after — creators | Douglas, Shepherd, Prentice | Chicone (author); Larkin, McGuire (editors) |

29 new tests. Against the unfixed converter 18 of them fail; the 4 that pass are the `book`/`report` coincidences and the two checks that spec spellings still work. Full suite: 1781 passed. `ruff check` on the changed files is clean (the repo's 120 pre-existing findings elsewhere are unchanged).

## On the fixtures

Every CSL fixture in `tests/test_citation_import.py` and `tests/test_add_by_csl_json.py` is hand-written in spec form — `"type": "article-journal"`, `"ISSN": ""` as a string — which is why a well-covered converter was green while failing on every payload it met in production. A fixture we wrote ourselves can only pin the shape we already believed. Where a converter's input comes from a third party, at least one fixture should be a captured real response, and now two are.

The fixture tests also build their template from `zotero_mcp.schema.valid_fields` rather than the hand-written `make_template` in that file, which lists only the fields its own tests assert on — the same blind spot one level down.

## Deliberately not changed

Crossref payloads carry `indexed`, `license`, `content-domain`, `member`, `score`, `deposited` and a long `reference` list, all of which the converter preserves as labelled `extra` lines. That is noisy for Crossref and right for hand-written CSL, and it is the caller's call to forward only mappable fields. Same for `id`, which Crossref sets to the DOI: the resulting `Citation Key:` line would pin a garbage key in a field Better BibTeX reads, but for genuine CSL carrying a real citekey the behaviour is correct.

One judgment call worth flagging: `reference-entry` maps to `dictionaryEntry`. Crossref uses that type for entries in reference works generally, so `encyclopediaArticle` would be defensible too. Happy to switch it.


mronkko:
Two commits added to this branch since it was opened, both from a downstream pipeline exercising the fix against a real corpus.

**`fix: map CrossRef's container types in csl_json_to_zotero`** — same defect as Defect 2 here, found in a live library rather than through the transform endpoint. DOI `10.13165/vpa-20-19-2-04` is an ordinary journal article whose publisher registered it as a `journal-issue`. That type was not in `CSL_TYPE_MAP` either, so it took the same `document` fallthrough and lost its journal, volume, issue and page range. Adds the container types (`journal-issue`, `journal-volume`, `journal` → `journalArticle`), the remaining book/proceedings spellings, and moves `dataset` and `standard` to Zotero's native item types.

**`test: pin that a pre-flattened ISSN agrees with Crossref's array form`** — test-only. A downstream pipeline hit Defect 1 independently on 2026-08-16, months before this issue was filed, and shipped a transform that flattens `ISSN` to a plain string before calling. They raised the right concern: a caller pre-flattening ISSN would never exercise the array handling and could mask a regression in it.

Checking that turned up something worth recording. On the unfixed code, flattening ISSN alone *did* make the call succeed — and produced a `document` with `publicationTitle`, `volume`, `issue` and `pages` silently dropped, because Defect 2 typed it. **The workaround converted a loud crash into quiet data loss**, which is the more expensive failure. Injecting an `id`, the other half of their workaround, changed nothing either way — Crossref's transform endpoint emits no `id` at all and conversion never required one.

The test asserts the two spellings produce byte-identical items, that both keep `journalArticle` with its fields, and that an absent `id` is not a requirement.

Also worth noting as independent confirmation: that pipeline hit Defect 1 against real Crossref payloads at a second site, before this issue existed, and has now verified the fix against the byte-identical output of its own fetch script — `ISSN` unflattened, no `id` — importing cleanly as `journalArticle` with volume, issue, pages, journal, publisher, authors and abstract. They are removing the workaround.

The `add_by_doi` sibling of Defect 2 lives in `CROSSREF_TYPE_MAP` and is filed separately as #476, since it is a different code path reached by a different tool.

Issue #439, opened 2026-08-10
Feature Request: Text extraction for `.docx` (and other Office) attachments


Michael-Logies:
**Related to:** `zotero_get_item_fulltext`, `zotero_get_attachment_path`, attachment handling  
**Author:** Michael Logies (on behalf of Kimi Code user)  
**Date:** 2026-08-10

---

## Summary

When a Zotero item has a `.docx` (or `.doc`) file as a child attachment — very common for journal supplementary materials — the MCP server currently returns only binary/ZIP data via `zotero_get_item_fulltext`. This makes it impossible for clients to read the actual content of supplementary files without external conversion tools. The server should extract and return the textual content of such files, similar to how it handles PDFs.

## Current behavior

1. `zotero_get_item_fulltext` on a `.docx` attachment returns the raw ZIP/XML bytes of the OOXML package (e.g. `[Content_Types].xml`, `word/document.xml`, etc.), which is unusable for downstream LLM analysis.
2. `zotero_get_attachment_path` is not useful in remote/local-mode setups where the Zotero storage resides on a different machine (e.g. the MCP server runs on Windows, the client on a Linux VM). It returns "No attachments found".
3. As a result, `.docx` supplementary files are effectively opaque to MCP clients.

## Real-world impact

Many journals (Springer Nature, Elsevier, Wiley, etc.) provide supplementary materials as `.docx` files, for example:

- lists of included studies,
- detailed methods,
- raw LLM outputs,
- item-level rating tables,
- checklists and forms.

In evidence-synthesis and meta-research, these files often contain data that is as important as the main article. Users currently have to download and convert them manually or via publisher URLs, bypassing Zotero.

## Proposed solution

Add native text extraction for `.docx` attachments in `zotero_get_item_fulltext` (or as a dedicated new tool, e.g. `zotero_get_attachment_text`).

### Minimum viable behavior

For `.docx` (Office Open XML):
1. Read `word/document.xml` from the OOXML package.
2. Strip XML tags and convert the remaining text into a readable format.
3. Optionally preserve simple structure (paragraphs, tables) as Markdown or HTML.
4. Return the extracted text to the client.

### Suggested output formats

Allow an optional `format` parameter:

- `text` — plain text, paragraphs separated by blank lines (default).
- `markdown` — basic Markdown with headings, lists, and tables preserved (ideal for LLM consumption).
- `html` — simplified HTML, similar to the existing simplified HTML used for notes.

Example tool call:

```json
{
[...]

Issue #438, opened 2026-08-10
Feature Request: Expose retracted/publication-retraction status via MCP tools


Michael-Logies:
**Author / Source:** Drafted by Kimi (AI assistant) during a user session; posted on behalf of the user.

---

## Summary

Zotero desktop has a built-in retraction-notification feature: it checks items against the Retraction Watch database and flags retracted publications in the item list and item pane. Currently, this status is **not exposed through the `zotero-mcp-server` tools**. For AI-assisted literature workflows (e.g. automated evidence summaries, systematic notes, bibliography verification) it would be very valuable to surface retraction status through the MCP layer so that assistants can warn users or exclude retracted papers without requiring manual visual inspection in Zotero.

---

## Background

Per the [Zotero blog post "Retracted Item Notifications"](https://www.zotero.org/blog/retracted-item-notifications/):

- Zotero flags retracted items in the items list.
- Selecting a retracted item shows a warning at the top of the item pane with details and links.
- The check uses privacy-preserving partial identifiers so Zotero servers do not learn which papers a user owns.
- The feature works for items with a DOI or PMID.

This is a **client-side Zotero feature**. The retraction data is already present in the local Zotero database after Zotero has performed its check, so an MCP server running in local mode should be able to read it without additional network calls or privacy concerns.

---

## Problem

None of the current MCP tools expose retraction status:

- `zotero_get_item_metadata` does not return a `retracted` field.
- `zotero_search_items` does not allow filtering by retraction status.
- There is no tool to list retracted items in the library.

As a result, an MCP client cannot automatically:

- Exclude retracted articles from evidence syntheses.
- Flag retracted items when compiling literature lists.
- Verify that a cited paper has not been retracted.

---

## Proposed solutions

Any of the following (or a combination) would solve the use case:

### Option 1: Add `retracted` to item metadata

Extend `zotero_get_item_metadata` (and any item-returning search tools) to include a boolean `retracted` field:

```json
{
  "key": "ABC12345",
  "itemType": "journalArticle",
  "title": "...",
  "retracted": true
}
```

Optionally include retraction details (reason, date, URL) if Zotero stores them locally.

### Option 2: Add a dedicated tool

Introduce a new tool, e.g. `zotero_get_retracted_items`, which returns the list of item keys (and basic metadata) for all retracted items in the current library.

```json
{
  "name": "zotero_get_retracted_items",
  "arguments": {"limit": 100, "start": 0}
}
```

### Option 3: Add a search filter

[...]

PR #468, opened 2026-08-22
fix: resolve type-specific base fields when rendering an item's title

### LINKED ISSUE #452: Bug report: Zotero MCP connector drops type-specific base fields when formatting items
## Summary
 
The connector's item formatter reads `data["title"]` and `data["date"]` directly from the Zotero API record. Zotero does not store those fields under those names for every item type: several types use a **type-specific field mapped to a base field**. For those types the formatter finds nothing and emits the literal strings `Untitled` and `No date`.
 
The data is present and correct in Zotero throughout. Only the connector's presentation layer loses it.
 
---

## Affected version

Zotero MCP v0.9.1
 
## Affected item types
 
Confirmed against the Zotero schema (`https://api.zotero.org/schema`) and verified empirically in a 1,585-item library.
 
### Title lost
 
| Item type | Actual field | baseField | Verified |
|---|---|---|---|
| `case` | `caseName` | `title` | yes — 16/16 items render as `Untitled` |
| `statute` | `nameOfAct` | `title` | yes — 26/26 items render as `Untitled` |
| `email` | `subject` | `title` | yes — 6/6 items render as `Untitled` |
 
### Date lost
 
| Item type | Actual field | baseField | Verified |
|---|---|---|---|
| `case` | `dateDecided` | `date` | yes — renders as `No date` |
| `statute` | `dateEnacted` | `date` | yes — renders as `No date` |
| `patent` | `issueDate`, `priorityDate` | `date` | yes — 4/4 patents render as `No date` (titles are fine; patents use plain `title`) |
 
### Latent — same root cause, not yet verified
 
The formatter I exercised doesn't display these, but any code path that does will fail identically:
 
- `baseField: publicationTitle` — `bookSection`→`bookTitle`, `conferencePaper`→`proceedingsTitle`, `webpage`→`websiteTitle`, `forumPost`→`forumTitle`, `encyclopediaArticle`→`encyclopediaTitle`, `dictionaryEntry`→`dictionaryTitle`, `presentation`→`sessionTitle`, `radioBroadcast`/`tvBroadcast`→`programTitle`
- `baseField: number` — `case`→`docketNumber`, `statute`→`publicLawNumber`, `bill`→`billNumber`, `patent`→`patentNumber`, `podcast`/`radioBroadcast`/`tvBroadcast`→`episodeNumber`, `dataset`→`identifier`, `preprint`→`archiveID`
 
---
 
## Affected tools
 
Every tool sharing the formatter is affected, which indicates a single shared code path rather than per-tool bugs:
 
- `zotero_get_collection_items` (all detail levels: `keys_only`, `summary`, `full`)
- `zotero_search_items`
- `zotero_advanced_search`
- `zotero_get_item_metadata` with `format='markdown'`
- `zotero_get_item_metadata` with `format='bibtex'` — **worst case, see below**
**Not affected:**
 
- `zotero_get_item_metadata` with `format='json'` — returns the raw record correctly. This is the workaround.
- `zotero_export_bibliography` — delegates to Zotero's own CSL engine and renders everything correctly.
---
 
## Reproduction
 
Any `case`, `statute` or `email` item will do. Using `QI92BRVI` (a UK Upper Tribunal decision):
 
**1. Listing / search — title and date lost:**
 
```
zotero_search_items(query='DSG Retail')
```
```
## 2. Untitled
**Type:** case
**Item Key:** QI92BRVI
**Date:** No date
**Authors:** No authors listed
```
 
Note the search *matched* on the case name — Zotero's search layer resolves base fields correctly. Only the renderer doesn't. So the item is findable but unidentifiable, which is a particularly confusing failure mode.
 
**2. Single-item markdown view — nearly everything lost:**
 
```
zotero_get_item_metadata(item_key='QI92BRVI', format='markdown')
```
```
# Untitled
**Type:** case
**Item Key:** QI92BRVI
**Collections:** 6BY2N2VG
**Notes/Attachments:** 2
```
 
Beyond title and date, this also silently drops `court`, `docketNumber` and `dateDecided` — the entire type-specific field set.
 
**3. BibTeX export — empty entry:**
 
```
zotero_get_item_metadata(item_key='QI92BRVI', format='bibtex')
```
```bibtex
@misc{nodate_QI92BRVI
}
```
 
**4. The same item via `format='json'` — all data present:**
 
```json
{
  "caseName": "DSG Retail Limited v The Information Commissioner",
  "court": "Upper Tribunal",
  "dateDecided": "8 October 2024",
  "docketNumber": "[2024] UKUT 287 (AAC)",
  "itemType": "case"
}
```
with `"meta": {"parsedDate": "2024-10-08"}`.
 
**5. The same item via `zotero_export_bibliography` — correct:**
 
```
zotero_export_bibliography(item_keys=['QI92BRVI'], style='apa')
→ DSG Retail Limited v The Information Commissioner, [2024] UKUT 287 (AAC)
  (Upper Tribunal October 8, 2024).
 
zotero_export_bibliography(item_keys=['QI92BRVI'], export_format='bibtex')
→ @misc{noauthor_dsg_2024,
    title = {{DSG} {Retail} {Limited} v {The} {Information} {Commissioner}},
    month = oct, year = {2024}, ... }
```
 
The contrast between reproduction 3 and reproduction 5 is the clearest statement of the bug: the same item, the same requested format, one path hand-rolled and empty, one path delegated to Zotero and correct.
 
### 1. Wasted output tokens (small, constant)
 
Each affected item in a listing spends roughly 40 tokens saying nothing:
 
```
## 14. Untitled
**Type:** case
**Item Key:** QI92BRVI
**Date:** No date
**Authors:** No authors listed
```
 
Four of those five lines carry no information. This is minor in isolation — but it is *negative*-value output, because it costs tokens **and** actively misleads the consumer.
 
### 2. Recovery round-trips (moderate, scales linearly)
 
The only way to recover an affected item's identity is a second call with `format='json'`, which returns the full raw record including library URLs, link objects and attachment metadata. Measured in this session: **≈450 tokens per item**, against ~15 tokens if the formatter had simply printed the name it already had in hand.
 
- 6 items recovered in this session ≈ **2,700 tokens**
- Recovering all 52 affected items in this library ≈ **23,000 tokens**
- Marginal cost of the fix: **zero** — the formatter is already emitting a string for that field; it just emits the wrong one.

### 3. Rework from acting on wrong information (large, unbounded)
 
In this session the connector was used to file a 123-item inbox. Claude reported that five items had "no usable metadata". The error only surfaced because I noticed a case I knew by name was still sitting in the inbox and challenged it. Recovering cost a diagnostic round, a correction, and re-filing: **≈4,000–5,000 tokens**, on top of the human interruption.

AmirF194:
## Root cause

`format_item_metadata` (`src/zotero_mcp/client.py`) reads the title with a
raw `data.get("title", "Untitled")`. Several Zotero item types store the
title under a different, type-specific key: a statute's is `nameOfAct`, a
case's is `caseName`, an email's is `subject` (`zotero_mcp/data/zotero_basefields.json`).
For those types `data` has no `title` key at all, so every statute, case
or email item renders as `# Untitled` regardless of its real title.

The write path already solves this for `zotero_update_item` by routing
generic field names through the type's base-field schema (94e3c1f, #402).
`format_item_metadata` predates that fix and was never updated.

## Fix

Resolve the type's actual title field with the same `schema.resolve_field`
used by the write path, then read the title from that key.

## Scope

Issue #452 also names about 15 other raw `data.get("title", ...)` sites
across `client.py`'s attachment helpers, `utils.py`, `retrieval.py`,
`annotations.py` and `discovery.py`. This PR covers only
`format_item_metadata`, the formatter behind `zotero_get_item_metadata`
and `zotero_get_item_fulltext`, since it is the highest-traffic path and
keeps this diff to one reviewable concern. The remaining sites are the
same one-line pattern and are a natural follow-up; happy to send that as
a second PR if useful. Using `Refs #452` rather than `Fixes #452` since
this does not close the whole issue.

## Verification

- New test (`tests/test_format_item_metadata_base_field.py`): statute,
  case and email items fail on `main` (title renders as `Untitled`) and
  pass on this branch; an ordinary type (title is its own base field) and
  a missing-title item are also covered and unaffected either way.
- Full suite, matching CI's own install and invocation
  (`pytest --ignore=tests/test_lifespan.py`): 1751 passed, 3 skipped, no
  new failures, run on Python 3.10 and 3.12 (the CI matrix's extremes).
- Not checked: the other read-path call sites named in #452, and a live
  round trip against the real Zotero API (the fix and its test only
  exercise the formatter against constructed item dicts).

### DIFF
--- a/src/zotero_mcp/client.py
+++ b/src/zotero_mcp/client.py
@@ -15,6 +15,7 @@ import httpx
 from dotenv import load_dotenv
 from pyzotero import zotero
 
+from zotero_mcp import schema
 from zotero_mcp.extract import (
     categorize_attachment,
     extract_file,
@@ -299,9 +300,14 @@ def format_item_metadata(item: dict[str, Any], include_abstract: bool = True) ->
     data = item.get("data", {})
     item_type = data.get("itemType", "unknown")
 
+    # Some item types keep the title under a type-specific key (a statute's
+    # is "nameOfAct"); resolve_field maps "title" to whichever key this
+    # item type actually uses.
+    title_field = schema.resolve_field(item_type, "title")
+
     # Basic information
     lines = [
-        f"# {data.get('title', 'Untitled')}",
+        f"# {data.get(title_field, 'Untitled')}",
         f"**Type:** {item_type}",
         f"**Item Key:** {data.get('key')}",
     ]
--- a/tests/test_format_item_metadata_base_field.py
+++ b/tests/test_format_item_metadata_base_field.py
@@ -1,0 +1,88 @@
+"""Tests for base-field title resolution in format_item_metadata.
+
+Several Zotero item types store the title under a type-specific key instead
+of "title" itself: a statute's is "nameOfAct", a case's is "caseName", an
+email's is "subject" (see zotero_mcp.schema). The write path
+(zotero_update_item) already routes through schema.resolve_field for this
+(commit 94e3c1f, #402); format_item_metadata did a raw data.get("title", ...)
+lookup, so every one of these item types rendered as "# Untitled" even
+though the real title was present under its type-specific key.
+"""
+
+import importlib.util
+import pathlib
+import sys
+from unittest.mock import MagicMock
+
+_SRC = pathlib.Path(__file__).parent.parent / "src" / "zotero_mcp"
+
+# Stub out heavy optional dependencies so client.py can be imported in
+# isolation; schema is loaded for real below, not stubbed.
+for _mod_name in (
+    "pyzotero",
+    "pyzotero.zotero",
+    "dotenv",
+    "fastmcp",
+    "mcp",
+    "mcp.server",
+    "zotero_mcp",
+    "zotero_mcp.utils",
+    "zotero_mcp._app",
+):
+    if _mod_name not in sys.modules:
+        sys.modules[_mod_name] = MagicMock()
+
+if "zotero_mcp.schema" not in sys.modules:
+    _schema_spec = importlib.util.spec_from_file_location("zotero_mcp.schema", _SRC / "schema.py")
+    _schema_mod = importlib.util.module_from_spec(_schema_spec)
+    sys.modules["zotero_mcp.schema"] = _schema_mod
+    _schema_spec.loader.exec_module(_schema_mod)
+
+_client_spec = importlib.util.spec_from_file_location("zotero_mcp.client", _SRC / "client.py")
+_client_mod = importlib.util.module_from_spec(_client_spec)
+sys.modules["zotero_mcp.client"] = _client_mod
+_client_spec.loader.exec_module(_client_mod)
+format_item_metadata = _client_mod.format_item_metadata
+
+
+def _make_item(item_type, title_field, title):
+    data = {
+        "key": "TESTKEY1",
+        "itemType": item_type,
+        title_field: title,
+        "creators": [{"creatorType": "author", "lastName": "Smith", "firstName": "J."}],
+        "date": "2024",
+    }
+    return {"data": data}
+
+
+class TestBaseFieldTitle:
+    def test_statute_title_stored_under_name_of_act(self):
+        item = _make_item("statute", "nameOfAct", "The Clean Air Act")
+        output = format_item_metadata(item, include_abstract=False)
+        assert "# The Clean Air Act" in output
+        assert "# Untitled" not in output
+
+    def test_case_title_stored_under_case_name(self):
+        item = _make_item("case", "caseName", "Marbury v. Madison")
+        output = format_item_metadata(item, include_abstract=False)
+        assert "# Marbury v. Madison" in output
+        assert "# Untitled" not in output
+
+    def test_email_title_stored_under_subject(self):
+        item = _make_item("email", "subject", "Re: manuscript review")
+        output = format_item_metadata(item, include_abstract=False)
+        assert "# Re: manuscript review" in output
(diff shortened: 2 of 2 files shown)


54yyyu:
Merged into `main` (commit 77fd28d). Thanks. Following up separately on `utils.format_item_result`, which resolves the title the same way and so still shows `Untitled` for those types in search results — that's the path #452 originally reported.

AmirF194:
Thanks for the merge. Good catch on utils.format_item_result, it resolves the title the same way and belongs in that follow-up along with the other sites #452 lists. Happy to pick it up if useful.

54yyyu:
You were right, and thanks for pushing on it — I'd closed #452 on the strength of the title fix alone.

Reproduced on `main`: the title now resolves, but `**Date:** No date` and `@misc{nodate_QI92BRVI}` were both still there. The date half (`dateDecided`, `dateEnacted`, `issueDate`) was untouched, and so were the two knock-on effects in the report — the dropped `court`/`docketNumber` field set, and the empty BibTeX entry.

#482 has all of it. I went ahead rather than waiting, since I'd already reopened the issue and had the repro in hand — no reflection on your offer, which I appreciated. For the metadata formatter I took the general route: it now asks the schema what fields the item type has and renders the populated ones, instead of adding a branch per type. That also covers the `publicationTitle` and `number` cases the issue lists as latent.

If you want another one, #438 (expose retracted/publication-retraction status) and #439 (text extraction for `.docx` attachments) are both well-specified and unclaimed.

PR #470, opened 2026-08-22
fix: keep brackets that belong to the DOI when stripping prose punctuation (#469)

### LINKED ISSUE #469: normalize_doi truncates DOIs that legitimately end in ')' (TAO's parenthesised suffixes)
### What happens

`normalize_doi()` strips a trailing `)` unconditionally, so a DOI whose suffix legitimately
ends in a parenthesis is truncated before the CrossRef lookup. `zotero_add_item` then reports
the DOI as not registered, which reads like a CrossRef problem rather than a client-side one.

Real example — *Terrestrial, Atmospheric and Oceanic Sciences* uses a parenthesised suffix:

```
10.3319/TAO.2009.05.25.02(IWNOP)
```

```
zotero_add_item(source='10.3319/TAO.2009.05.25.02(IWNOP)', source_type='doi')
→ DOI not found on CrossRef: 10.3319/TAO.2009.05.25.02(IWNOP      # note the missing ')'
```

The DOI is registered and resolves fine:

```bash
curl -sS -H 'Accept: application/vnd.citationstyles.csl+json' \
  "https://api.crossref.org/works/10.3319%2FTAO.2009.05.25.02%28IWNOP%29/transform/application/vnd.citationstyles.csl+json"
# → 200, "Numerical Simulation of Barotropic Tides around Taiwan", Terr. Atmos. Ocean. Sci. 21(1)
```

### Root cause

`src/zotero_mcp/identifiers.py` (current `main`, v0.9.1):

```python
_TRAILING_PUNCT = ".,);]"
...
s = s.rstrip(_TRAILING_PUNCT)
```

The intent is clearly to clean a DOI copied out of prose (`... see 10.1234/foo.`), which is
worth keeping. The problem is only that `)` is stripped without checking whether it closes
anything. Parenthesised suffixes are uncommon but legal, and TAO uses them for a whole series
of articles.

### Suggested fix

Strip a trailing `)` only when it is unbalanced:

```python
_TRAILING_PUNCT = ".,;]"

s = s.rstrip(_TRAILING_PUNCT)
while s.endswith(")") and s.count("(") < s.count(")"):
    s = s[:-1]
    s = s.rstrip(_TRAILING_PUNCT)
```

Behaviour on the cases that matter:

| input | current | with fix |
|---|---|---|
| `10.3319/TAO.2009.05.25.02(IWNOP)` | `...02(IWNOP` ❌ | `...02(IWNOP)` ✅ |
| `10.1234/foo.bar).` (from prose) | `10.1234/foo.bar` ✅ | `10.1234/foo.bar` ✅ |
| `10.1029/2004JC002484` | unchanged ✅ | unchanged ✅ |

### Why it is worth fixing

The failure is silent in the sense that matters: the message points at CrossRef, so the natural
next step is to work around it with `source_type='csl_json'` — which lands on #465 and files the
item as `document` with journal, volume, issue and pages empty. Two quiet failures in a row for
what is one stripped character.

Reproduced through the MCP server on the released **v0.9.1** (`src/zotero_mcp/tools/_helpers.py`, `s = s.rstrip(".,);]")`), and the same strip is present in `main` after the refactor into `src/zotero_mcp/identifiers.py` (`_TRAILING_PUNCT = ".,);]"`).

_Correction to an earlier draft of this report: the version was v0.9.1, not 0.6.x — I first read a stale wheel out of the uv cache._

skhyun-ocean:
Closes #469.

`normalize_doi` finishes with `rstrip(".,);]")`. That is the right move for a DOI copied out of a sentence — `"(see 10.1038/nphys1170)"` — and the wrong one for a DOI that opens a bracket itself.

TAO (*Terrestrial, Atmospheric and Oceanic Sciences*) puts a parenthesised topical suffix on its DOIs:

```python
>>> normalize_doi("10.3319/TAO.2009.05.25.02(IWNOP)")
'10.3319/TAO.2009.05.25.02(IWNOP'      # before
'10.3319/TAO.2009.05.25.02(IWNOP)'     # after
```

The closer is part of the DOI — CrossRef resolves the full string and 404s without it — so every DOI-driven path, `zotero_add_item(source_type="doi")` included, answered *"DOI not found on CrossRef"* for a record CrossRef holds:

```bash
curl -sS -H 'Accept: application/vnd.citationstyles.csl+json' \
  "https://api.crossref.org/works/10.3319%2FTAO.2009.05.25.02%28IWNOP%29/transform/application/vnd.citationstyles.csl+json"
```

## The change

`_strip_trailing_punct()` walks the tail instead of calling `rstrip`, and stops at a closer whose opener is present in the string. A stray bracket still goes; a balanced pair stays.

```python
while s and s[-1] in _TRAILING_PUNCT:
    opener = _CLOSERS.get(s[-1])
    if opener is not None and s.count(opener) >= s.count(s[-1]):
        break
    s = s[:-1]
```

## Tests

Five new cases in `TestNormalizeDoi.test_balanced_brackets_are_part_of_the_doi` — the bare DOI, the `doi:` form, the `doi.org` URL, the DOI followed by a sentence period, and `10.1234/a[b]` for square brackets. They fail on `main` and pass here. The existing `test_trailing_punctuation_from_prose` cases (`10.1038/nphys1170)` and `10.1038/nphys1170];`) are unchanged and still pass, which is what pins the stray-bracket behaviour.

`pytest tests/` — 1,396 passed on this branch. The 5 failures and 15 collection errors in my environment are `pymupdf` / `chromadb` not installed and reproduce identically on `main`.

## Why it lasted

The failure reads as a missing record rather than a client bug, so the natural response is to go look for the paper somewhere else. Bracketed suffixes are rare — TAO is the one I hit — but they resolve fine everywhere except here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - **`zotero_set_item_parent` sets or clears an item's parent.** It can parent standalone notes and attachments, move them between parents, or make them top-level again.
 
+### Fixed
+- **`normalize_doi` no longer truncates DOIs that legitimately end in a bracket (#469).** The trailing-punctuation strip was `rstrip(".,);]")`, which cannot tell prose punctuation from a bracket the DOI itself opened. TAO (Terrestrial, Atmospheric and Oceanic Sciences) puts a parenthesised topical suffix on its DOIs, so `10.3319/TAO.2009.05.25.02(IWNOP)` was normalised to `…02(IWNOP` and every path that resolves a DOI — `zotero_add_item(source_type="doi")` among them — reported "DOI not found on CrossRef" for a record CrossRef holds and resolves. The strip now walks the end of the string and stops at a closer whose opener is present in the DOI, so `10.1038/nphys1170)` copied out of "(see 10.1038/nphys1170)" still loses its stray bracket while a balanced pair survives. Bracketed suffixes are rare enough that the failure reads as a missing record rather than a client bug, which is how it went unnoticed.
+
 ## [0.9.1] - 2026-08-05
 
 ### Changed
--- a/src/zotero_mcp/identifiers.py
+++ b/src/zotero_mcp/identifiers.py
@@ -33,14 +33,34 @@ _DOI_IN_URL_RE = re.compile(
 #: Punctuation that trails a DOI copied out of prose or a reference list.
 _TRAILING_PUNCT = ".,);]"
 
+#: Closing brackets and the opener each one belongs to. A closer is prose
+#: punctuation only when the DOI holds no matching opener.
+_CLOSERS = {")": "(", "]": "["}
+
+
+def _strip_trailing_punct(s):
+    """Strip prose punctuation from the end of a DOI, keeping brackets that
+    the DOI itself opened.
+
+    ``rstrip(_TRAILING_PUNCT)`` cannot tell "(see 10.1234/foo)" from TAO's
+    ``10.3319/TAO.2009.05.25.02(IWNOP)``, where the parentheses are part of
+    the suffix and CrossRef 404s without the closer.
+    """
+    while s and s[-1] in _TRAILING_PUNCT:
+        opener = _CLOSERS.get(s[-1])
+        if opener is not None and s.count(opener) >= s.count(s[-1]):
+            break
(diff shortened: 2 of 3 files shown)


54yyyu:
Merged into `main` as part of the 0.10.0 batch (commit ce93c0f). Thanks — the unbalanced-closer check is exactly right, and the TAO case is now covered by tests.

Issue #471, opened 2026-08-22
Implement local write support


raffaelemancuso:
The Zotero API now supports [local write](https://www.zotero.org/support/dev/web_api/v3/local_api#write_requests).

It should be implemented in this MCP server and the remote web server removed since it has now lost is purpose

mronkko:
This belongs to upstream and is handled there https://github.com/urschrei/pyzotero/issues/344

54yyyu:
Released in 0.12.0 (`uv tool install --upgrade "zotero-mcp-server[all]"` or `pip install -U zotero-mcp-server`). Thanks for the report.

Issue #463, opened 2026-08-20
add_from_file reports success when file is uploaded to cloud but not present locally (local-only setups)


alexey-ezhikov:
### Summary

With `ZOTERO_LOCAL=true` for reads and a Web API key for writes, `zotero_add_from_file` uploads the file to Zotero's cloud storage and reports `File attached`. On installations where Zotero's file syncing is disabled (`extensions.zotero.sync.storage.enabled = false`), the file never lands in `~/Zotero/storage/<KEY>/`. The attachment row is valid, so nothing looks wrong, but the attachment is unreadable locally and every downstream tool that resolves a local path fails.

### Environment

- zotero-mcp-server 0.6.4, pyzotero 1.14.0
- Zotero 7 desktop, macOS
- `extensions.zotero.sync.storage.enabled = false` (metadata sync on, file sync off)

### Minimal repro

1. Ensure `extensions.zotero.sync.storage.enabled = false` in the Zotero desktop profile.
2. Call `zotero_add_from_file` with any local PDF and a parent item key.
3. Tool returns success (`File attached`).
4. Inspect the attachment:
   ```sql
   SELECT ia.linkMode, ia.path, ia.syncState
   FROM itemAttachments ia JOIN items i ON i.itemID = ia.itemID
   WHERE i.key = '<ATTACHMENT_KEY>';
   -- 0 | storage:<filename>.pdf | 1      <- syncState 1 = TO_DOWNLOAD
   ```
5. `ls ~/Zotero/storage/<ATTACHMENT_KEY>/` → **no such directory**.
6. `zotero_get_attachment_path` → `(missing on disk)`.

Observed on 4 of 4 attachments created this way; 0 of 2177 attachments created by the desktop client show `syncState = 1`.

### Expected

The tool should distinguish "uploaded to cloud" from "available locally". Either:

- after upload, check whether `~/Zotero/storage/<KEY>/<filename>` exists and, if not, return a warning alongside success — e.g. `File uploaded to Zotero cloud storage; not present in local storage (file syncing appears to be disabled)`; or
- optionally place a copy into local storage when the data directory is writable and `ZOTERO_LOCAL` is set.

### Why it matters

The success message is the only signal most callers have. A silently cloud-only attachment passes every metadata check and fails only at the point of reading the PDF, which can be much later.

### Not a correctness bug

The upload itself works; this is about the reporting contract on local-only setups. Filing as an enhancement.

54yyyu:
[...]

Issue #399, opened 2026-07-24
PDF attach silently fails on free-tier cloud quota (HTTP 413) for local-storage users without WebDAV — propose connector-endpoint fallback


how2star:
## Summary

For users on **Zotero local storage + the free 300 MB cloud quota, without WebDAV configured**, every `zotero_add_by_*` PDF attach path eventually fails once the quota is near full — the metadata item is created, the tool **reports "PDF attached"**, but the file never lands and the desktop client shows "Attached file could not be found". This is the same silent-failure class as #278 / #306, but the existing fixes don't cover this storage profile because they all assume the user has (or should set up) WebDAV.

I'd like to propose a **connector-endpoint fallback** that routes the PDF through the running Zotero desktop into local storage, bypassing Web API file upload entirely — no WebDAV required, no cloud quota consumed. I've validated it end-to-end and it reproduces the browser-Connector experience. A minimal repro script is included below.

## Environment

- zotero-mcp: 0.6.2 (uv tool install)
- Mode: hybrid (`ZOTERO_LOCAL=true` + Web API key) — local reads, Web API writes
- Zotero: 7.x, desktop running with local API enabled (`Allow other applications…`)
- Local storage: `D:\paper\storage`, currently **255.2 MB / 79 attachments** (~85% of the 300 MB free tier)
- No `ZOTERO_WEBDAV_*` configured (deliberate — user wants local storage only)

## The problem, concretely

Reproduced with three arXiv papers (2405.16444, 2312.07104, 2407.00079). Across the board:

| Tool | Metadata | PDF | What it reported | Reality |
|------|----------|-----|------------------|---------|
| `zotero_add_by_url` (arXiv) | ✅ | ❌ | "PDF attached" | no child attachment, no file in storage |
| `zotero_add_by_doi` | ✅ (or "Already in library") | ❌ | "Already in library" | PDF never fetched |
| `zotero_add_from_file` (local PDF) | ❌ (only filename as title) | ❌ | "File attached: …" | created a *document* item, no real attachment; also created duplicates |

The item is created via Web API, but `attachment_both()` uploads the bytes to Zotero Storage, which **returns HTTP 413 once the 300 MB quota is full**. The exception is swallowed by the non-fatal `except` in `_add_by_arxiv` / `_attach_pdf_with_fallback`, and the tool returns an optimistic "PDF attached" string. So the failure is **silent and self-reported as success** — the exact pattern called out in #278 and #306.

[...]

PR #472, opened 2026-08-23
feat(local-api): support Zotero 10 local write endpoints (#471, #277)

raffaelemancuso:
Zotero's local API was read-only for most of its life, so every mutation in this server went through the Zotero web API — meaning local-mode users needed cloud credentials to write at all, and writes that did land took a sync round-trip to come back. Zotero 10 added local write endpoints, and pyzotero 1.14.0 wraps them.

On Zotero 10 a local-only user can now create items, edit metadata, manage collections and tags, write notes and annotations, trash things, and attach files with no zotero.org account. That also sidesteps the Zotero Storage quota behind the HTTP 413 attachment failures (#399) and the cloud-only attachments of #463. Users on older Zotero are unaffected: the existing hybrid path stays as the automatic fallback, chosen by a runtime capability probe rather than a version guess.

Getting the key
  Writes need a local API key, unrelated to a zotero.org key and impossible to
  create in advance — Zotero grants it through a dialog. `zotero-mcp
  authorize-local` drives that, and the zotero_authorize_local_writes and
  zotero_write_capabilities tools let an MCP client do the same without a
  shell. A remembered key is stored in its own `local_api` section of
  ~/.config/zotero-mcp/config.json, not in `client_env`, which `zotero-mcp
  setup` rebuilds from scratch and would silently drop.

  A single-use key (plain "Allow") is deliberately never written to disk: it is
  consumed by the first write, and a dead key on disk would outrank working web
  credentials on every later run. A key Zotero rejects is dropped for the same
  reason, so the next write can fall back instead of failing forever.

Three shims, because the local API has no /items/new
  item_template() raises there, and attachment_simple/attachment_both fail
  transitively through it. Templates are synthesized from the field endpoints,
  with the attachment and note shapes reproduced verbatim from
  api.zotero.org/items/new; attachments route to upload_attachments; and the
  four trash PATCHes now go through pyzotero's write dispatcher, which is what
  attaches the headers a local write requires.

  Those PATCHes also needed an explicit Content-Type: httpx infers none for a
  raw body, and the local API answers "400 Empty request body" without it. The
  web API tolerates its absence, which is why it went unnoticed while writes
  were cloud-only.

One write path
  The three divergent gateways in write.py, annotations.py and the two inline
  annotation checks collapse into one resolver returning (read, write, mode).
  When nothing can write, the message names both remedies and leads with
  whichever one applies to the running Zotero.

Also fixed
  - The injected local httpx client had no timeout and inherited httpx's 5s default. Reads were unaffected (pyzotero passes its own on read paths) but every write shared that budget, and it made the authorization dialog impossible to answer in time.
  - The zotero.org API key is no longer sent to localhost:23119. Restoring pyzotero's default headers would have turned it into an Authorization bearer aimed at a server that has no use for it.
  - A config file that exists but cannot be parsed is no longer silently replaced, and writes are atomic — the same file holds semantic_search and client_env.
  - test_description_tokens.py and test_add_from_file.py could not run on Windows at all (locale codec, and a monkeypatch assuming str paths).
  - The bibtex place-field tests depended on whether Zotero happened to be running, since generate_bibtex prefers Better BibTeX.

Verified against a live Zotero 10.0: authorization, item/collection/tag/note creation and edits, a file attachment landing in Zotero's own storage with a matching MD5, trashing, and a rejected key producing an actionable error and self-healing. 1,020 tests pass.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,7 +1,12 @@
 # Zotero MCP 环境变量配置
 # 复制此文件为 .env 并填写你的配置
 
-# 必需配置
+# 连接方式: 本地 Zotero API (需要 Zotero 正在运行)
+# 设为 true 后, 读取无需任何凭据
+# ZOTERO_LOCAL=true
+
+# Web API 凭据
+# 本地模式下也可以设置, 用于混合模式 (本地读取 + Web API 写入)
 ZOTERO_API_KEY=your_api_key_here
 ZOTERO_LIBRARY_ID=your_library_id_here
 
@@ -9,6 +14,14 @@ ZOTERO_LIBRARY_ID=your_library_id_here
 # 个人库: user
 # 群组库: group
 ZOTERO_LIBRARY_TYPE=user
+
+# 本地写入 (需要 Zotero 10 或更新版本)
+# 通常由 `zotero-mcp authorize-local` 自动写入配置文件,
+# 这些变量适用于容器和 CI 等不便使用配置文件的场景
+# ZOTERO_LOCAL_API_KEY=local_api_key_here
+# ZOTERO_LOCAL_SERVER_ID=server_id_here
+# 设为 false 可强制通过 Web API 写入 (默认 auto)
+# ZOTERO_LOCAL_WRITE=auto
 
 # OpenAI API Key (用于语义搜索功能，可选)
 # OPENAI_API_KEY=your_openai_key_here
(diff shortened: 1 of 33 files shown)


raffaelemancuso:
Closes #471

54yyyu:
Thanks for this — it's the most-requested capability in the tracker (#471, #277) and the writeup is genuinely good. The capability probe rather than a version guess is the right call, and so is never persisting a single-use key.

It needs a rebase before I can take it. Its base is `ad85ec4`, and `main` has since taken #442, #443-adjacent write-path work, #477, #479, #480, #466, #449, #450, #451, #417, #464, #468, #470, #461. A local merge produces 34 conflicts across 13 files, concentrated exactly where this PR works: 10 in `tools/write.py`, 5 in `tools/annotations.py`, 3 in `tools/_helpers.py`, 4 in `client.py`.

I'm not going to hand-resolve those. You verified this against a live Zotero 10 and I can't re-verify that claim after resolving 34 conflicts on your behalf — I'd be shipping a write path that nobody has actually run.

Two things that should make the rebase easier rather than harder:
- `pyproject.toml` on `main` now requires `pyzotero>=1.13.5` (#442, for the 429 handling). Your 1.14.0 floor supersedes it, so that conflict resolves in your favour.
- #442 introduced `_update_item_with_version_retry` and a `_ZoteroClient`-style factory pattern in `_helpers.py`/`client.py`. Your single `(read, write, mode)` resolver overlaps with it conceptually — worth reading that diff before you start, since the two want to be one thing rather than two.

Ping me when it's rebased and I'll review it properly.

54yyyu:
Saw you closed this. To be clear, that comment was a rebase request, not a rejection. The work is good and local write support is the most-requested thing in the tracker (#471, #277 are both still open).

If you'd rather not redo the rebase yourself, say so and I'll take it on. In that case I'd want to merge it behind the existing capability probe and ask you to confirm it still works against your Zotero 10 before I tag a release with it, since I have no way to test that path here.

Either way the branch is worth keeping. Happy to reopen whenever you want to pick it back up.

raffaelemancuso:
I'm not understanding

image

mronkko:
I did not study the PR in detail, but is this not somethign that should be impelemted in pyzotero instead? There is an active issue https://github.com/urschrei/pyzotero/issues/344 and PR addressing this https://github.com/urschrei/pyzotero/pull/345

54yyyu:
@raffaelemancuso Oops sorry there was something wrong with my maintainer agent! I think it accidentally misunderstood your comment "Closes #471" as closing the PR.

54yyyu:
**Maintainer agent response**

---

@raffaelemancuso The PR is open again, and sorry again for the noise. To be unambiguous: the close was my agent's mistake, nothing about the PR was rejected, and #471 / #277 are still the most-requested thing in the tracker.

@mronkko On the upstream question, you are right about where the transport belongs, and it is already there. urschrei/pyzotero#345 is merged and shipped in 1.14.0: `Zotero.__init__` takes `server_id` and `local_api_key`, and `Zotero.authorize_local` exists. That is why this PR sets a 1.14.0 floor instead of implementing the protocol itself.

What is left is the part pyzotero should not own, and it is most of this diff: acquiring the key at all (Zotero grants it through a dialog, so it cannot be configured in advance), deciding at runtime whether the running Zotero can write, persisting a remembered key where `zotero-mcp setup` will not clobber it, and the local API having no `/items/new` to build templates from.

I have Zotero 10 locally now, which I did not when I asked for the rebase. Probing it:

```
POST /api/users/<id>/items                    -> 428 "Zotero-Server-ID not provided"
POST + Zotero-Server-ID                       -> 401 "API key required -- POST /api/local/authorize to obtain one"
GET  /api/users/<id>/items/new?itemType=book  -> 404
```

That matches the PR description exactly, missing template endpoint included.

So the offer I made is now a real one rather than a hopeful one: if you would rather not redo the rebase, I will take it, and I can verify the result against a live Zotero 10 myself instead of asking you to re-verify 34 resolved conflicts you did not resolve. If you would rather drive it, better still and I will review promptly. Tell me which and I will get out of the way.

raffaelemancuso:
We should be fine

raffaelemancuso:
@54yyyu Do we merge this PR?

Issue #476, opened 2026-08-23
CROSSREF_TYPE_MAP covers 13 of CrossRef's 30 types, so add_by_doi silently drops journal, volume, issue and pages for the other 17


mronkko:
`add_by_doi` maps CrossRef's type to a Zotero item type through
`CROSSREF_TYPE_MAP`. Seventeen of CrossRef's thirty published types are not in
the table and fall through to `document`.

That is not a labelling problem. The item is built from a Zotero item template
and only keeps fields the template has, and a `document` template has no
`publicationTitle`, `volume`, `issue` or `pages` — so those values are discarded
on the way in, with nothing reported.

This is the `add_by_doi` sibling of #465, which is the same defect in
`CSL_TYPE_MAP` on the `csl_json` path. Different file, different code path,
different tool; filing separately so they can be reviewed independently.

Reproduced against `main` @ f331646 on 2026-08-23, against live CrossRef.

## Reproduction

```python
import requests
from zotero_mcp.tools._helpers import CROSSREF_TYPE_MAP

DOI = "10.13165/vpa-20-19-2-04"   # an ordinary journal article
m = requests.get(f"https://api.crossref.org/works/{DOI}",
                 headers={"User-Agent": "repro (mailto:you@example.com)"},
                 timeout=20).json()["message"]

print(m["type"])                                    # 'journal-issue'
print(CROSSREF_TYPE_MAP.get(m["type"], "document")) # 'document'   <- the bug
```

The publisher registered this article's DOI as a `journal-issue`. It sits in the
reporting library as a `document` with no journal, no volume, no issue and no
pages.

## The gap

`tools/_helpers.py:23` lists 13 types. Missing, against
[api.crossref.org/types](https://api.crossref.org/types):

```
book-section     book-part        book-track         book-set       book-series
journal-issue    journal-volume   journal            proceedings    proceedings-series
reference-book   report-series    report-component   component      database
grant            other
```

`tools/write.py:1020` is where they land on the default:

```python
        zot_type = CROSSREF_TYPE_MAP.get(cr_type, "document")
```

Two entries are also lossy rather than absent: `dataset` and `standard` both map
to `document`, although Zotero has had native `dataset` and `standard` item types
for some time.

## Suggested fix

Cover the vocabulary, and — more importantly — stop treating a lookup miss as a
confident assertion.

- Container types (`journal-issue`, `journal-volume`, `journal`) to
[...]

PR #479, opened 2026-08-23
feat: read a publisher page's embedded citation in add_by_url

### LINKED ISSUE #474: zotero_add_by_url never reads a publisher page's embedded citation, so an article landing page becomes a webpage item whose only field is the URL
Adding a journal article's landing page produces a `webpage` item with no title,
no authors and no journal — its only populated field is the URL. The metadata is
on the page, in the same Highwire `citation_*` tags Zotero's own browser
connector reads.

Reproduced against `main` @ f331646 on 2026-08-23.

## Reproduction

```python
from unittest.mock import patch
from zotero_mcp.tools import write

URL = "https://journals.plos.org/plosone/article?id=10.1371/journal.pone.0287795"
write.add_by_url(url=URL, ctx=ctx)
```

Result: `itemType: "webpage"`, `title: "<the URL>"`, `url: "<the URL>"`. Nothing
else.

What the page serves, from `curl` on that URL:

```
citation_title, citation_author (x3), citation_journal_title,
citation_volume, citation_issue, citation_firstpage, citation_lastpage,
citation_doi, citation_pdf_url
```

## Cause

The generic-URL branch never fetches the page. `tools/write.py:1195-1199`:

```python
        ctx.info(f"Creating webpage item for: {url}")
        template = write_zot.item_template("webpage")
        template["url"] = url
        template["title"] = url
        template["accessDate"] = ""
```

DOI URLs and arXiv URLs are routed to dedicated handlers above this point;
everything else takes this path, which treats a URL as an opaque string.

## Why this is not an edge case

OJS/PKP alone runs a large share of the world's journals, and Atypon,
Silverchair, Frontiers, PLOS and DSpace repositories all publish the same tag
block. Every such save needs manual repair afterwards — and
`zotero_update_item` cannot perform the repair, because `publicationTitle`,
`volume`, `issue`, `pages` and `ISSN` are not valid fields on a `webpage`, so
they are skipped. See the companion report on `update_item` reporting that skip
as a success.

Zotero's browser connector reads exactly these tags, which is why saving a paper
from the browser gives a full record while saving the same URL through this
server gives a stub. That gap is the whole of the bug.

## Suggested fix

Fetch the page, read Highwire `citation_*` (and Dublin Core, and Open Graph as a
last resort for a title), and create a `journalArticle` or `bookSection`
accordingly. A page that declares its own `citation_doi` can be routed to
`add_by_doi`, which already handles DOI de-duplication, CrossRef enrichment and
open-access PDF attachment.

Two things worth building in, both learned the hard way while testing this:

1. **CrossRef is not always the richer source.** A publisher can register an
   article's DOI as a `journal-issue`, whose CrossRef record legitimately
   carries no title, no authors and no volume/issue/page while the landing page
   advertises all of them. Routing such a page to the DOI produces a *titleless*
   item — worse than the stub. The page's tags should fill whatever CrossRef
   leaves empty, with CrossRef winning wherever it says anything at all.
2. **Say when the enrichment could not run.** A `webpage`-only item that does not
   explain itself is how a library fills with URL-titled stubs nobody notices.
   Not every failure is the importer's fault: one host in the reporting corpus
   serves a certificate chain browsers and `curl` accept but OpenSSL will not
   verify, so `requests` cannot fetch it at all. Naming the reason is the fix
   there; weakening verification is not.

Parsing needs no new dependency — the stdlib HTML parser is enough for a head.

PR to follow.

mronkko:
Fixes #474 — an article landing page became a `webpage` item whose only populated field was the URL.

**Best reviewed together with the `fix/crossref-type-coverage` PR (#476).** They are independent commits on independent branches, but the DOI in the report is only fully repaired by both: that one gives it the right item type, this one gives it the fields CrossRef omits. Either alone is an improvement.

## What was wrong

The generic-URL branch never fetched the page. It built a `webpage` template, set `title = url`, and created it.

The metadata was on the page the whole time. Journal platforms publish the citation in the head as Highwire `citation_*` tags — exactly what Zotero's browser connector reads, hence the long-standing gap between saving a paper from the browser and saving the same URL through this server. OJS/PKP alone runs a large share of the world's journals, so every such save needed manual repair afterwards — and `zotero_update_item` could not even perform the repair, since the fields that would fix it are invalid on a `webpage`.

## What changed

- **New `zotero_mcp/html_metadata.py`** reads Highwire `citation_*`, Dublin Core, and Open Graph (last resort, title only) from a page head.
- **`add_by_url` builds a `journalArticle` or `bookSection`** with title, authors, journal, volume, issue, pages, date, DOI, ISSN and abstract.
- **A page declaring its own DOI routes to `add_by_doi`**, which already handles DOI de-duplication, CrossRef enrichment and OA PDF attachment.
- Stdlib HTML parser, **no new dependency**. Only the head is wanted, so parsing stops at `</head>` or the body opening, and the fetch reads a bounded 512 KB prefix — a hostile or enormous page must not hold a write open.
- Failures degrade to the old behaviour **and now say why**: a webpage-only item names the reason (not HTML / no citation metadata / fetch failed / certificate could not be verified) and points at the remedy.

## Second commit: CrossRef is not always the richer source

Preferring `add_by_doi` when the page declares a DOI assumes CrossRef's record is at least as complete as the page's. That assumption is wrong often enough to matter.

The URL in the report is the case in point: its publisher registered the *article's* DOI as a `journal-issue`, so CrossRef returns `title: []`, no author, no volume, no issue and no page, while the landing page advertises all of them. Routing to the DOI therefore produced a **titleless item** — worse than the bare webpage this feature replaced.

So `add_by_doi` gained an optional `supplemental`: the page's metadata fills only fields CrossRef left empty, and a value CrossRef supplied always wins. `add_by_doi` is internal (`zotero_add_item` is the public facade), so no tool schema changed.

**How it was caught:** every mocked test passed, before and after. It surfaced only on running the reporter's own re-test against live CrossRef.

## Third commit: the DOI route needed the same treatment

That fill-in only ran when a *page* supplied the metadata. A caller passing a bare DOI has no page to offer — and it is precisely the DOIs where CrossRef says least that needed it. `zotero_add_item(source_type="doi")` on the reported DOI produced a titleless, creatorless item while the url route on the same article got everything.

`add_by_doi` now resolves the DOI to its own landing page when CrossRef's answer cannot stand alone: an empty title, or a container type (`journal-issue` / `journal-volume` / `journal`). Gated on that condition, so a complete CrossRef record never pays for a page load, and a page already supplied by the url route is not re-fetched. Both pinned by tests.

**How it was caught:** the reporting agent re-ran the verification through MCP using the entry point I had told them to use. My own check had passed `supplemental` in by hand — it exercised the url route while claiming to verify the DOI one.

## Testing

22 mocked integration tests in `tests/test_add_by_url_embedded_metadata.py`, 21 parser unit tests in `tests/test_html_metadata.py`. Full suite green.

Verified against live pages from PLOS, Frontiers, First Monday and a DSpace repository.

## One caveat worth stating

The host in the original report, `ojs.mruni.eu`, serves a certificate chain that browsers and `curl` accept but OpenSSL will not verify (`Verify return code: 21`), so `requests` cannot fetch it by any route. That specific article stays incomplete — the class it belongs to is fixed. Weakening TLS verification is not the fix; naming the reason in the result is what this does instead.


mronkko:
Pairs with #480. These are independent commits on independent branches, but the DOI in #474 is only fully repaired by both: #480 gives it the right item type, this one gives it the fields CrossRef omits. Either alone is an improvement.

PR #480, opened 2026-08-23
fix: map CrossRef's whole type vocabulary in add_by_doi

### LINKED ISSUE #476: CROSSREF_TYPE_MAP covers 13 of CrossRef's 30 types, so add_by_doi silently drops journal, volume, issue and pages for the other 17

mronkko:
Fixes #476 — 17 of CrossRef's 30 types fell through to `document`, dropping journal, volume, issue and pages.

**Best reviewed together with the `fix/add-by-url-embedded-metadata` PR (#474).** See that PR's note; the two complete each other on the reported DOI.

## What was wrong

`add_by_doi` writes fields into an item template and keeps only what the template has room for. A `document` template has no `publicationTitle`, `volume`, `issue` or `pages` — so an article that missed the table arrived with all four gone, no warning, and no way to tell from the item that anything had been lost.

The case that surfaced it: an ordinary journal article whose publisher registered it with CrossRef as `journal-issue`, sitting in a real 15k-item library as a `document` with no journal, volume, issue or pages.

## What changed

- **Container types** (`journal-issue`, `journal-volume`, `journal`) → `journalArticle`. Zotero has no record for "a whole issue", publishers do register articles as these, and `journalArticle` has somewhere to put the values the metadata carries.
- **Book part spellings** (`book-part`, `book-section`, `book-track`) join `book-chapter` on `bookSection`.
- **`proceedings` / `proceedings-series`** → `book`, distinct from `proceedings-article`, which is the paper inside.
- **`dataset` and `standard`** → Zotero's native item types instead of `document`. Both existed all along while the table dropped their fields.
- Types with no Zotero equivalent (`component`, `grant`, `other`) still resolve to `document` — but by decision, not fallthrough.
- **An unmapped type is now reported.** CrossRef adds types over time, so full coverage today is not a fix on its own. A miss names the type, what was dropped, and the `zotero_update_item(item_type=...)` remedy.

The suggestion to stop treating a lookup miss as a confident assertion came from the reporting agent, and is the part of this worth keeping.

## Relationship to #465

Same defect, different table and different code path: #465 is `CSL_TYPE_MAP` in `citation_import.py` on the `csl_json` path, this is `CROSSREF_TYPE_MAP` in `tools/_helpers.py` on the `add_by_doi` path. #466 covers the other one, and has been extended with the same container types.

## Note for the reviewer

One existing test asserted `standard -> document` with no rationale, pinning the lossy behaviour. Updated, with the reasoning stated.

## Testing

50 new tests in `tests/test_crossref_type_coverage.py`, including a parametrised check that every type in CrossRef's published vocabulary is mapped and every target is a real Zotero item type — so the next vocabulary addition fails a test rather than silently dropping fields. Full suite green.

Verified against live CrossRef for the three DOIs in the report.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - **`zotero_set_item_parent` sets or clears an item's parent.** It can parent standalone notes and attachments, move them between parents, or make them top-level again.
 
+### Fixed
+- **`zotero_add_by_doi` maps CrossRef's whole type vocabulary, instead of dropping fields for seventeen of its thirty types.** An unmapped type is not a labelling problem: the item is built from a Zotero template, and the `document` template has no `publicationTitle`, `volume`, `issue` or `pages`, so everything type-specific was discarded on the way in and nothing said so. `journal-issue`, `journal-volume`, `book-part`, `book-section`, `book-track`, `proceedings`, `reference-book`, `report-series` and nine others were all missing. The case that surfaced it was an ordinary journal article whose publisher registered it with CrossRef as `journal-issue`, sitting in a real library as a `document` with no journal, volume, issue or pages. `dataset` and `standard` now use Zotero's native types rather than `document`. Types with no Zotero equivalent (`component`, `grant`, `peer-review`, `other`) still become `document`, but by decision rather than by fallthrough — and a type outside the table entirely is now named in the result, with the `zotero_update_item(item_type=...)` remedy, so the next vocabulary addition cannot repeat the silent loss. Reported by the systematic-review agent, which found the mistyped items by cross-checking its library against CrossRef.
+
 ## [0.9.1] - 2026-08-05
 
 ### Changed
--- a/src/zotero_mcp/tools/_helpers.py
+++ b/src/zotero_mcp/tools/_helpers.py
@@ -20,21 +20,79 @@ from zotero_mcp.utils import _paginate
 # Constants
 # ---------------------------------------------------------------------------
 
+# Crossref's own type vocabulary -> Zotero item types.
+#
+# An unmapped type does not merely get a wrong label: the caller writes
+# fields into an item template, and ``document`` has no publicationTitle,
+# volume, issue or pages, so everything type-specific is dropped on the
+# floor. Seventeen of Crossref's thirty types used to be missing here — a
+# journal article registered by its publisher as ``journal-issue`` arrived
+# as a ``document`` with the journal name, volume, issue and page range gone
+# and nothing said. So the table covers the whole vocabulary, and anything
+# genuinely outside it is reported rather than quietly defaulted (see
+# ``crossref_type_note``).
 CROSSREF_TYPE_MAP = {
     "journal-article": "journalArticle",
+    # Container types. Zotero has no record for "a whole issue" or "a whole
+    # volume", and publishers do mis-register ordinary articles as these —
+    # journalArticle keeps the journal, volume, issue and pages that the
+    # metadata actually carries, where document would discard all four.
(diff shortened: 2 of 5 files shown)


mronkko:
Pairs with #479 — see that PR's note. #480 maps the type; #479 fills the fields CrossRef leaves empty. Reviewable separately, but the reported DOI needs both to come out right.

PR #483, opened 2026-08-23
fix: accept a stringified rect in zotero_create_annotation (#459)

### LINKED ISSUE #459: zotero_create_annotation area mode (rect) is unreachable from MCP clients that stringify untyped params
## Summary

`create_annotation`'s `rect` parameter is declared `rect: list[float] | None` (tools/annotations.py:1435). The JSON schema FastMCP publishes for it carries no type information (`"rect": {}`), so at least one client (Claude Code 2.1.220) serializes the array argument as a JSON **string**. Pydantic then rejects it before the tool body runs:

```
1 validation error for call[create_annotation]
rect
  Input should be a valid list [type=list_type, input_value='[0.1628, 0.0985, 0.7783, 0.7272]', input_type=str]
```

Result: **area annotations cannot be created at all from Claude Code**, including via the ready-to-paste call that `zotero_get_page_layout` itself suggests. Highlight mode (`text=`) is unaffected.

## Reproduction (zotero-mcp 0.9.1, Claude Code 2.1.220, macOS, Zotero 9.0.6)

1. `zotero_get_page_layout(attachment_key='...', page=3)` → returns a region and a suggested call.
2. Invoke `zotero_create_annotation` with that exact `rect=[0.1628, 0.0985, 0.7783, 0.7272]`.
3. Pydantic `list_type` validation error, reproducibly (two attempts, including with no spaces in the array literal).

## Suggested fix

The codebase already has the right pattern one parameter over: `tags: list[str] | str | None` in the same signature, with JSON-string parsing in the body. Apply the same to `rect`:

```python
rect: list[float] | str | None = None
...
if isinstance(rect, str):
    try:
        rect = json.loads(rect)
    except json.JSONDecodeError:
        return "Error: rect must be a JSON array of four numbers, e.g. [0.15, 0.22, 0.6, 0.35]"
```

The existing `len(rect) != 4` / range validation then applies unchanged. `zotero_get_item_children` (`item_key`) documents the same string-tolerant behavior, so this brings `rect` in line with the package's own convention.

## Environment

- zotero-mcp 0.9.1 (PyPI `zotero-mcp-server`, uv tool install)
- Claude Code 2.1.220 (MCP over stdio)
- Zotero 9.0.6, local API + hybrid writes
- Python 3.12

AmirF194:
## Root cause

`zotero_create_annotation`'s `rect` parameter was typed as `list[float] | None`
and dispatch checked `isinstance(rect, (list, tuple))`. The sibling `tags`
parameter on the same tool already tolerates a client that stringifies
untyped/loosely-typed arguments before dispatch (a common behavior for
JSON-schema-strict MCP transports), via `_helpers._normalize_str_list_input`.
`rect` had no equivalent, so a client sending `rect="[0.15, 0.22, 0.6, 0.35]"`
(JSON text) instead of an actual list got "rect must be exactly four numbers"
and area-mode annotation was unreachable through that transport.

## Fix

Added `_normalize_float_list_input` in `_helpers.py`, mirroring
`_normalize_str_list_input`'s JSON-parse fallback for a fixed-length numeric
list, and routed `rect` through it before the length/type check in
`create_annotation`. Invalid shapes (wrong length, non-numeric entries,
malformed JSON, unsupported types) still return the same "rect must be
exactly four numbers" message as before.

## Verified

- New regression tests in `tests/test_area_annotations.py`: fail on `main`
  (stringified rect rejected), pass on this branch, run in Docker
  (`python:3.10-slim` and `python:3.12-slim`, the CI matrix floor and ceiling).
- Full suite (`pytest --ignore=tests/test_lifespan.py`, matching CI's own
  command): 2379 passed, 16 skipped on both Python versions.
- Coverage: every changed/added line in `_helpers.py` and `annotations.py` is
  hit by the new tests (`coverage run --source=src/zotero_mcp/tools`).
- Not verified: macOS/Windows runners (CI's matrix includes both at 3.12).
  This change has no OS-specific dependency, but I only ran Linux containers.

Fixes #459.

### DIFF
--- a/src/zotero_mcp/tools/_helpers.py
+++ b/src/zotero_mcp/tools/_helpers.py
@@ -366,6 +366,37 @@ def _parse_library_id_param(value: int | str | None) -> int | None:
     return int(value)
 
 
+def _normalize_float_list_input(value, length, field_name="value"):
+    """Normalize a fixed-length numeric list that MCP clients may stringify.
+
+    Mirrors ``_normalize_str_list_input``: some MCP transports stringify
+    untyped/loosely-typed arguments before dispatch, so a client-side
+    ``[x, y, w, h]`` list can arrive here as the JSON text ``"[x, y, w, h]"``
+    instead. Returns ``None`` (never raises) when ``value`` is not a JSON
+    array of exactly ``length`` numbers, so callers keep their own
+    user-facing error message for the invalid-shape case.
+    """
+    if isinstance(value, (list, tuple)):
+        candidate = value
+    elif isinstance(value, str):
+        try:
+            parsed = json.loads(value)
+        except json.JSONDecodeError:
+            return None
+        if not isinstance(parsed, list):
+            return None
+        candidate = parsed
+    else:
+        return None
+
+    if len(candidate) != length:
+        return None
+    try:
+        return [float(v) for v in candidate]
+    except (TypeError, ValueError):
+        return None
+
+
 def _normalize_str_list_input(value, field_name="value"):
     """Normalize list-like user input into a list of non-empty strings."""
     if value is None:
--- a/src/zotero_mcp/tools/annotations.py
+++ b/src/zotero_mcp/tools/annotations.py
@@ -1444,7 +1444,7 @@ def create_annotation(
     attachment_key: str,
     page: int,
     text: str | None = None,
-    rect: list[float] | None = None,
+    rect: list[float] | str | None = None,
     comment: str | None = None,
     color: str = "#ffd400",
     tags: list[str] | str | None = None,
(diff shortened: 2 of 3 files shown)


54yyyu:
Reviewed and verified. Merging once `docker (all)` finishes — everything else is green, Windows and macOS included.

I ran the edge cases against your `_normalize_float_list_input` directly and they all behave as described: the issue's exact stringified rect parses, a real list passes through untouched, and wrong length / malformed JSON / a JSON object / non-numeric entries / unsupported types all return `None` so the caller keeps its own error message. Mirroring `_normalize_str_list_input` rather than inventing a second convention is the right call.

**One thing your fix led me to.** I swept every registered tool for list/dict-typed params without string tolerance. 29 already tolerate it; exactly two did not — your `rect`, and `zotero_advanced_search.conditions`. That second one is the same bug and arguably worse:

- its description promises `conditions: list of {field, operation, value} dicts (also accepts a JSON string)`
- its body has an `isinstance(conditions, str)` → `json.loads` branch to handle exactly that
- but the annotation is `list[dict[str, str]]`, so Pydantic rejects the string before the body runs and **that branch is dead code**

Confirmed against the validation layer:

```
advanced_search with STRING conditions: REJECTED -> ValidationError
conditions
  Input should be a valid list [type=list_type, input_value='[{"field":"itemType",...}]', input_type=str]
```

So the same client that can't reach area annotations also can't use the documented JSON-string form of advanced search. I'll fix that in a follow-up on top of yours rather than asking you to widen the scope of a PR that's already complete and correct for what it claims.

Two notes for the record, neither affecting the merge:

- On current FastMCP the published schema for `rect` is `{"anyOf": [{"items": {"type": "number"}, "type": "array"}, {"type": "null"}]}`, not the bare `{}` the issue reports. That's likely a FastMCP version difference since 0.9.1. It doesn't change anything — a client can still stringify, and tolerating it is right either way.
- `field_name` is unused in the new helper. Harmless, and it keeps the signature parallel to its sibling, so I'd leave it.

Thanks for picking this up.

AmirF194:
Thanks for the thorough check, and for catching the conditions bug while you were in there. Appreciate you mirroring the existing helper convention back at me instead of just taking the patch as-is.

PR #487, opened 2026-08-24
Global search across all libraries, for metadata and semantic search (#163)

### LINKED ISSUE #163: Feature request: global search across all libraries (metadata and semantic)

mronkko:
Closes #163. Builds on #396 (phase 1, `group_id` tagging) and #417 (the SQLite backend), which is what makes this possible.

## The problem

Every search is scoped to one library, so finding an item means already knowing which library holds it: `zotero_switch_library`, search, repeat. A paper in a group library is invisible to a personal-library search, and the empty result gives no hint to look elsewhere — the only clue is already knowing the item exists.

Reproduced against a real library. Searching for a paper that lives in a group library returns three unrelated papers via the fallback cascade; with the new flag it returns the paper, labelled with its library.

## What this adds

`search_all_libraries=True` on `zotero_search_items`, `zotero_advanced_search` and `zotero_semantic_search` (`--all-libraries` on `zotero-cli`). One pass over the personal library and every group library, each result labelled:

```
**Library:** AI in entrepreneurship (groupID=6015547)
```

Duplicates across libraries are returned as-is. The same paper filed twice is two items, and collapsing them would hide which copies exist where.

## Why it requires `ZOTERO_SEARCH_BACKEND=sqlite`

It is gated on the SQLite backend and **refused** without it, rather than falling back. The Zotero API can only search one library per request, so an API implementation could only replay a single-library search against each library in turn — a different and far slower operation than the one being asked for.

The same reasoning applies inside the global path: nothing falls back. A query the SQL backend cannot express returns an error naming the reason, never a quietly narrower answer. The failure this exists to prevent is a result that looks global and is not.

## Tags are global; collections are not

Two limits follow from Zotero's own storage, and they go in opposite directions.

**Collections are per-library** — `collections.libraryID` is `NOT NULL` with `UNIQUE (libraryID, key)` — so `collection_key`, `collection` conditions and `include_subcollections` are rejected in this mode. "Global, but inside collection X" is not a coherent question.

**Tags are not.** Zotero keeps one database-wide table, `tags(tagID INTEGER PRIMARY KEY, name TEXT NOT NULL UNIQUE)`, with no `libraryID` at all; every library's items point at the same `tagID`. On the library this was developed against, one `Entrepreneurship` tagID covers 458 items spanning 12 libraries. So tag filters work globally and are the natural way to slice a global search — a global advanced search on that tag returns items from 11 distinct libraries.

That required teaching the SQL backend the boolean tag DSL, which it never had: `search_items_sql` returned `None` for *any* `tag` argument and punted the whole query to pyzotero. The documented syntax (`["methods OR methodology", "-draft"]` — entries ANDed, `OR` or Zotero's `||` disjoining within an entry, leading `-` excluding) now compiles to SQL on top of the existing `_tag_condition`.

**Independent of #163, this removes a silent fallback that cost every single-library SQLite user the backend's speed on every tagged query.**

Exclusion is a plain `NOT EXISTS`, matching what Zotero's API returns — an item carrying no tags satisfies `-draft` — rather than advanced search's "a missing value satisfies nothing" rule, which answers a different question. Wildcard entries still fall back rather than answering differently.

## Behaviour change: `zotero_semantic_search` defaults to the active library

It previously searched every indexed library whenever `library_id` was omitted, making it the one search tool whose unqualified behaviour differed from the rest — the same query meant "this library" through two tools and "all libraries" through the third. It now defaults to the active library like the others, with global behind the same flag. `library_id` still scopes to a specific library and is mutually exclusive with it.

This is the only breaking change here and is called out in the changelog's **Upgrading** note.

## Fix: a semantic hit from another library could be found but not read

This is the concrete failure #163 reports. Results are hydrated through the pyzotero client, which is bound to exactly one library, so a hit from any other library 404'd and came back as `Could not fetch full item data` — the index knew about the paper and the tool could not show it. `tests/test_semantic_multilibrary.py` recorded this in its own docstring as *"still deferred upstream"*.

Foreign hits are now hydrated from `zotero.sqlite` in one batched query, arriving with their library attribution attached. Hits from the client's own library still go through the client, which is authoritative for them.

**No re-index is required.** An index built before #396's tagging carries no `group_id` at all, so a foreign hit there cannot be recognised as foreign up front — those are tried against the client first and fall back to the local database when it cannot serve them. That is the state of every index that has not been rebuilt since #396, so making the fix depend on re-indexing would have left the reported bug in place for most users.

## Testing

36 new tests. Full suite: **2474 passed, 13 skipped, 1 failed** — the failure is `test_issue_455_toc_stdout_pollution`, which fails identically on pristine `main` in this environment (venvs set `site.ENABLE_USER_SITE = False`, so the test's own `usercustomize` precondition never loads); both that test and `pdf_utils.py` are byte-identical to `main` here.

- `tests/test_global_search.py` — global scope across libraries, library attribution, feed-library exclusion, cross-library duplicates preserved, `collection` conditions refused when global, full tag-DSL coverage, `get_items_by_keys`.
- `tests/test_search_backend_routing.py` — the refusal names the env var; `collection_key` rejected; tag filters served in SQL; `**Library:**` labels rendered. All using a pyzotero double that fails the test if the API path is touched, so "no silent fallback" is enforced rather than asserted. Includes a guard that the fallback cascade keeps global scope — a retry reverting to the active library is exactly the silent narrowing this design is meant to prevent.
- `tests/test_semantic_multilibrary.py` — the default flip, the `library_id` conflict, cross-library enrichment, and the untagged-index fallback.

Verified live against a ~57k-item library with 27 accessible libraries, including the exact case from #163.

Single-library output is byte-identical: `format_item_result`'s new `show_library` defaults to `False`, and only global searches pass `True`.


mronkko:
I have been running this feature locally for about a month now. The PR is just a port against the current main of something that I use as my daily driver.

Issue #485, opened 2026-08-24
Windows: server stalls on tool calls — `_sync_semantic_update` imports chromadb before checking config


llrllr0123:
## Describe the bug

On Windows, after starting the MCP server, **every tool call hangs until the client timeout** (e.g. `-32001 Request timed out` after ~120s). The server never responds to `tools/call`; only `initialize` and `tools/list` work. Each fresh server process reproduces the stall on its first tool call.

## Environment

- OS: Windows 11 x64
- zotero-mcp-server **0.10.0** (installed via `uv tool install 'zotero-mcp-server[all]'`)
- fastmcp 3.4.7 / Python mcp SDK 1.29.0 (server side)
- Client: @modelcontextprotocol/sdk 1.29.0 (Node) over stdio
- Zotero 10.0 desktop, local API on 23119

## Root cause

In `zotero_mcp/_app.py`, `_sync_semantic_update()` does `from zotero_mcp.semantic_search import create_semantic_search` **at the top of the function, before the config check**. The `semantic_search` module imports `chromadb` → `numpy`, a heavy dependency chain.

This import runs in a background thread (via `asyncio.create_task(_background_update())` in the server lifespan). On Windows the import wedges the process — the background thread gets stuck inside the `numpy` import (import-lock / GIL contention), and while it is stuck the whole process cannot execute any tool. The event loop stays idle (it can still answer `initialize`/`tools/list`), but tool execution threads cannot acquire the GIL, so `tools/call` is dispatched and never completes.

The critical detail: when `auto_update` is `false` in `~/.config/zotero-mcp/config.json` (the default), the function would return immediately after the config check — **but the import already happened before the check**, so the damage is done on every startup.

## Evidence

`py-spy dump` of the wedged process shows:
- MainThread: idle in the asyncio event loop (`run_forever`)
- Background thread `asyncio_0`: stuck in `numpy/_core/multiarray.py` module import (import chain: `zotero_mcp.embeddings` → `chromadb` → `numpy`)

A tool call is logged as `Dispatching request of type CallToolRequest` and then nothing happens until the client times out.

## Suggested fix

Move the import **after** the config check so a server with auto-update disabled never imports chromadb:

```python
def _sync_semantic_update() -> None:
    config_path = Path.home() / ".config" / "zotero-mcp" / "config.json"
    if not config_path.exists():
        return
    try:
        with open(config_path) as f:
            cfg = json.load(f)
[...]

PR #491, opened 2026-08-24
fix: serialize adds per normalized identifier (#486)

### LINKED ISSUE #486: The Zotero API lock has no stated policy for item creation, and the guard the code names doesn't cover it
Splitting this out of #443, where it came up while narrowing the API lock. It
isn't a blocker for that PR — it's a question about the codebase's concurrency
model that predates it and is better answered once, in one place, than per-hunk.

### The lock does two jobs, and only one is written down

`_zotero_api_lock` is documented as protecting the single-threaded local API on
port 23119 from concurrent access. That's job one, and it's explicit.

Job two is implicit: wherever the lock happens to be held across both a dedup
lookup and the create that follows it, it also makes that check-then-write
sequence atomic within the process. Nothing says this is intended, and nothing
tests it — it's a side effect of how wide the lock happened to be.

Narrowing the lock (correctly, to keep third-party network work out of it)
removes job two wherever a fetch now sits between the check and the write:

```
lock → find_existing_items → unlock → CrossRef / page fetch → lock → create_items → unlock
```

Two tool calls can both pass the dedup check and both create.

### The guard the code names cannot close this

`write.py` justifies the narrowing with "A6's version-checked retry is what
guards concurrent writes, not this lock." That is true for updates and not for
creates:

- `_update_item_with_version_retry` is wired into two call sites,
  `batch_update_tags` and `batch_update_extra`. Neither is on an add path.
- It could not help if it were. It catches `PreConditionFailedError` — a stale
  `If-Unmodified-Since-Version` on an *existing* item. Two concurrent
  `create_items()` POSTs for the same DOI produce two new keys and no version
  conflict at all. There is nothing for a version check to detect.

So the fallback the comment points at is load-bearing in the argument and absent
in the code.

### Who actually hits it

Any caller issuing adds in parallel — which is the normal shape for an agent
with this toolset, and explicitly the ask in #46 (batch SLR imports). Serial
callers never see it.

Worth noting the interaction with #441: while `if_exists` dedup doesn't match by
DOI on the Web API at all, this race is mostly latent there and live on the local
API. Fixing #441 makes dedup work — and makes the window real on both.

### What I'm asking for

A decision on the intended model, not necessarily code. Roughly:

1. **Accept and document it.** Say plainly that adds are not serialized and
   concurrent adds may duplicate, and drop the claim that the version retry
   covers it.
2. **Serialize per identifier.** A short-lived in-process lock keyed by
   normalized DOI/ISBN/URL, held across check-and-create. Closes the common case
   (one process, parallel tool calls) without putting third-party network work
   back under the global lock.
3. **Converge after the fact.** Leave the race and lean on
   `zotero_merge_duplicates(auto=True)` from #395.

Option 2 looks like the smallest thing that makes the comment true, but I'd
rather have the maintainer's read on the intended model than guess — hence an
issue rather than a patch.

mronkko:
Implements option 2 from your answer on #486, to the scope you set there.

Closes #486

## What it does

A short-lived in-process lock keyed on the normalized identifier, held across a dedup re-check and the create.

The first dedup check still runs before the fetch, exactly as now. The check inside the lock is a **repeat, not a move** — the CrossRef and page fetches have to stay outside a held lock, which is what the narrowing was for, so the only way to make the pair atomic is to check again on the other side of the fetch.

## Against the four points you set

- **Keyed on the normalized identifier**, reusing `normalize_doi` / `_normalize_isbn` / `_normalize_arxiv_id`, so ISBN-10 and ISBN-13 of one book take the same lock, as do a bare DOI and its `doi.org` URL. URLs have no normalizer, so exact-after-strip, matching what #443 settled on for collapsing repeats — deliberately no case or trailing-slash folding, since either can be a genuinely different page.
- **Held across check-and-create only**, never across CrossRef or a page fetch.
- **In-process only, said out loud** — the module comment states plainly that two servers against one library still race. Being vague there is the mistake the issue was about.
- **An unnormalizable token takes no lock** rather than sharing one, so a batch of junk identifiers cannot serialize itself.

## Two things you didn't specify, and what I chose

**Lock ordering.** The identifier lock is always acquired *outside* the API lock, so the two have one consistent order everywhere. On the batched DOI path the chunk's identifier locks are taken in sorted key order, which is what makes holding several at once deadlock-free.

**Registry lifetime.** Entries are reference-counted and dropped when the last holder leaves. Without that, a 16k-item import would leave 16k lock objects behind in a long-running server.

## Cost, stated plainly

On the batched DOI path this adds **one extra dedup read per DOI**. It does not undo #A5/#A4 — CrossRef is still one batched request per 50 DOIs and creates are still one POST per 50. Only the dedup read, which was always per-DOI, now happens twice. There is no batched form of `find_existing_items` to do better with; if you'd rather have one, that's a bigger change and a separate one.

On the batched path the OA PDF attach inside `_create_and_attach_batch` runs while the chunk's identifier locks are held. That is deliberate and is not what the narrowing was protecting: a per-DOI lock held across that DOI's own download blocks only a concurrent add of the same item, never unrelated work — the opposite of the global lock's blast radius.

## What is covered, and what isn't

Covered: `add_by_doi` (single and batched), `add_by_isbn`, `add_by_url`, `_add_by_arxiv`.

**Not covered: the BibTeX and CSL-JSON importers.** They dedup in `_maybe_reuse_existing` and create in `_create_and_attach_batch` with nothing spanning the two. The window there is narrower — the metadata comes from the input file, so there is no network fetch in between — but it is not closed, and I'd rather say so than let this PR imply otherwise. Closing it means either threading the lock through `_create_and_attach_batch` (shared by three importers, and filtering entries inside it would desynchronise the `chunk`/`titles`/`results` indices) or hoisting the dedup into each caller the way `add_by_doi` now does. Happy to do the second in a follow-up if you want it.

## On the test I removed

I wrote a timing-based test that reproduced the race with two threads, and then deleted it. Every version either **passed with the lock taken out** — proving nothing — or deadlocked on its own barriers. Shipping it would have been the same shape of mistake as the comment this issue was filed about: a guard that is load-bearing in the argument and absent in fact.

What is pinned instead is deterministic and is what the fix actually claims: each add path enters the lock with the right key, and performs its dedup re-check *while inside it*. Those tests fail when the lock is removed from the path — I checked, rather than assuming.

## Testing

- `2698 passed, 42 skipped`, no failures.
- 14 new tests: keying (ISBN-10/13 equivalence, DOI spellings, kind collisions, URL exactness, unnormalizable → no key), mutual exclusion (same identifier serializes, different identifiers don't, junk doesn't, re-entrancy, registry doesn't leak, entry survives a waiter), and the two ordering tests above.

## Commits

| | |
|---|---|
| `cb2b1e2` | fix: serialize adds per normalized identifier (#486) |

4 files changed, +519 / −16.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- **Concurrent adds of the same identifier no longer create duplicates (#486).** The Zotero API lock exists to protect the single-threaded local API, and making check-then-create atomic was only ever a side effect of how wide it happened to be. Narrowing it — correctly, to keep CrossRef and page fetches out of a held lock — removed that side effect wherever a fetch now sits between the dedup check and the write, so two parallel tool calls could both pass the check and both create. Nothing downstream would catch it: the version-checked retry guards updates to an item that already exists, and two `create_items()` POSTs yield two new keys with no version to conflict on. Adds by DOI, ISBN, URL and arXiv ID now take a short-lived lock keyed on the *normalized* identifier — so ISBN-10 and ISBN-13 of one book, or a bare DOI and its doi.org URL, take the same lock — held across a re-check and the create only, never across third-party network work. An identifier that cannot be normalized takes no lock rather than sharing one, so a batch of junk tokens cannot serialize itself. This is in-process: it closes one server issuing parallel tool calls, which is the reported case and the normal shape for an agent driving this toolset; two servers against one library still race, and nothing here changes that.
+
 **Upgrading:** `zotero_semantic_search` now defaults to the active library instead of every indexed library. If you relied on the old implicit behaviour, pass `search_all_libraries=True`.
 
 ### Added
--- a/src/zotero_mcp/tools/_helpers.py
+++ b/src/zotero_mcp/tools/_helpers.py
@@ -1,11 +1,13 @@
 """Shared private helpers used across tool modules."""
 
+import contextlib
 import hashlib
 import json
 import os
 import re
 import socket
 import tempfile
+import threading
 from ipaddress import ip_address
 from urllib.parse import urljoin, urlparse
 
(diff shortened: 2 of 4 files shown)


54yyyu:
**Maintainer agent response**

---

Merged, along with #490! Reviewed by execution rather than by reading, since that is where this kind of change goes wrong.

Lock ordering holds on every path: the identifier lock is acquired outside `zotero_api_lock` in `add_by_url`, `_add_by_arxiv`, `add_by_isbn`, and the batched DOI path through `_create_and_attach_batch`.

I went looking for a deadlock in the batched path, because it sorts by `payload["doi"]` while the lock key is the normalized DOI, and two concurrent batches naming the same DOIs in different forms would then acquire in different orders. It is not there! `payload["doi"]` is already normalized (`write.py:1530`, from `:1381`), `normalize_doi` is idempotent, and every key carries the same `doi:` prefix, so both sorts give the same order. Worth a comment at the sort if you ever touch it, since the guarantee rests on a fact two functions away.

Measured with real threads:

| | |
|---|---|
| same DOI | serialized |
| bare DOI vs `doi.org` URL | serialized, same lock |
| ISBN-10 vs ISBN-13 | serialized, same lock |
| different DOIs | not serialized, correct |
| unnormalizable junk | not serialized, correct |
| registry after all runs | empty, refcounting does not leak |

**One gap.** DOI case is not folded, so `10.1007/ABC` and `10.1007/abc` take different lock keys and two concurrent adds of that one DOI can still both create. It is the race this closes, just narrower: it needs two concurrent calls naming one DOI in different case, which mixed-source imports do produce.

Your own helper in the same file already handles it, with a docstring that states the reason:

```python
def _doi_dedup_key(tok: str):
    """DOIs are case-insensitive, and CrossRef echoes canonical case."""
    normalized = _helpers._normalize_doi(tok)
    return normalized.lower() if normalized else None
```

`identifier_lock_key` reaching for the same folding would close it. I did not do it in the merge because it is your code and where to fold belongs with you: inside `identifier_lock_key` only, or in `normalize_doi` itself, which is wider and has dedup consequences.

One thing I changed in your work while merging, so it does not surprise you later. #488 re-added `load_update_config` to `semantic_search.py`, which had moved to a ChromaDB-free module while your branch was out. I kept your streaming helpers and dropped the duplicate definition. It is re-exported, so nothing importing it from `semantic_search` changed. That module has since been renamed `config_light` and grown the reranker readers too, after @w-clary found the same import-before-gate pattern in `cli.py` that #485 was about.

All six of your PRs are in 0.11.0!

Issue #492, opened 2026-08-24
Semantic search hits from group libraries fail enrichment: results render as bare item keys


w-clary:
## Describe the bug

Since #163 the indexer stamps each document with its library's `group_id`, and `search()` spans all indexed libraries by default. But `_enrich_search_results` fetches every hit with `self.zotero_client.item(item_key)`, a single client bound to the *active* library. A group item is not reachable at `/api/users/<id>/items/<key>`, so the fetch 404s and the result renders as `## 1. Item WTDQVCR3` with no title or authors.

The `library_id` parameter on `zotero_semantic_search` does not help here: it filters the Chroma query but never rescopes the enrichment client. `zotero_switch_library` before every search works, but only for one library at a time, which defeats cross-library search.

Still present on `main` (`semantic_search.py`, ~line 3716). In my library (~97% group items across 3 group libraries) essentially every result was affected.

## Environment

- Windows 11 x64, Python 3.12
- zotero-mcp-server 0.9.1 (also verified on `main` @ 8f8df8d)
- Local mode (`ZOTERO_LOCAL=true`) with personal + 3 group libraries indexed (5,294 docs)

## Reproduction

1. Index a personal library plus at least one group library (`zotero-mcp update-db`).
2. Call `zotero_semantic_search` with a query whose best matches live in the group library, with env/active library set to the personal library.
3. Group hits come back with `Could not fetch full item data: ... 404` and render as bare item keys; the same keys resolve fine after `zotero_switch_library` to that group.

## Suggested fix

Resolve the client per result from `meta["group_id"]` (0 for personal, otherwise that group), cached per group id. Running locally without issue:

```python
def _client_for_meta(self, meta):
    if not isinstance(meta, dict):
        return self.zotero_client
    gid = meta.get("group_id")
    try:
        gid = int(gid)
    except (TypeError, ValueError):
        return self.zotero_client
    try:
        if gid == self._client_group_id():
            return self.zotero_client
    except Exception:
        pass
    cache = getattr(self, "_group_clients", None)
    if cache is None:
        cache = self._group_clients = {}
    if gid not in cache:
        local = os.getenv("ZOTERO_LOCAL", "").lower() in ("true", "yes", "1")
        api_key = os.getenv("ZOTERO_API_KEY")
        if gid == PERSONAL_LIBRARY_GROUP_ID:
            lib_id = os.getenv("ZOTERO_LIBRARY_ID") or ("0" if local else None)
            lib_type = "user"
[...]

PR #497, opened 2026-08-25
fix: hydrate cross-library hits through a client scoped to their library (#492)

w-clary:
Fixes the web-API half of #492, in the shape discussed there: the per-library client resolver is used where `_hydrate_locally` comes up empty, so local mode keeps the batched single-query sqlite path and web-API mode stops being a dead end.

## What changes

`_attach_zotero_items` gains a third source for foreign hits. The order is unchanged for everything that worked before:

1. Foreign hits are batch-hydrated from `zotero.sqlite` (#487) — untouched, still first in local mode.
2. A foreign hit the local database could not serve goes to `_fetch_via_scoped_client`, which resolves a pyzotero client scoped to the hit's own `group_id` — built from the same credentials, cached per id on the instance, with an unreachable library recorded once rather than failing per hit.
3. In-scope hits keep going to the bound client, and the untagged-index retry path is untouched.

Per the credential point raised in the issue: when a hit cannot be served, the error now names the library and what is missing — no local database and no usable credentials, or "the configured credentials could not read it from there: ..." — instead of a bare 404 from a library that was never going to have the item.

One deliberate refusal: under a group-primary web config (`ZOTERO_LIBRARY_TYPE=group`), `ZOTERO_LIBRARY_ID` names a group, so the personal user ID is unknowable and a personal-library (`group_id=0`) hit gets the explicit unreachable message rather than a client built for `users/<groupID>`, whose 404 would blame the wrong library.

Known boundary, unchanged from your fallback: an untagged hit (index predating #396) carries no `group_id`, so web mode has no way to pick a client for it and it still errors; reindexing is the fix there.

## Verification

Unit tests — 6 new/updated in `tests/test_semantic_multilibrary.py`, following the existing `_ScopedZot`/`_FakeReader` conventions, via a `_new_scoped_client` seam:

- web mode, credentials present: foreign hit served by a client built as `(group_id, "group", key, local=False)`
- web mode, no credentials: error names the library and `ZOTERO_API_KEY` (replaces `test_foreign_hit_without_a_local_db_keeps_reporting_the_error`, whose asserted behavior is the bug being fixed)
- credentials that cannot see the group: error names the library
- one client construction per library across multiple hits
- local mode: sqlite stays first; the scoped client only serves a key the snapshot missed
- group-primary web config: a personal-library hit constructs no client and gets the named-library message (an exploding seam proves nothing is built)

Live, against api.zotero.org with a real key on the 4-library index from my #492 report, one query happened to exercise both designed paths:

- hits from a group the key can read: enriched with full metadata over the web API (previously bare keys)
- hits from a group the key genuinely cannot read (verified: 403 on that group's listing, and the group is absent from the key's `/users/<id>/groups`): the named-library credential error, not a silent 404

Full test suite minus `tests/live` passes (2648 passed; the 4 failures in `test_pdf_coordinates.py` fail identically on pristine `main` in my environment). `ruff check` clean — includes one import-order fix in `semantic_search.py` that the `I` rules flag after #485's insertion.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **Cross-library semantic hits are hydrated through a client scoped to their own library when the local database cannot serve them (#492).** The `zotero.sqlite` fallback that #487 added for foreign hits does not exist outside local mode, so a web-API user with group libraries got every cross-library hit back as a bare item key with a 404 — the bound client was asked for a key its library never held, and there was no second chance. Enrichment now falls through to a per-library pyzotero client, built from the same credentials and cached per `group_id`, so ten hits from one group cost one client and an unreachable library fails once rather than per hit. Local mode keeps the batched single-query sqlite path first and only reaches the scoped client for a key the snapshot could not serve, such as an item newer than the copy on disk. When a hit genuinely cannot be served — no credentials, or a key the credentials cannot see — the error now names the library and what is missing, instead of surfacing a 404 from a library that was never going to have the item.
+
 ## [0.11.0] - 2026-08-25
 
 **Upgrading:** `zotero_semantic_search` now defaults to the active library instead of every indexed library. If you relied on the old implicit behaviour, pass `search_all_libraries=True`.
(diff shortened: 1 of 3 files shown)


w-clary:
No rush at all, just confirming this is still on your radar after the 0.11.0 push. The branch is still conflict-free against main, and I'm happy to rebase or adjust the shape whenever you get to it. (Same goes for the #492 re-run confirmation and #498, whenever you're back in the repo.)

PR #502, opened 2026-08-29
RFC: serve reads through a backend-neutral port, implemented over zotero.sqlite

mronkko:
> **This is a proof of concept, not a merge request.** I'm looking for a verdict
> on the direction before polishing anything. Happy to split it into a reviewable
> sequence — a proposed split is at the bottom — or to drop it entirely if you'd
> rather not take this on.

## The problem

Every read tool calls pyzotero directly. Two consequences:

1. **Closing Zotero desktop breaks reading a library that is sitting readable on disk.** With both APIs unreachable, 12 read tools fail outright.
2. **The access patterns are shaped by HTTP** — paged at 100 rows, one round trip per item — so they stay expensive even when the data is local. `OFFSET` on a 2.3 GB file re-scans; `children()`-per-key is N+1 only because each call used to be a round trip.

SQLite is not the bottleneck. The interface shape is.

---

# Architecture

## The port

`src/zotero_mcp/library.py` declares `LibraryBackend`, a `Protocol` of 14 read
operations shaped around what the tools actually need — **plural by default,
unpaged by default**:

```python
class LibraryBackend(Protocol):
    name: str
    # items
    def get_item(key) -> dict | None
    def get_items(keys) -> dict[str, dict]
    def get_children(keys, *, item_type=None) -> dict[str, list[dict]]
    def get_related(key) -> list[dict]
    def find_by_citation_key(citekey) -> dict | None
    def recent_items(*, limit, collection_key=None) -> list[dict]
    def top_items(*, limit) -> list[dict]
    def list_items(item_type=None, *, limit, tag=None) -> list[dict]
    # collections
    def list_collections(*, include_trashed=False) -> list[dict]
    def get_collection(key) -> dict | None
    def collection_items(key, *, include_subcollections=False) -> list[dict] | None
    # tags
    def list_tags(*, limit=None) -> list[str]
    # search
    def search_items(query, *, qmode, item_type, tag, limit,
                     collection_keys=None, include_subcollections=False) -> list[dict]
    # files
    def attachment_paths(key) -> list[dict]
```

Two implementations: `SqliteBackend` answers from `zotero.sqlite` in a fixed
number of queries with no HTTP at all; `ApiBackend` preserves today's pyzotero
calls, paging included.

**The move that keeps this tractable:** both return **pyzotero-shaped dicts**
(`{"key", "version", "data": {...}}`), not a new record type.
`format_item_metadata`, `generate_bibtex`, `item_display_title` and every tool's
formatting body already consume that shape, and `local_db.row_to_api_item`
already produces it. Tools changed at their **fetch** lines, not their **format**
lines — so the behaviour accumulated in them (trash handling, base-field
resolution, collection scope expansion) is untouched.

## Layering, and the deliberate hole in it

```mermaid
flowchart TD
    RT["read tools<br/><small>retrieval · search · annotations · synthesis</small>"]
    WT["write tools<br/><small>write · item_parent · manage_note</small>"]
    PORT["<b>LibraryBackend</b><br/><small>14 read operations · plural · unpaged</small>"]
    SB["SqliteBackend"]
    AB["ApiBackend"]
    LZR["LocalZoteroReader"]
    PYZ["pyzotero"]
    DB[("zotero.sqlite")]
    HTTP{{"HTTP<br/>:23119 · api.zotero.org"}}

    RT -->|get_library_backend| PORT
    PORT -->|ZOTERO_BACKEND=sqlite| SB
    PORT -->|otherwise| AB
    SB --> LZR --> DB
    AB --> PYZ --> HTTP
    WT -.->|"get_zotero_client()<br/>bypasses the port"| PYZ
```

The port is **read-only**. `get_zotero_client()` is unchanged and still serves
~50 write call sites, `zot._retrieve_data` in `fetch_trashed_collections`, and
the `library_id` / `library_type` attribute access. That is why the API client
cannot simply be removed, and why writes are entirely out of scope here.

## Choosing a route

`get_library_backend()` resolves per call. Selection is configuration plus a
liveness check, and it degrades in one direction only:

| Condition | Result |
|---|---|
| `ZOTERO_BACKEND=sqlite` (or legacy `ZOTERO_SEARCH_BACKEND`) **and** a readable database | `SqliteBackend`, scoped to `get_active_group_id()` |
| SQLite asked for, but no local mode or no readable database | `ApiBackend` — silent fallback |
| Anything else, including unset | `ApiBackend` |

Both names resolve through `utils.get_search_backend()`, so the port and the
older search paths cannot disagree about which backend is active.

The reader is cached in a `threading.local`: sqlite3 connections are bound to
the thread that opened them and MCP tool calls arrive on a pool, so one
process-wide reader would raise as soon as a second thread touched it.

The two backends are also **scoped differently**, which matters when a library
switch is in play. `SqliteBackend` is pinned to a `group_id` at construction;
`ApiBackend` inherits whatever library its pyzotero client was built for.

---

# Call paths

Same operation, same returned shape, different work. Three cases cover the whole
difference: a point lookup, a batch, and a full listing.

## `get_item(key)` — a point lookup

<table>
<tr><th align="left">SqliteBackend</th><th align="left">ApiBackend</th></tr>
<tr valign="top"><td>

```
get_item(key)
└─ reader.get_full_items([key])
   ├─ SELECT … FROM items
   │    LEFT JOIN deletedItems      ← trash flag
   │    LEFT JOIN itemAttachments
   │    LEFT JOIN itemNotes
   │    LEFT JOIN itemAnnotations
   └─ _build_full_items()
      ├─ _fetch_fields()      every itemData field
      ├─ _fetch_creators()
      ├─ _fetch_tags()
      ├─ _fetch_item_collections()
      └─ _fetch_relations()
```

**7 queries · 0.4 ms**

</td><td>

```
get_item(key)
└─ zot.item(key)
   └─ GET /items/<key>
      → server-rendered record,
        including meta / numChildren
        that SQLite does not compute
```

**1 request · 3.7 ms**

</td></tr>
</table>

The one case where the API does less absolute work — and it still loses on wall
clock, because a local HTTP round trip costs more than seven indexed queries.

## `get_children(keys)` — the batch that motivated the port

<table>
<tr><th align="left">SqliteBackend</th><th align="left">ApiBackend</th></tr>
<tr valign="top"><td>

```
get_children([25 keys])
└─ reader.get_children_of(keys)
   ├─ SELECT itemID, key … WHERE key IN (…)
   ├─ SELECT … itemAttachments
   │       WHERE parentItemID IN (…)
   │  UNION ALL … itemNotes
   │  UNION ALL … itemAnnotations
   └─ _build_full_items()   ← 5 queries
```

**9 queries · 1.1 ms** — constant in N

</td><td>

```
get_children([25 keys])
└─ for key in keys:            ← N+1
   └─ _paginate(zot.children, key)
      └─ GET /items/<key>/children
             ?start=0&limit=100
         (and again per page)
```

**25 requests · 1,220 ms** — linear in N

</td></tr>
</table>

Child ids come from a `UNION` over the three side tables rather than a
`COALESCE(...)` join on `items`: the union hits an index on each table's
`parentItemID`, while the COALESCE cannot use one — 2.9 ms vs 99.7 ms for 100
parents, measured.

## `list_tags()` — the paging trap

<table>
<tr><th align="left">SqliteBackend</th><th align="left">ApiBackend</th></tr>
<tr valign="top"><td>

```
list_tags()
└─ reader.list_tags()
   └─ SELECT DISTINCT t.name
        FROM tags t
        JOIN itemTags … JOIN items …
       WHERE libraryID IN (…)
       ORDER BY t.name
```

**2 queries · 167 ms** — 17,302 tags

</td><td>

```
list_tags()
└─ _paginate(zot.tags)
   └─ while True:
        GET /tags?start=N&limit=100
        if len(batch) < 100: break
        N += 100
```

**185 requests · 96,090 ms**

</td></tr>
</table>

> **Why paging could not simply be switched off.** `_paginate` stops when a page
> comes back shorter than the page size. Hand it a backend that returns all
> 17,302 tags on the first call and it sees `17302 > 100`, advances `start`, gets
> 17,302 again, and **loops forever**. Un-chattying had to happen by moving
> operations off `_paginate` entirely, not by making a smarter client behind it.
> `_paginate` now lives only on the API side of the port.

---

# Measurements

Run them yourself — `scripts/measure_read_backend.py` is part of this PR and
follows `measure_context_cost.py`'s conventions (`--json`, `--keys N`). SQL
statements are counted by proxying the sqlite3 connection, HTTP requests by
wrapping `httpx.Client.send`; both observed at execution, not estimated.

Against a 36,265-item / 2.3 GB library:

```
operation                     sqlite                      api
                             queries         time    requests           time
----------------------------------------------------------------------------
get_item(1 key)                   7        0.4 ms          1          3.7 ms
get_items(25 keys)                7        0.9 ms          1        100.0 ms
get_children(25 keys)             9        1.1 ms         25      1,220.4 ms
list_collections()                2        3.1 ms          3         42.5 ms
list_tags()                       2      166.6 ms        185     96,090.0 ms
recent_items(25)                  7      164.7 ms          1        183.3 ms
```

**The ratio is not the interesting part — the shape is.** The sqlite column is
flat: the same handful of queries whether you ask for one key or a hundred. The
api column tracks either batch size (`get_children`) or library size
(`list_tags`, 185 pages).

End to end through the tools: `zotero_get_tags` 14,916 ms → 122 ms;
`zotero_get_item_children` on 100 keys 5,658 ms → 8 ms.

**Caveat on generalising these.** This is one library, and an unusually large
one — 17,302 tags is what makes `list_tags` so lopsided. A smaller library will
show smaller absolute gaps. The *scaling shape* is the claim; the specific
numbers are one data point, which is exactly why the measurement script is in
the PR rather than just its output.

---

# Where the routes genuinely differ

Divergences that are not just speed. Each is a place the two implementations
must be kept honest against each other, which is what
`tests/live/test_read_backend_parity.py` is for.

- **Record completeness.** SQLite hydrates every `itemData` field, plus creators, tags, collection membership, relations, `version`, `parentItem` and the trash flag. It does *not* compute the API's server-side `meta` block or `numChildren`. Parity therefore compares keys and rendered fields, never whole records.
- **Trash.** Both return trashed items from a lookup by key and mark them with `data["deleted"]`, which is what renders the trash status. List operations exclude them on both sides unless asked.
- **Error versus empty.** `get_children` omits a parent whose lookup failed and maps a childless parent to `[]`. Callers rely on that to report a bad key differently from an empty one. SQLite, answering in one query, cannot partially fail — so every requested key is always present.
- **Citation keys.** SQLite matches Zotero 7's real `citationKey` field exactly. The API has nothing to query server-side, so it ranks a substring search and verifies the top 25.

---

# Bugs this exposed in existing code

- **`items(itemKey=…)` returns child notes too.** Confirmed at the HTTP level against the local server, so not a pyzotero artifact: two requested keys came back as six items. The old inline `_fetch_projected` filtered these out incidentally; anything else consuming that call did not.
- **`search_items(collection_key=…)` denied collections that exist.** It validated the key over HTTP, so with Zotero closed it answered "Collection not found" for a collection holding 26 items — a confident wrong answer, not an error.
- **`switch_library`'s probe was redundant.** `validate_library_switch` already validates against `zotero.sqlite`, and every read after the switch is served from that same file; the HTTP call only proved the desktop app was running. Skipping it needed one addition, since that validator covered `group` and `feed` but never `user`.
- **`search_by_citation_key` could miss keys that exist.** It ranked a `q=` substring search and checked the top 25, so a key outside that window was reported missing. Zotero 7 stores `citationKey` as a real field — 38,302 of 38,311 items in my library — so SQLite matches it exactly.

---

# Testing

- **`tests/live/test_sqlite_only_reads.py`** — 32 tests. Blocks `127.0.0.1:23119` and `api.zotero.org` and *raises on contact*, so "zero API calls" is asserted structurally rather than claimed.
- **`tests/live/test_read_backend_parity.py`** — 13 tests. Both backends must return identical keys, rendered fields, creators, children, collections, tags and trash status. This is the real safety net for a change this size, and it caught the `itemKey` bug on its first run against a live Zotero.
- Full unit suite unchanged: the same 15 pre-existing environmental failures (subprocess `ModuleNotFoundError`) before and after.

Both live files are gated behind `ZOTERO_MCP_LIVE_TESTS=1` and require opposite
environments — one needs the APIs unreachable, the other needs them live — so on
any given machine one of the two will skip.

---

# Known gaps, stated rather than hidden

- **`UnsupportedByBackend` is raised in 5 places and caught nowhere.** The intended per-call API fallback is not wired up; unsupported reads surface as error strings instead. Needs a decision: implement the fallback, or drop the exception in favour of explicit errors.
- **`search.py`'s two pre-existing SQLite paths** (`_search_items_via_backend`, `advanced_search`) still call `get_local_zotero_reader()` directly rather than going through the port. Folding them needs a scope parameter for global search (`group_id=None`) that the port does not currently have.
- **Pre-existing, independent of this PR:** `search_items_sql`'s `titleCreatorYear` mode omits `publicationTitle`. For the query *Family*, 23 items the API finds are absent from SQLite's full match set — every one matching only in the journal name. Adding `pub_val.value LIKE ?` to the clause list takes the misses to zero. Relatedly, `test_search_items_free_text_parity` compares set-equality at `limit=200` on a query with 592 true matches, so it cannot pass until both are addressed. **That test fails on `main` today**, unrelated to this branch.

There is also a small companion fix I've held back so it doesn't compete for
attention here: `sql_reader` in `tests/live/conftest.py` is a generator fixture
that `return`s without yielding when local Zotero is unreachable, so every
dependent test errors instead of skipping — 13 errors where there should be 13
skips, whenever Zotero desktop is closed. Happy to send it separately whenever
suits.

---

# Happy to split this

If the direction is worth pursuing but this is too much to review at once:

1. `feat: LibraryBackend port + SQLite readers` — `library.py`, `local_db.py`, no tools migrated *(~1,400 lines)*
2. `feat: migrate retrieval.py onto the port` — 10 tools, plus the sqlite-only live suite *(~700)*
3. `feat: migrate annotations, search, synthesis, write` *(~600)*
4. `feat: migrate resources, connectors, discovery, CLI` *(~400)*
5. `test: cross-backend parity suite` *(~180)*
6. `fix: the four bugs above`, one commit each *(~200)*
7. `feat(scripts): measure_read_backend.py` *(~230)*

Each is independently reviewable and the suite stays green at every step.

### DIFF
--- a/scripts/measure_read_backend.py
+++ b/scripts/measure_read_backend.py
@@ -1,0 +1,222 @@
+#!/usr/bin/env python3
+"""Measure what each read backend costs to answer the same question.
+
+`zotero_mcp.library` serves every read operation through one of two
+implementations: `SqliteBackend`, which queries `zotero.sqlite` directly, and
+`ApiBackend`, which makes the pyzotero calls the tools used to make. They
+return the same records; what differs is the work.
+
+This counts that work rather than estimating it. SQL statements are counted by
+proxying the reader's sqlite3 connection; HTTP requests by wrapping
+`httpx.Client.send`. Both counts are observed at the point of execution, so a
+retry, a redirect or a pagination loop shows up as what it is.
+
+The number that matters here is not per-call latency -- it is how each column
+*scales*. The SQLite column is flat: an operation costs the same handful of
+queries whether it is asked for one key or a hundred. The API column tracks
+either the batch size (one request per parent for children) or the size of the
+library (one request per 100 rows for tags). That shape is what the port
+exists to remove, and it is why the ratios below grow with the library rather
+than staying constant.
+
+What this deliberately does NOT measure: correctness or agreement between the
+backends -- that is `tests/live/test_read_backend_parity.py`, which compares
+their actual results. A fast backend that returns the wrong rows would look
+excellent here, so the two are kept separate on purpose.
+
+Both routes are measured when both are available. With Zotero desktop closed
+the API column is reported as unavailable rather than skipped silently: those
+reads working at all is the point of the exercise.
+
+    python scripts/measure_read_backend.py              # table
+    python scripts/measure_read_backend.py --json       # machine-readable
+    python scripts/measure_read_backend.py --keys 100   # bigger batch
+
+Needs ZOTERO_LOCAL=true and a readable zotero.sqlite. Reads only; it never
+writes to the database or to Zotero.
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+import os
+import sys
+import time
+from pathlib import Path
+
+REPO = Path(__file__).resolve().parent.parent
+sys.path.insert(0, str(REPO / "src"))
+
+#: Batch size for the plural operations, unless --keys says otherwise. Large
+#: enough that an N+1 shape is unmistakable, small enough to stay polite to a
+#: live API.
+DEFAULT_KEYS = 25
+
+
+class _CountingConnection:
+    """Proxy that counts `execute` calls on a sqlite3 connection.
+
+    sqlite3.Connection.execute is read-only, so it cannot be patched in
+    place; wrapping is the only way to observe the statement count without
+    changing the reader.
+    """
+
+    def __init__(self, inner, counter):
+        object.__setattr__(self, "_inner", inner)
+        object.__setattr__(self, "_counter", counter)
+
+    def execute(self, *args, **kwargs):
+        self._counter["sql"] += 1
+        return self._inner.execute(*args, **kwargs)
+
+    def __getattr__(self, name):
+        return getattr(self._inner, name)
+
+
+def _install_counters(reader, counter):
+    """Count SQL statements and HTTP requests for the rest of the process."""
+    import httpx
+
+    reader._connection = _CountingConnection(reader._get_connection(), counter)
+
+    original_send = httpx.Client.send
+
+    def counting_send(self, request, *args, **kwargs):
+        counter["http"] += 1
+        return original_send(self, request, *args, **kwargs)
+
+    httpx.Client.send = counting_send
+
+
+def _operations(keys):
+    """(label, callable) pairs, each taking a backend.
+
+    Chosen to cover the three shapes that behave differently: a point lookup,
+    a batch keyed on N, and a listing keyed on library size.
+    """
+    return [
+        ("get_item(1 key)", lambda b: b.get_item(keys[0])),
+        (f"get_items({len(keys)} keys)", lambda b: b.get_items(keys)),
+        (f"get_children({len(keys)} keys)", lambda b: b.get_children(keys)),
+        ("list_collections()", lambda b: b.list_collections()),
+        ("list_tags()", lambda b: b.list_tags()),
+        (f"recent_items({len(keys)})", lambda b: b.recent_items(limit=len(keys))),
+    ]
+
+
+def _measure(backend, operation, counter, kind):
+    """Run one operation, returning its cost or the reason it could not run."""
+    counter["sql"] = counter["http"] = 0
+    started = time.perf_counter()
+    try:
+        operation(backend)
+    except Exception as exc:
+        return {"error": f"{type(exc).__name__}: {exc}"[:110]}
+    elapsed_ms = (time.perf_counter() - started) * 1000
+    return {"calls": counter[kind], "ms": round(elapsed_ms, 1)}
+
+
+def collect(key_count):
+    """Measure every operation on whichever backends are usable here."""
+    os.environ.setdefault("ZOTERO_LOCAL", "true")
+    os.environ["ZOTERO_BACKEND"] = "sqlite"
+
+    from zotero_mcp.client import get_local_zotero_client
+    from zotero_mcp.library import ApiBackend, SqliteBackend
+    from zotero_mcp.local_db import get_local_zotero_reader
+
+    reader = get_local_zotero_reader()
+    if reader is None:
+        raise SystemExit(
+            "No readable zotero.sqlite. Set ZOTERO_LOCAL=true, and ZOTERO_DB_PATH "
+            "if your Zotero data directory is in a custom location."
+        )
+
+    counter = {"sql": 0, "http": 0}
+    _install_counters(reader, counter)
+
+    sqlite_backend = SqliteBackend(reader, 0)
+    keys = [item["key"] for item in sqlite_backend.recent_items(limit=key_count)]
+    if not keys:
+        raise SystemExit("The personal library has no items to measure against.")
+
+    zot = get_local_zotero_client()
+    api_backend = ApiBackend(zot) if zot is not None else None
+
+    rows = []
+    for label, operation in _operations(keys):
+        row = {
+            "operation": label,
+            "sqlite": _measure(sqlite_backend, operation, counter, "sql"),
+        }
+        row["api"] = (
+            _measure(api_backend, operation, counter, "http")
+            if api_backend is not None
+            else {"unavailable": "Zotero API not reachable"}
+        )
+        rows.append(row)
+
+    result = {
+        "library_items": reader.get_item_count(),
+        "keys_measured": len(keys),
+        "api_available": api_backend is not None,
+        "operations": rows,
+    }
+    reader.close()
+    return result
+
+
+def _cell(measurement, unit):
+    if "error" in measurement:
+        return "error", "—"
+    if "unavailable" in measurement:
+        return "—", "—"
+    return f"{measurement['calls']} {unit}", f"{measurement['ms']:,.1f} ms"
+
+
+def render(result):
+    lines = []
+    add = lines.append
+
+    add(f"Library: {result['library_items']:,} items · "
+        f"batch size: {result['keys_measured']} keys")
+    if not result["api_available"]:
+        add("")
+        add("Zotero API not reachable — SQLite column only. Every operation below")
+        add("is answered with no network access at all, which is the point: these")
+        add("reads used to be impossible with Zotero desktop closed.")
+    add("")
+
+    header = f"{'operation':<26}{'sqlite':>10}{'':>13}{'api':>12}{'':>15}"
+    add(header)
+    add(f"{'':<26}{'queries':>10}{'time':>13}{'requests':>12}{'time':>15}")
+    add("-" * len(header))
+
+    for row in result["operations"]:
+        s_calls, s_ms = _cell(row["sqlite"], "")
+        a_calls, a_ms = _cell(row["api"], "")
+        add(f"{row['operation']:<26}{s_calls:>10}{s_ms:>13}{a_calls:>12}{a_ms:>15}")
+
+    if result["api_available"]:
+        add("")
+        add("The sqlite column is flat in batch size; the api column is not.")
+    return "\n".join(lines)
+
+
+def main():
+    parser = argparse.ArgumentParser(
+        description=__doc__, formatter_class=argparse.RawDescriptionHelpFormatter
+    )
+    parser.add_argument("--json", action="store_true",
+                        help="emit machine-readable JSON instead of a table")
+    parser.add_argument("--keys", type=int, default=DEFAULT_KEYS,
+                        help=f"how many item keys to batch (default {DEFAULT_KEYS})")
+    args = parser.parse_args()
+
+    result = collect(max(1, args.keys))
+    print(json.dumps(result, indent=2) if args.json else render(result))
+
+
+if __name__ == "__main__":
+    main()
--- a/src/zotero_mcp/cli_standalone.py
+++ b/src/zotero_mcp/cli_standalone.py
@@ -132,32 +132,26 @@ def _keys_from_markdown(markdown: str) -> list[str]:
     return keys
 
 
-def _fetch_projected(zot, keys: list[str], detail: str = "summary") -> list[dict]:
+def _fetch_projected(backend, keys: list[str], detail: str = "summary") -> list[dict]:
     """Fetch *keys* and project them, preserving the order they were given in.
 
     Order carries meaning -- relevance for a search, recency for `get recent`
-    -- so it is restored explicitly rather than left to whatever the API
+    -- so it is restored explicitly rather than left to whatever the backend
     returns. Keys the fetch does not return (deleted between the two calls,
-    or not visible to this client) are dropped rather than faked.
+    or not visible here) are dropped rather than faked. Batching is the
+    backend's business now, so there is no chunk loop left here.
     """
     if not keys:
         return []
-    found: dict[str, dict] = {}
-    # itemKey takes up to 50 per request.
-    for i in range(0, len(keys), 50):
-        chunk = keys[i:i + 50]
-        try:
-            batch = zot.items(itemKey=",".join(chunk), limit=len(chunk))
-        except Exception:
-            batch = []
-        for item in batch or []:
-            if isinstance(item, dict) and item.get("key"):
-                found[item["key"]] = item
+    try:
+        found = backend.get_items(keys)
+    except Exception:
+        found = {}
     return [_cli_json.project_item(found[k], detail) for k in keys if k in found]
 
 
-def _zot(args):
-    """The Zotero client, for the JSON paths that project raw records.
+def _read_backend():
+    """The read backend, for the JSON paths that project raw records.
 
     Structured output reads the same records the markdown formatters read;
     it just skips the formatting. Nothing about *which* records to fetch is
@@ -165,8 +159,9 @@ def _zot(args):
     variants, the semantic cascade), the JSON path calls that same logic and
     projects its result.
     """
-    _search, _retrieval, _annotations, _write, _client = _import_tools()
-    return _client.get_zotero_client()
+    from zotero_mcp import library as _library
+
+    return _library.get_library_backend()
 
 
 # ---------------------------------------------------------------------------
@@ -237,7 +232,7 @@ def cmd_search(args):
 
     if _json_mode(args):
         keys = _keys_from_markdown(result)
-        items = _fetch_projected(_client.get_zotero_client(), keys,
+        items = _fetch_projected(_read_backend(), keys,
                                  getattr(args, "detail", "summary"))
         _out(args, "search", data={
             "query": args.query, "mode": args.mode,
@@ -284,8 +279,7 @@ def cmd_get(args):
              text=result)
     elif sub == "collections":
         if json_mode:
-            from zotero_mcp.utils import _paginate
-            cols = _paginate(_client.get_zotero_client().collections, max_items=args.limit)
+            cols = _read_backend().list_collections()[:args.limit]
             _cli_json.emit("get collections", {
                 "count": len(cols),
                 "collections": [_cli_json.project_collection(c) for c in cols],
@@ -303,7 +297,7 @@ def cmd_get(args):
                 "collection_key": args.collection_key,
                 "offset": getattr(args, "offset", 0),
                 "count": len(keys),
-                "items": _fetch_projected(_client.get_zotero_client(), keys, args.detail),
+                "items": _fetch_projected(_read_backend(), keys, args.detail),
             })
             return
         print(result)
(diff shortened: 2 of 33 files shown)


mronkko:
Now, the human written explanation of why this is needed: The main problem is that when working with large libraries, Zotero MCP becomes the bottleneck. This is particularly when using Google's Antigravity and Gemini Flash models, which are super fast. Now the agent spends a bulk of the clock time waiting for the MCP server to answer. The culprit is the API overhead and using repeated API calls for retrieving something that could be a single SQL query. This problem is compounded if subagents hit the API-based retrieval at the same time, forcing a wait for the API to become available. The solution is to expand the SQLite search backend to be a full backend that also serves item data etc without hitting the Zotero API. This is and should always remain a read-only route to avoid corrupting the data.

The PR is a proof of concept. It implements the read tools without any API calls. I developed this by first implementing a set of live tests that block both APIs and adjuted the code until these tests no longer failed. The outcome is between 1-3 orders of magnitude faster tool call responses compared to the API route. In multiagent systems, the performance gains are even more better because the sqlite database allows parallel reads without blocking.

Please let me know if this is something that might be considered and I can implement it in a set of staged PRs that are easier to evaluate and merge.

mronkko:
I have been using my private build with this patch for a couple of weeks now. It makes a huge difference in usability when combined with an efficient model like Gemini Flash 3.8 in Antigravity. I can ask the agent questions about my Zotero data and get a near-chatbot-like experience compared to the deep-research-like experience with the current distribution.

54yyyu:
@mronkko we want this in. The numbers make the case, and the parity suite plus the network-blocked suite are what make it reviewable at this size.

Rather than have you split it into seven PRs, I'm going to integrate it myself on a branch off `maint/0.12.1`, keeping your commits and authorship. Plan:

- Rebase onto 0.12.x. A lot moved under you: #472 (Zotero 10 local writes), #536 (local reads now include the WAL via a snapshot copy), and a few read paths you touched.
- Drop the Finder duplicates that came along (`html_metadata 2.py`, `README 3.md`, and so on, 11 files).
- Ship it opt-in (`ZOTERO_BACKEND=sqlite`) in 0.13.0, with `ApiBackend` as the default for one release.
- One thing I need to fix before it can be the fast path: since #536, a read made while Zotero's WAL holds changes goes through a copy of `zotero.sqlite` plus its WAL, refreshed when either changes. On a 2.3 GB library that is expensive if Zotero writes often, so the refresh needs to be bounded.
- Wire up or drop `UnsupportedByBackend`, per your known gaps.

I'll ping you when the branch is up. Once it is, running your `measure_read_backend.py` and the parity suite against your library would be the most useful check, since mine is much smaller. If you'd rather keep driving it yourself, say so and I'll get out of the way.

54yyyu:
@mronkko the integration branch is up: https://github.com/54yyyu/zotero-mcp/pull/550 (`feat/sqlite-read-backend`). Your commits are kept as-is under a merge commit, and the PR description lists what changed on top. When you have a moment, running `scripts/measure_read_backend.py` and `tests/live/test_read_backend_parity.py` against your 2.3 GB library on that branch would tell us whether the reader-refresh change costs anything at your scale.

mronkko:
(zotero-mcp-server) mronkko@Mikko-Ronkko-Home-Mac-4 zotero-mcp %  python scripts/measure_read_backend.py
Library: 44,105 items · batch size: 25 keys

**Library:** 44,105 items · **batch size:** 25 keys

| operation | sqlite queries | sqlite time | api requests | api time |
|---|---:|---:|---:|---:|
| `get_item(1 key)` | 7 | 0.4 ms | 1 | 43.5 ms |
| `get_items(25 keys)` | 7 | 4.1 ms | 1 | 140.5 ms |
| `get_children(25 keys)` | 9 | 1.8 ms | 25 | 2,476.3 ms |
| `list_collections()` | 2 | 1.2 ms | 3 | 45.7 ms |
| `list_tags()` | 2 | 110.3 ms | 181 | 103,903.7 ms |
| `recent_items(25)` | 7 | 558.0 ms | 1 | 268.9 ms |

54yyyu:
Thanks for running it, those numbers settled it. Two updates on the branch (#550):

- **`recent_items` is fixed.** It was the one row where SQLite lost (558 ms vs 269 ms). The query sorted the whole hydration join, with the attachment, note, annotation and parent LEFT JOINs for every item, before applying LIMIT. It now ranks bare item ids in one pass over `items` and hydrates only the page. Results are identical for every sort field and direction on my library.
- **SQLite becomes the default in local mode, shipping in 0.12.1.** A `FallbackBackend` re-asks the API for any call that raises `UnsupportedByBackend`, which also closes that known gap, and `ZOTERO_BACKEND=api` opts out. Readers kept across tool calls now reopen when Zotero writes, and the WAL snapshot of a large database is refreshed at most once per `ZOTERO_MCP_DB_SNAPSHOT_MIN_INTERVAL` seconds (default 5).

If you can rerun `scripts/measure_read_backend.py` on the branch, the `recent_items` row and anything that looks slower than on your original branch are what I'd most like to see at 44k items.

54yyyu:
@mronkko this shipped in 0.12.1 today, as the default read path in local mode with the API as a per-call fallback (`ZOTERO_BACKEND=api` opts out). Your 44k-item numbers are in the release notes and the README, credited to you. Thank you for building it and for measuring it properly. If anything looks off on your library after upgrading, open an issue and tag me.

mronkko:
There are at least two bugs. One is pre-existing date filtering bug and the other is in the tag list. 

I posted this as a RFC initially because I was not sure if it fit the plans for the overall architecture of the project. But thans for the quick merge! A 100 fold increase in speed makes a big difference.

mronkko:
#552 is new and #551 existed in other backends too.

54yyyu:
Thank you so much! Both issues are clear and the analysis is excellent, as usual.

- **#552:** I'll make SQLite match the API, so trashed-only tags are listed again and `get_tags` stays a pure speed-up. I'll also give that parity test its own timeout so it can reach the assertion on a large library.
- **#551:** agreed on all of it. I'll compare date ranges on `meta.parsedDate` padded to `YYYY-MM-DD`, drop the lexical fallback for dates, make dateless items match nothing, and add `date` range cases to the parity suite.

Both will go into 0.12.2.

mronkko:
|operation|sqlite queries|time|api requests|time|
|---|---:|---:|---:|---:|
|get_item(1 key)|7|0.6 ms|1|2.7 ms|
|get_items(25 keys)|7|1.2 ms|1|118.3 ms|
|get_children(25 keys)|9|0.7 ms|25|1,340.8 ms|
|list_collections()|2|1.0 ms|3|47.1 ms|
|list_tags()|2|116.2 ms|181|98,236.3 ms|
|recent_items(25)|8|299.5 ms|1|280.7 ms|

54yyyu:
Thank you so much! These numbers are great to see, and thank you for rerunning on the branch.

`recent_items` is now on par with the API instead of twice as slow. On a small library it runs in about 9 ms over 9 queries, so the remaining time at 44k items is probably the date sort (Zotero's schema has no index on `items.dateAdded`, so SQLite sorts the whole library for the top 25) or a WAL snapshot refresh landing while Zotero was writing. If you have a minute, a `cProfile` of `recent_items` on your library would tell us which.

0.12.2 is out with both of your fixes (#551, #552).

mronkko:
I profiled it on 0.12.2 (`fa51adb`), same library. The personal library is 52,183 of the 90,328 `items` rows, and `zotero.sqlite` is 504 MB. M1 Max, Python 3.11.11, SQLite 3.47.1, Zotero running, WAL empty for every run.

**It's neither the sort nor a snapshot refresh. The time goes into how the ranking query walks the index.**

`recent_items(25)` takes 158 ms on the first call on a new connection, and a median of 96 ms over 20 warm calls. Almost all of that is one statement:

| statement | rows | warm median | share |
|---|---:|---:|---:|
| ranking: `SELECT i.itemID FROM items i JOIN itemTypes … ORDER BY i.dateAdded DESC, i.itemID DESC LIMIT ?` | 25 | 95.4 ms | 99% |
| the other 8 (scope, page, fields, creators, tags, collections, relations, library labels) | | under 1 ms together | 1% |

Python outside SQL is 0.2 ms. In cProfile, `sqlite3.Connection.execute` accounts for 2.054 s of 2.081 s across the 20 warm calls.

- **The sort costs nothing measurable.** 28,402 rows pass the filters. The same scan and filters without `ORDER BY … LIMIT`, run as a `COUNT(*)` with the same plan minus the temp B-tree, take 98.0 ms, against 96.3 ms with the sort.
- **A snapshot refresh can't explain it.** `measure_read_backend.py` opens its connection before timing starts and builds `SqliteBackend` directly, so `refresh_if_stale` never runs inside a timed call. When a refresh does happen on the server path, it costs 849 ms to copy this 504 MB database, plus 101 ms for the first `recent_items` on the copy. That would show up as about a second.
- **The cost is the plan.** `SEARCH i USING INDEX sqlite_autoindex_items_1 (libraryID=?)` walks all 52k personal-library entries in `(libraryID, key)` order, and fetches each row from the table by rowid to read `itemTypeID` and `dateAdded`. A sequential scan of the whole table is cheaper:

| variant (all return the same 25 ids; 30 interleaved runs) | median | min |
|---|---:|---:|
| as written | 93.9 ms | 89.0 ms |
| `FROM items i NOT INDEXED` | 36.8 ms | 34.9 ms |
| `i.itemTypeID NOT IN (SELECT itemTypeID FROM itemTypes WHERE typeName IN (…))` instead of the `itemTypes` join | 85.1 ms | 80.9 ms |
| both | 31.1 ms | 29.9 ms |

I only checked these on the personal library, sorted by `dateAdded` descending. The index should still win when the scope is a small share of `items`, such as a small group library in a large database, so a blanket `NOT INDEXED` could make those scopes slower. I haven't measured that.

**About the 299.5 ms I posted earlier:** I can't reproduce it. A full `measure_read_backend.py` run today gives `recent_items(25)` 91.7 ms on SQLite against 257.0 ms on the API, and three SQLite-only runs gave 93.9, 94.0 and 98.3 ms. The WAL was empty in all of today's runs, so every read went to the file in place. I didn't record the WAL state during the earlier run.

One more observation: through `get_library_backend()`, the path the tools use, warm calls take a median of 111 ms. That is about 15 ms more than calling `SqliteBackend` directly, and I haven't profiled where it goes.

<details><summary>cProfile, 20 warm calls, sorted by tottime</summary>

```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
      160    2.054    0.013    2.054    0.013 {method 'execute' of 'sqlite3.Connection' objects}
      140    0.018    0.000    0.018    0.000 {method 'fetchall' of 'sqlite3.Cursor' objects}
       20    0.002    0.000    0.016    0.001 local_db.py:2507(_fetch_fields)
      500    0.002    0.000    0.003    0.000 local_db.py:774(_row_to_full_api_item)
       20    0.001    0.000    0.004    0.000 local_db.py:2220(_fetch_creators)
       20    0.001    0.000    2.081    0.104 local_db.py:2841(get_recent_items)
       20    0.001    0.000    0.003    0.000 local_db.py:2608(<listcomp>)
       20    0.000    0.000    0.026    0.001 local_db.py:2593(_build_full_items)
```

</details>

<details><summary>EXPLAIN QUERY PLAN, as written and fastest variant</summary>

```
as written
     SEARCH i USING INDEX sqlite_autoindex_items_1 (libraryID=?)
     USING ROWID SEARCH ON TABLE deletedItems FOR IN-OPERATOR
     SEARCH it USING INTEGER PRIMARY KEY (rowid=?)
     USE TEMP B-TREE FOR ORDER BY

NOT INDEXED + itemTypeID filter
     SCAN i
     USING ROWID SEARCH ON TABLE deletedItems FOR IN-OPERATOR
     LIST SUBQUERY 2
     SCAN itemTypes
     CREATE BLOOM FILTER
     USE TEMP B-TREE FOR ORDER BY
```

</details>

I can share the profiling script if that helps.

54yyyu:
This is really useful, thanks for digging into it. So it's the index walk, not the sort or the snapshot. That makes sense given how much of `items` your personal library is.

I'll switch the type filter to the `itemTypeID` subquery, since that shouldn't hurt anywhere. For the scan, I'll check how much of `items` the scope covers and only skip the index when it's most of the table, so a small group library still uses it. I'll also look at the extra ~15 ms through `get_library_backend()`.

Could you share the profiling script? I'd like to add it to `scripts/` alongside `measure_read_backend.py`.

PR #504, opened 2026-08-30
fix: parse the detailed children listing's - Key: line in _keys_from_markdown

AmirF194:
## Root cause

`zotero-cli --json get children <parent_key>` reports `count: 0` even when
markdown mode lists real children. The JSON path extracts item keys from
the tool's own markdown output with `_keys_from_markdown()`, whose regex
recognizes `**Item Key:** KEY` and the keys-only `` - `KEY` `` form, but
`_format_children_detailed()` (the single-parent path `get_item_children`
takes) writes each child's key on its own indented `   - Key: KEY` line, a
third shape the regex never matched. The fetch step then runs on an empty
key list and returns nothing.

## Fix

Add a third alternative to `_ITEM_KEY_RE` for the indented `- Key: KEY`
form and read it into the extracted key list alongside the other two.

## Verification

- New regression test (`test_reads_keys_from_a_detailed_children_listing`)
  builds a fake parent with one note and one attachment child, renders it
  through the real `_format_children_detailed()`, and asserts
  `_keys_from_markdown()` recovers both keys. Confirmed it fails on `main`
  (`count: 0`) and passes on this branch.
- Full suite (`pytest --ignore=tests/test_lifespan.py`, Python 3.11):
  2709 passed, 42 skipped, same 7 pre-existing failures as on unmodified
  `main` (rate-limit-guard and webdav tests that depend on network/timing,
  unrelated to this change).
- Checked every other place in the codebase that emits an indented
  `- Key:` line: only the three call sites inside
  `_format_children_detailed` (attachments, notes, other items) do, so no
  other markdown shape is affected by the new alternative.
- Not verified: the batch form (`get_item_children` called with several
  keys) renders through `_format_children_grouped`, which uses a different
  `- [KEY] ...` line shape that this regex still does not match, so
  `--json get children` with multiple keys likely has the same symptom.
  That path is not part of this issue's reproduction and is out of scope
  here.

### DIFF
--- a/src/zotero_mcp/cli_standalone.py
+++ b/src/zotero_mcp/cli_standalone.py
@@ -114,18 +114,21 @@ def _keys_from_markdown(markdown: str) -> list[str]:
     ranking, the collection scoping -- instead of reimplementing it and
     drifting from what markdown mode returns. The two modes therefore always
     agree on *which* items matched; JSON only changes how they are rendered.
+    A detailed children listing writes a third shape, `   - Key: KEY`.
     """
     global _ITEM_KEY_RE
     if _ITEM_KEY_RE is None:
         import re
         _ITEM_KEY_RE = re.compile(
-            r"^\*\*Item Key:\*\*\s*`?([A-Z0-9]{8})`?\s*$|^- `([A-Z0-9]{8})`",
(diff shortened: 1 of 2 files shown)


AmirF194:
The five red CI legs here are the test suite hitting the live Zotero API and getting rate limited with 429s, not anything from this diff, in case that's confusing at a glance.

PR #506, opened 2026-08-31
fix(dedup): if_exists='file' misses existing items, because identifiers aren't server-side searchable

danmackinlay:
We hit repeated duplicate items when adding papers that were already in the library, and this is what we think is going on. We're outside contributors and new to this codebase, so please treat the diagnosis as a proposal rather than a conclusion — there may well be context or constraints we've missed, and we'd rather be corrected than merged.

**Updated after self-review.** Two claims in the first version of this description were wrong, and one measurement we'd guessed at turned out the other way. Both are corrected below and marked ~~struck~~ where the original said otherwise.

## What we think is happening

`if_exists='file'` (the CLI default) is meant to reuse an item already in the library. In our testing it doesn't, and creates a duplicate instead. Our reading is that the search it relies on can't answer the question being asked of it.

`find_existing_items` narrows candidates with a search on the identifier itself:

```python
candidates = zot.items(q=query, qmode="everything", itemType="-attachment", limit=50)
```

Per the [Web API docs](https://www.zotero.org/support/dev/web_api/v3/basics), `q` "Searches titles and individual creator fields by default", and `qmode` only adds full-text content — with the note that **"Searching of other fields will be possible in the future."** That reads to us as saying `DOI`, `url`, `archiveID` and `extra` aren't searchable server-side today.

If that's right, the query returns zero candidates for an item that *is* present, the client-side comparison never runs, the caller concludes "not in the library", and a duplicate is created.

### What we measured (one real library, 16,859 top-level items, Web API)

Six items were selected *for having a DOI*, then searched by their own DOI:

| Item | DOI | hits |
|---|---|---|
| F7GQZKCV | `10.1561/2200000100` | 0 |
| QDCLI6FI | `10.48550/arXiv.2504.13837` | 0 |
| 6W268K77 | `10.48550/arXiv.2504.07086` | 0 |
| 2B7KNWCG | `10.48550/arXiv.2210.03493` | 0 |
| GWE24YPT | `10.48550/arXiv.2201.11903` | 0 |
| CMKZQQIG | `10.1038/s41586-025-09422-z` | 0 |

Six of six. Searching `archiveID` fares no better — `q='arXiv:2504.13837'` also returns 0 for an item whose `archiveID` is exactly that. For a preprint that is in the library, no spelling of its arXiv ID found it, while a title query found it immediately:

```
q='2509.04259'                mode=everything       -> 0 hits
q='arXiv:2509.04259'          mode=everything       -> 0 hits
q='10.48550/arXiv.2509.04259' mode=everything       -> 0 hits
q="RL's Razor"                mode=titleCreatorYear -> 1 hit  ['VLYF39QV']
```

The current tests wouldn't catch this: `FakeZotero.items()` ignores `q` and returns everything, so the client-side matcher is exercised but the search never is. That's worth noting mainly because it means "tests pass" was never evidence against this.

### How quick search actually matches

Since the fix leans on the title query, we went and measured what that query does, against a real item titled `The Invisible Leash: Why RLVR May or May Not Escape Its Origin`:

| query | hits |
|---|---|
| the exact title | 1 |
| the same words in reverse order | 1 |
| the title with runs of spaces injected mid-string | 1 |
| the title, lowercased | 1 |
| the title + one junk word | **0** |
| the title with one `<i>…</i>` pair | **0** |
| the title with `&amp;` in place of `&` | **0** |

So quick search splits the query on whitespace and requires **every** token to match. Two consequences, one in each direction:

- Whitespace doesn't matter. ~~We initially assumed arXiv's line-wrapped Atom titles — which `_add_by_arxiv` turns into runs of spaces — would break the query.~~ They don't; the tokenizer eats them.
- Markup and entities matter a great deal. CrossRef ships JATS in `title[0]`, and a literal `<i>` or `&amp;` is a token that matches nothing. One italicised species name takes the whole fallback to zero against an item whose stored title is clean — which is exactly the item the fallback exists to find.

## What this PR changes

Six commits, all narrow.

**1. Compare arXiv identity across the forms Zotero actually stores.** The arXiv matcher checked two of the four places an arXiv identity can live, and compared version-sensitively:

```
url        https://arxiv.org/abs/2509.04259     (arXiv add)      checked
archiveID  arXiv:2509.04259                     (connector)      NOT checked
DOI        10.48550/arXiv.2509.04259            (DOI add)        NOT checked
extra      arXiv:2509.04259 [cs]                (translator)     checked
```

So an item saved by the browser connector, or added by DOI, looked absent to a later arXiv add of the same paper. Separately `.../abs/2509.04259v1` and a stored `.../abs/2509.04259` compared unequal.

`_normalize_arxiv_id` keeps the version deliberately — its other callers fetch a specific version — so rather than change that contract we added `_arxiv_identity` for the version-independent form, which also accepts arXiv's DataCite DOI. Fetch behaviour is untouched.

**2. Fall back to the one field the API does index.** `find_existing_items` gains an optional `title=`. When the identifier query finds nothing *and* a title was supplied, it queries the title with `qmode='titleCreatorYear'` and runs *the same identifier comparison* over those candidates.

The identifier still decides, so this widens the net without loosening the test — a same-title-different-paper is rejected exactly as before.

**3. Normalize the title before it becomes a query.** Tags stripped, entities resolved, whitespace collapsed. Order matters: tags first, so an escaped `&lt;i&gt;` — literal text in a title — survives rather than being unescaped into a tag and then eaten. Without this, the fallback in commit 2 silently returns nothing for any CrossRef record with markup in its title.

**4. Wire it into all four call sites that hold metadata by the time they check.** ~~We only wired the two call sites where a title was already in hand.~~ There were four, not two:

| call site | title source |
|---|---|
| `_add_by_arxiv`, post-fetch check | arXiv Atom `title` |
| `add_by_doi`, phase 3b | CrossRef `item_data['title']` |
| `add_by_isbn`, second check | Open Library / Google Books |
| `_maybe_reuse_existing` | the parsed BibTeX/CSL-JSON entry |

The batch-import one is the one that matters most, and we missed it first time round: a BibTeX or CSL-JSON import carries hundreds of entries per call and dedups each on its DOI alone. Its docstring's "title matching is out of scope (#4)" still holds — the title is a search key, the DOI still decides. The ISBN site is the second of that path's two checks, i.e. the one holding the identifier lock and actually guarding the create, so leaving it blind meant a book the first check found could still be shelved twice.

**5. Widen the candidate window to 100, the API maximum.** `limit=50` never bound while the only query was on an identifier the API cannot search — that query returns nothing, so the window was never full. The title query fills it, which makes the window load-bearing for the first time, so this is the new code's own edge case rather than a pre-existing one we happened to notice.

Three things compound. Quick search matches each token as a **substring**, so a short title pulls in far more than it appears to. Results are sorted `dateModified` descending. And the item being deduped against is *by definition already in the library*, so it is competing for the window with everything modified since — the ordering runs against the caller.

We swept the library for it. Of 154 items with a title of three words or fewer (7.7% of titles), three saturated a 50-item window, and for one the item itself sorted outside it:

```
q='Dependence'            91 matches   target at rank  2  -> found
q='Noise'                 87 matches   target at rank ..  -> found
q='Stochastic Processes'  83 matches   target OUTSIDE 50  -> *** would duplicate ***
```

`35YSBXVH` is a book, last modified 2024 — exactly the old-item profile the sort order penalises. All three fit under 100, so raising the cap is the whole fix rather than a mitigation. Full titles are not affected at all: of 45 sampled, 43 returned exactly one candidate and none came close to the cap.

**We deliberately did not add creator refinement.** Appending the first author's surname is far more selective — `'Stochastic Processes Ross'` returns 1 candidate against 50 — and quick search does AND tokens across title *and* creator fields, so it works. But it guards a failure we have never observed (>100 candidates) at the cost of one we cannot enumerate from a single library: a token can only narrow, so any disagreement between the stored creator and the fetched one silently loses the match — transliterations, `van der Berg` vs `Berg`, corporate authors, items with no creators at all. A 39-item sample showed zero regressions, but that library's items were largely added by the same tooling, so stored and fetched creators agree there by construction. That is no evidence of harm, not evidence of no harm. Happy to add it if you'd rather have the belt and braces.

**6. CHANGELOG entry** under `## [Unreleased]`, matching the convention every other bugfix here follows.

### What it costs

~~The second query only runs when the first found nothing, so the common path is unchanged.~~ This was wrong, and wrong in the direction that flatters the patch. Because the identifier query almost never matches against the Web API — that's the whole premise — "only on a miss" means "on nearly every call". A caller that passes a title should expect **two searches per check, not one**. For `add_by_doi`, a 50-DOI batch goes from 100 dedup reads to as many as 150.

We kept the identifier query in front of the fallback anyway, and the docstring now says why: `qmode='everything'` also searches child-attachment full text, where a paper's own DOI genuinely does appear, and it finds an item stored under a title that no longer matches the one just fetched. Neither is something the title query can do. But that's a coverage argument, not the rarity argument we originally made, and if you'd rather have the single query we'd happily skip the identifier pass when a title is available — `_confirm` still decides either way, so correctness is identical.

## Verification

Read-only, against the same library, before and after:

```
                                        before      after
'2509.04259'                            []          ['VLYF39QV']
'2509.04259v1'                          []          ['VLYF39QV']
'arXiv:2509.04259'                      []          ['VLYF39QV']
'https://arxiv.org/abs/2509.04259v3'    []          ['VLYF39QV']
'10.48550/arXiv.2509.04259'             []          ['VLYF39QV']

negative control (same title, wrong ID)             []
```

Five further items, searched by their own DOI with their own title supplied — all five previously returned `[]`, all five now resolve to themselves; a real title paired with `10.9999/does.not.exist` still returns `[]`.

We also ran it end to end through the tool against a scratch collection. Seeding an item shaped the way the **browser connector** stores one — identity only in `archiveID` and the DataCite DOI, `url` and `extra` empty — a later `add_by_url` of the same paper reused it instead of duplicating, versioned spellings included:

```
seeded ZT4JWRXS: archiveID='arXiv:1207.7214', DOI='10.48550/arXiv.1207.7214', url='', extra=''

  find_existing_items(arxiv_id='1207.7214')                   -> []            (pre-patch and post-patch, no title)
  find_existing_items(arxiv_id='1207.7214',   title=...)      -> ['ZT4JWRXS']
  find_existing_items(arxiv_id='1207.7214v3', title=...)      -> ['ZT4JWRXS']

  add_by_url('https://arxiv.org/abs/1207.7214', if_exists='file')
      -> Already in library: ... (`ZT4JWRXS`, matched by arXiv ID 1207.7214)
```

Read-only, the case the sweep found:

```
35YSBXVH  'Stochastic Processes'  book, ISBN 978-81-265-1757-2, dateModified 2024-11-06

  limit=50   -> 50 candidates, target NOT in window
  limit=100  -> 83 candidates, target in window
  find_existing_items(isbn=..., title='Stochastic Processes') -> ['35YSBXVH']
  same title, wrong ISBN                                      -> []
```

25 new tests. Each one fails without its corresponding fix, checked by reverting the source files and re-running; every commit is green on its own. Full suite 2,749 passed, 42 skipped. `ruff check` clean on both changed source files (the four hits in the test files are pre-existing on `3cb3e2e`).

The 5 `test_rate_limit_guard.py` failures are pre-existing on an untouched `3cb3e2e` and unrelated to this branch — filed as #511, with the runtime half as #512. Correcting an earlier version of this section, which blamed a `whenever` `to_stdlib`/`py_datetime` rename: that is not what CI is failing on. The CI log contains no `to_stdlib` error, and the failures there are `httpx.HTTPStatusError` leaking out of pyzotero.

The actual cause is that pyzotero 1.15.x moved from `httpx` to `httpx2`, which are disjoint class hierarchies. The test module builds fixtures with `import httpx` (0.28.1, arriving via `mcp`/`httpx-sse`), while pyzotero's `_post_check` catches only `httpx2.HTTPError` — so the error escapes uncaught, no retry happens, and `TooManyRetriesError` is never raised. `pyproject.toml` has no ceiling on `pyzotero` and CI installs it bare, so this began the moment 1.15 was released; main's last green run was Aug 25, before this branch existed.

Independent of this PR either way, and we haven't touched it.

## Where we could be wrong

- **We only tested the Web API.** `ZOTERO_LOCAL` goes through a different backend. It serves the same `q`/`qmode` contract, so we'd expect identical behaviour, but we haven't confirmed it — and if local-API search *does* index identifiers, the fallback is unnecessary work on that path and should be conditional.
- **One library, one point in time.** Everything above is measured on a single large personal library. We haven't checked group libraries, a different server version, or a library with full-text indexing in a different state.
- **Recall is only as good as the title.** The fallback finds an item whose stored title matches the one just fetched. A paper filed under its published title and re-added from its preprint title still won't match. The identifier comparison guarantees no false positives, but false negatives remain.
- **We assumed the current behaviour is unintended.** If `find_existing_items` was always meant to be best-effort and cheap, with duplicates accepted as the cost, then this changes a deliberate trade-off rather than fixing a bug.

On the "wrong shape of fix" worry from the original description: we think the answer is that this is the right fix *for the Web API path*, but it isn't the best available one everywhere. `local_db.py` already joins `itemData`/`itemDataValues`/`fields` and pulls DOI and url per item, so a `ZOTERO_SEARCH_BACKEND=sqlite` user could get exact identifier dedup in a single indexed query with no false negatives at all. That would be a natural layer above this rather than a replacement for it, and we'd be glad to attempt it as a follow-up if you think it's worth having.

## Not addressed here

Adjacent things noticed while investigating, left out to keep this reviewable — we can open separate issues or PRs, whichever you prefer:

- `find_duplicates` refuses libraries over 5,000 items, but only discovers that after paging in 5,100 of them: 163 seconds to return "too large for duplicate scan" on the library above. `zot.num_items()` would answer in one call. It also pages with `zot.items()`, which includes attachments and notes (36.7% of a sampled 1,000) that are then discarded.
- The plain-`url` dedup path still has the unsearchable-identifier problem. We didn't wire it because no title is in hand at either call site. (The `isbn` path *was* in that list in the original description and is now fixed — that was our error.)
- `_add_by_arxiv`'s first dedup check, and `add_by_doi`'s phase-1 check, both run before any metadata is fetched and so can never supply a title. Given the diagnosis above they can essentially never match against the Web API, which makes them a round trip that always misses. We left them alone because they're load-bearing for the concurrency narrowing (#486) and removing them is a bigger call than this PR should make.
- `_add_by_arxiv` stores arXiv's wrapped title with the wrap left in — `.strip().replace("\n", " ")` turns the newline into a space but leaves the indent, so long titles are stored containing runs of spaces. Harmless for search (see the measurements above) but visible in the item pane. One-line fix, deliberately not smuggled into this PR.
- `add` reports the item key but not the BetterBibTeX citation key, which is what ends up in a document. Awkward because BBT assigns keys asynchronously for newly-created items, so it may only be cheap for the already-in-library path.

Happy to split, rework, or drop any of it.

---

## Update, 14 September 2026

Rebased onto `maint/0.12.1` (`c1abf12`), since 0.12.0 shipped with #544. The details are in [this comment](https://github.com/54yyyu/zotero-mcp/pull/506#issuecomment-5658921295). Against the description above:

- Eight commits now, not six. `390c699` fixes two bugs in the title cleaner: tags inside a word, and angle brackets that #523 decodes before the cleaner runs. `8c4df31` passes the page's title to the two checks #515 added. That covers pages with embedded metadata from the "plain-`url` dedup path" item under *Not addressed here*; a plain webpage with no usable metadata still has no title to pass.
- The `test_rate_limit_guard.py` failures described under *Verification* were fixed on maint by #513 and #514.
- Current numbers: every commit passes the full suite on its own (3140 passed, 48 skipped at the tip), and ruff reports nothing beyond what is already on maint.
- End-to-end adds against a live library are summarised in the comment.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **`zotero_get_item_fulltext` says when a document was cut short.** Reads stop at `fulltext_display_max_pages` (10 by default), but the text returned for the first ten pages of a 442-page book was byte-for-byte indistinguishable from the text returned for a complete three-page note: same `## Full Text` heading, no marker anywhere, and a tool description that promised "the entire paper". An agent summarized and cited the excerpt as if it had read the whole thing. The parser already reports `page_count` and `truncated` on every extraction, so both capped paths — local storage and the last-resort download+parse — now carry them through to the heading: `## Full Text (pages 1-10 of 442 — TRUNCATED)`, followed by the config key to raise and the `zotero_read_pdf_pages` call that resumes at the first unread page. A read that fits under the cap is unchanged. Zotero's own `.zotero-ft-cache` and the transient fulltext cache are not page-capped and are never reported as truncated. `LocalZoteroReader`'s parser seam returns the whole `ExtractedDoc` instead of just its text, so page bookkeeping has one source of truth rather than being re-derived by callers (#448).
+
 ## [0.11.0] - 2026-08-25
 
 **Upgrading:** `zotero_semantic_search` now defaults to the active library instead of every indexed library. If you relied on the old implicit behaviour, pass `search_all_libraries=True`.
--- a/src/zotero_mcp/local_db.py
+++ b/src/zotero_mcp/local_db.py
@@ -22,6 +22,7 @@ from . import fulltext_cache
 from . import search_semantics as _semantics
 from .config import load_config
 from .extract import (
+    ExtractedDoc,
     categorize_attachment,
     extract_file,
     normalize_attachment_priority,
@@ -65,6 +66,23 @@ class KeyGroupMap(NamedTuple):
 
     groups: dict[str, int]
     excluded_keys: set[str]
+
+
+@dataclass
+class FulltextExtraction:
+    """Extracted attachment text, plus how much of the document it covers.
+
+    ``page_count`` is the document's real length and ``truncated`` says the
+    page cap dropped pages from the tail — both straight from
+    :class:`~zotero_mcp.extract.ExtractedDoc`. They carry their defaults on
+    the cache paths, which are not page-capped and keep no page bookkeeping
+    (#448).
+    """
+
+    text: str
+    source: str
+    page_count: int | None = None
+    truncated: bool = False
 
 
 def _read_string_pref(prefs_path: Path, pref: str) -> str | None:
@@ -805,10 +823,15 @@ class LocalZoteroReader:
         except ValueError:
             return DEFAULT_PDF_MAX_PAGES
 
-    def _extract_text_from_file(self, file_path: Path) -> str:
-        """Extract text from an attachment file, or "" if nothing readable."""
-        doc = extract_file(file_path, max_pages=self._resolve_pdf_max_pages())
-        return doc.text if doc else ""
+    def _extract_doc_from_file(self, file_path: Path) -> ExtractedDoc | None:
+        """Parse an attachment file, or None if nothing readable came back.
+
+        Returns the whole document rather than its text: how long the source
+        is and whether the page cap cut it short are the parser's to report,
+        and a caller that has to re-derive them gets a second source of truth
+        (#448).
+        """
+        return extract_file(file_path, max_pages=self._resolve_pdf_max_pages())
 
     def _get_fulltext_meta_for_item(self, item_id: int):
         meta = []
@@ -997,10 +1020,10 @@ class LocalZoteroReader:
                 return result
         return None
 
-    def _extract_fulltext_for_item(
+    def extract_fulltext_detailed(
         self, item_id: int, item_key: str | None = None
-    ) -> tuple[str, str] | None:
-        """Attempt to extract fulltext and source from the item's best attachment.
+    ) -> FulltextExtraction | None:
+        """Extract fulltext from the item's best attachment, with page limits.
 
         Preference order:
         1. Our own extraction of the best attachment on disk, chosen by
@@ -1018,6 +1041,10 @@ class LocalZoteroReader:
         If the sqlite-recorded filename doesn't resolve on disk, scan the
         attachment's storage folder for a content-type-matching file before
         giving up (#291, #265).
+
+        ``page_count`` and ``truncated`` are whatever the parser reported, so
+        they are set on that path alone: both caches store flat text with no
+        page bookkeeping, and neither is page-capped to begin with (#448).
         """
         chosen = self._resolve_extraction_target(item_id)
 
@@ -1026,17 +1053,26 @@ class LocalZoteroReader:
             target, attachment_key = chosen
             hit = self._cache_lookup(target, attachment_key)
             if hit:
-                return hit
-            text = self._extract_text_from_file(target)
-            if text:
+                return FulltextExtraction(*hit)
+            doc = self._extract_doc_from_file(target)
+            if doc:
                 source = _source_for_path(target)
-                self._cache_store(target, attachment_key, item_key, text, source)
-                return (text, source)
+                self._cache_store(target, attachment_key, item_key, doc.text, source)
+                return FulltextExtraction(
+                    doc.text, source, doc.page_count, doc.truncated
+                )
 
         # 2. Zotero's own cache, for whatever step 1 could not read.
-        return self._zotero_ft_cache_fallback(item_id, chosen, item_key)
+        cached = self._zotero_ft_cache_fallback(item_id, chosen, item_key)
+        return FulltextExtraction(*cached) if cached else None
 
-        return None
+    def _extract_fulltext_for_item(
+        self, item_id: int, item_key: str | None = None
+    ) -> tuple[str, str] | None:
+        """``extract_fulltext_detailed`` as the ``(text, source)`` pair that the
+        indexing paths consume."""
+        extraction = self.extract_fulltext_detailed(item_id, item_key)
+        return (extraction.text, extraction.source) if extraction else None
 
     def close(self):
         """Close database connection."""
--- a/src/zotero_mcp/tools/retrieval.py
+++ b/src/zotero_mcp/tools/retrieval.py
@@ -14,6 +14,7 @@ from zotero_mcp._app import mcp
 from zotero_mcp._context import Context
 from zotero_mcp.client import with_zotero_api_lock
 from zotero_mcp.config import load_config
+from zotero_mcp.extract import extract_file
 from zotero_mcp.tools import _helpers
 
 #: Pages of a PDF to surface when an agent reads a paper inline, if nothing
@@ -36,6 +37,28 @@ def _fulltext_display_max_pages() -> int:
     except Exception:
         return DEFAULT_FULLTEXT_DISPLAY_MAX
     return DEFAULT_FULLTEXT_DISPLAY_MAX if configured is None else configured
+
+
+def _fulltext_section_heading(
+    truncated: bool, page_count: int | None, max_pages: int, item_key: str
+) -> tuple[str, str]:
+    """Return the "Full Text" heading and any page-limit notice.
+
+    A capped read is byte-for-byte indistinguishable from a complete one, so
+    an agent summarizes and cites the first ten pages of a 442-page book as
+    if it had read the paper (#448). Name the range that was read instead,
+    and where to pick the document up again.
+    """
+    if not truncated:
+        return "## Full Text", ""
+    return (
+        f"## Full Text (pages 1-{max_pages} of {page_count} — TRUNCATED)",
+        f"> Only the first {max_pages} of {page_count} pages were extracted. "
+        f"Raise `semantic_search.extraction.fulltext_display_max_pages` in the "
+        f"zotero-mcp config to read more of the document in one call, or use "
+        f"zotero_read_pdf_pages(item_key='{item_key}', start_page={max_pages + 1}) "
+        f"to read on from where this stops.",
+    )
 
 
 @mcp.tool(
@@ -114,13 +137,16 @@ def get_item_metadata(
 @mcp.tool(
     name="zotero_get_item_fulltext",
     description=(
-        "Return the full extracted text of a Zotero item's primary "
+        "Return the extracted text of a Zotero item's primary "
         "attachment (PDF or EPUB). "
-        "WARNING: returns the entire paper (often 10K+ tokens). Use ONLY "
-        "when the user explicitly wants to READ the paper — not for "
+        "WARNING: returns most or all of the paper (often 10K+ tokens). Use "
+        "ONLY when the user explicitly wants to READ the paper — not for "
         "searching or browsing. For topic search use "
         "zotero_semantic_search; for metadata only use "
         "zotero_get_item_metadata. "
+        "PDFs are read up to fulltext_display_max_pages (10 by default); "
+        "when that cuts a document short the heading names the page range "
+        "and TRUNCATED — read on with zotero_read_pdf_pages. "
         "Avoid calling this on multiple papers in one conversation unless "
         "the user specifically asked to read several. "
         "item_key: 8-character Zotero item key. Normally the parent item — "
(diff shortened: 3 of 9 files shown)


danmackinlay:
Verified the two new commits against a real library (Web API, ~16.9k items), and found one case the title cleaner doesn't reach.

## `0f07f88`'s ISBN call site, live

Measured on this branch's tip:

```
find_existing_items(isbn=9780387004549)                        -> []
find_existing_items(isbn=..., title='Partial Identification…') -> ['57TNTZZB']
```

End to end, `add_by_isbn(if_exists='file')` on a book already shelved reuses it rather than duplicating — items carrying that ISBN before: 1, after: 1. The log shows the mechanism as described: first check misses, Open Library resolves, the second check finds it.

Across a sample of 25 ISBN-bearing books run through the whole chain: **22 found, 0 missed, 3 provider misses**. The last are cases where Open Library *and* Google Books have no record, so `add_by_isbn` returns `ISBN not found…` and exits before the dedup check ever runs.

That last number is worth stating, because it's an argument for the shape of the fix: the fallback's coverage is co-extensive with the bug's reachability. A duplicate is only creatable when the metadata lookup succeeds, and that is exactly when a title is in hand.

## One case `d9d2faf` doesn't cover

The cleaner fixes titles carrying *junk* tokens. The other way to overshoot token-AND is a title that is legitimately **longer** than the stored one — same consequence, but there's nothing to strip. Open Library returns `Title: Subtitle` against a stored short title routinely:

```
stored:   'Other Minds'   ISBN 9780374227760
provider: 'Other minds: the octopus, the sea, and the deep origins of consciousness'

_title_search_query(provider)                  -> unchanged
find_existing_items(isbn, title=provider)      -> []          # would duplicate
find_existing_items(isbn, title='Other Minds') -> [<item>]
```

Measured on this branch's tip, cleaner active.

Incidence on our library was 0/22 — but only because our stored titles happen to match Open Library's. Anyone who trims subtitles hits it. It's a book problem more than a paper one: arXiv and CrossRef titles are usually exactly what Zotero stored, which is why the arXiv and DOI results look clean.

**Possible follow-up, not proposed.** Since the match is token-AND, a shorter query is strictly more permissive — it matches everything the longer one matches and more, and the identifier still decides, so precision is unchanged. Querying the portion before the first colon *instead of* the full title would cost no extra call:

```
full provider title                        -> []
pre-colon 'Other minds'                    -> [<item>]
negative control (wrong ISBN, same head)   -> []
```

The trade is a larger candidate set against `limit=50`, which we haven't characterised. So we're flagging it rather than pushing it — say the word if you want it.

54yyyu:
@danmackinlay this is the fix I want for #441. Your diagnosis matches what I found back in August (the web API quick search never reaches `DOI`, `url`, `archiveID` or `extra`), and your version goes further than the title-hint fallback I had sitting on an unmerged branch: the four places an arXiv identity can live, version-insensitive comparison, and the measurement that one `<i>` or `&amp;` token zeroes the title query. I'm dropping my branch in favour of this.

Two things before it can go in:

1. **Rebase onto `maint/0.12.0`** (#544). That branch now has your #522 to #526, and #523 adds `utils.strip_unsupported_markup` and `utils.repair_crossref_string`. The title you pass as a hint on the DOI path is already entity-decoded and stripped of JATS/MathML by the time `_crossref_to_item_data` returns, so check whether your title cleaner still needs to do that part or can reuse those two. `find_existing_items` also has callers from #525's page read now.
2. **Mark it ready** when you're happy with it. I'll review it against a live web API library and a local Zotero 10 before merging.

If you'd rather I do the rebase, say so and I'll push it to this branch (maintainer edits are on).

danmackinlay:
Thanks. 0.12.0 shipped while we were on this, so it's rebased onto `maint/0.12.1` (`c1abf12`) rather than `maint/0.12.0`, and marked ready. The first six commits are the original ones replayed, with two small simplifications and one test folded in; the last two are new.

**The title cleaner and #523's helpers.** The cleaner is still needed; neither helper can stand in for it:

- `strip_unsupported_markup` deliberately keeps the markup Zotero renders: `<i>`, `<b>`, `<sub>`, `<sup>`, and `<scp>` as a styled `<span>`. Those are the tokens that take a quick-search query to zero.
- `repair_crossref_string` deletes newlines outright, which glues two words into one token when the result is used as a query.
- Titles from arXiv, Open Library/Google Books, a landing page or a BibTeX/CSL-JSON entry pass through neither helper.

Checking that turned up two bugs in the cleaner itself, fixed in `390c699`:

- **Tags inside a word.** Deleting a tag glued a token the stored title doesn't contain. We searched the web API for the 4 items in our library whose stored title carries markup, each by its own title. The old cleaner found 3. It missed `DREAM<sub>(D)</sub>: …`, because `DREAM(D):` isn't a substring of it. Tags are now replaced by a space, which finds all 4.
- **Decoded angle brackets.** #523 decodes entities before the cleaner runs, so `&lt;10 Hz … &gt;` arrives as a bare `<` … `>`, and `clean_html`'s `<.*?>` deleted the words between them. A tag is now `<` or `</` followed by a letter.

**Callers.** #525 doesn't add a `find_existing_items` call. It changes which routes reach `add_by_doi`'s post-fetch check, and they all go through the call site this PR already wires. #515 does add two: the ISBN and URL checks before an item is created from embedded page metadata. `8c4df31` passes the page's title to both. The identifier still decides; the title only brings the item into the candidate set. That covers "matching by title is left to #506" in #515's changelog entry, so that sentence may want rewording.

**Also:**
- An end-to-end `add_by_doi` test against the identifier-blind fake. The DOI path was wired but had never been exercised that way.
- `payload["item_data"].get("title")` in place of a guard on a key that is always set.
- `return []` where the code returned a `matches` that was always empty.

**Live check.** On our own library over the web API (~17k items), running `zotero-cli add … --if-exists file` on this branch before the move to 0.12.1 (`f7df1ae`; that rebase changed no dedup code):
- Three DOI adds whose CrossRef titles carry markup, including `DREAM<sub>(D)</sub>: …`, reused the existing items.
- A BibTeX import with a plain title reused an item stored with `<sup>` markup.
- For #515, adding a DOI-less PMLR page created one item, and adding it again reused it, matched by URL. Looking it up by URL alone found nothing; with the page's title it was found at once.
- A dry run with writes blocked showed two adds would still create duplicates, so we didn't run them: a DOI whose published title differs from the stored preprint title, and an ISBN whose Open Library title is longer than the stored one. Both are the recall limits described above.
- arXiv was rate-limiting us, so that path wasn't checked live.

Every commit passes the full suite on its own (3140 passed, 48 skipped at the tip), each new test fails on its commit's parent, and ruff reports nothing that isn't already on maint.

PR #507, opened 2026-08-31
perf(duplicates): count before paging, and page only the items that get compared

danmackinlay:
Noticed while investigating #506. Smaller and more self-contained than that one, but it shares the caveat: we're outside contributors reading this codebase for the first time, and the ceiling's intent isn't documented anywhere we found, so please treat the second change below as a question rather than a correction.

## Counting before paging

`_collect_duplicate_groups` reaches its over-size refusal by paging in `_DUP_SCAN_MAX_ITEMS + 100` items and then measuring the list:

```python
while True:
    batch = zot.items(start=start, limit=page_size)
    ...
    if len(items) > _DUP_SCAN_MAX_ITEMS:
        break

if len(items) > _DUP_SCAN_MAX_ITEMS:
    return {}, (f"Library has {len(items)} items — too large ...")
```

On a 16,859-item library that's ~51 sequential requests to arrive at "too large for duplicate scan". Measured:

```
web API    before:  163.10s   "Library has 5100 items — too large for duplicate scan."
local API  before:    4.60s   (51 requests)
           after:     0.05s   "Library has 16861 top-level items — too large ..."
```

`num_items()` answers the same question in one request. The reported count also becomes the real one — "5100" was just where paging gave up, not the library's size.

Only the whole-library path gets the precheck. `num_collectionitems()` counts a collection's items *including* children, so it's only an upper bound on the top-level count and couldn't refuse on its own; the paging guard already covers the in-range case, so asking would just be another request. A collection-scoped scan is unchanged.

The count is treated as an optimisation rather than a correctness requirement — if it raises, the paging guard still bounds the scan. There's a test for that.

## Paging top-level items

Separately, the scan pages with `items()`, which returns child attachments, notes and annotations. The grouping loop immediately discards those:

```python
if data.get("itemType") in ("attachment", "note", "annotation"):
    continue
```

So they consume the budget without ever being compared. On the same library they are **49.7%** of `/items` — 33,506 items in total against 16,859 top-level. `top()` excludes them.

**`top()` loses nothing.** Paging both endpoints over the whole library and applying the itemType filter to each yields the *identical* groupable population — 16,850 items either way, zero difference in the key sets — in 169 requests rather than 336 (10.6s rather than 29.8s). Repeated per collection, grouping output is identical and requests roughly halve:

| collection | `collection_items` | `collection_items_top` | groups identical |
|---|---|---|---|
| thesis | 919 items / 10 reqs | 456 items / 5 reqs | ✅ |
| learning_gamelan | 581 items / 6 reqs | 276 items / 3 reqs | ✅ |
| point_processes | 415 items / 5 reqs | 202 items / 3 reqs | ✅ |

The 9-item gap between 16,859 top-level and 16,850 groupable is standalone attachments and notes: they have no parent, so they *are* top-level, and the itemType filter is still doing real work. That's also why the test fakes filter on `parentItem` rather than on `itemType`.

## What this does to the ceiling

It moves what the ceiling measures, from "all items" to "top-level items", so a library with many attachments that was refused before may now be scanned. At a 49.7% child share the admitted top-level population roughly **doubles** — 5,000 all-items is about 2,515 top-level; 5,000 top-level is 5,000.

We think that's the right direction, and that the ceiling does not need lowering to compensate, for two reasons:

1. **The cost it bounds does not move.** Both before and after, the scan makes at most 51 requests and holds at most 5,100 item records. That worst case is unchanged; what changes is how much of that fixed budget is spent on rows the grouping loop will keep rather than discard.
2. **The scan is not quadratic.** The tool description says "the whole-library scan is O(n²) on titles", but the grouping is a single `groups.setdefault(...)` hash pass — O(n). Doubling the admitted population doesn't quadruple anything. That description is corrected here.

Also corrected: the description advertised "> 5,000 items", which after this change means 5,000 *top-level* items. On this library that's the difference between refusing at 16,859 and at 33,506, and a caller comparing it against the count Zotero displays would read the wrong number.

## Tests

Six added; four fail without the source change. They cover: the refusal costing one count and no paging, an in-range library still scanning, child items not consuming the budget, and a failed count falling back to the paging guard. The remaining two are regression guards for branches nothing exercised and which pass either way — the paging guard refusing when no count is available, and a collection scan refusing without spending a request on a count that could never refuse.

`FakeZotero` gains `top()` / `num_items()` / `collection_items_top()` mirroring the real endpoints — filtering on `parentItem` rather than `itemType`, since a standalone attachment is top-level — and `FakeZoteroForDuplicates` gains the paging variants alongside its existing `items()`. That's a shared fixture, so for blast radius: `main` is 5 failed / 2725 passed and this branch is 5 failed / **2731** passed — exactly the six new tests, no existing test changing status.

The 5 `test_rate_limit_guard.py` failures are pre-existing on an untouched `3cb3e2e` (they assume pyzotero 1.13.5; installed is 1.15.1). `ruff check` on the touched files reports only pre-existing findings, none on added lines.

## Where we could be wrong

- **The ceiling's intent.** Still undocumented as far as we found. Our argument above is that the resource bound is unchanged and the grouping is linear, so this shouldn't need a compensating change to the constant — but if 5,000 was chosen against the all-items count for a reason we haven't seen, that's the thing to push back on.
- **One library, one point in time.** The 49.7% child share is from a single personal library that stores a lot of PDFs. A library with few attachments would see much less benefit from the `top()` change, though the counting change helps regardless.
- **Group libraries.** `num_items()` builds its query from `library_type`/`library_id`, so it should follow a group override, but we have not tested it against a group library.

~~We haven't tested `ZOTERO_LOCAL`.~~ Now verified: the local API does serve the header `num_items()` reads (`Total-Results: 16859` on `/items/top`), and every measurement above marked "local" was taken through `get_zotero_client()` against a running Zotero 9.0.6.


danmackinlay:
Pushed 1f85abf, and rewrote the description above. No behaviour change from the first commit — this is documentation, corrected measurements, and two more tests.

We went back and checked the two things we'd flagged as uncertain, against the same library through `get_zotero_client()` with `ZOTERO_LOCAL=true`, and both came back better than we'd guessed:

**`top()` loses nothing.** Paging `/items` and `/items/top` over the whole library and applying the same itemType filter to each gives the *identical* groupable population — 16,850 items, zero difference in the key sets — in 169 requests rather than 336. Same result per collection. So the equivalence isn't an inference from the endpoint's contract any more, it's measured.

**`ZOTERO_LOCAL` works.** The local API does serve the header `num_items()` reads — `Total-Results: 16859` on `/items/top`, answered in 83ms against Zotero 9.0.6. That caveat is struck.

Three things we'd got wrong, all in the direction of making the change look worse than it is:

- **36.7% was unrepresentative.** That was the first 1,000 items. Library-wide the child share is 49.7% — 33,506 items against 16,859 top-level.
- **"Raises the ceiling by roughly a third" inverted the arithmetic.** At that share the admitted top-level population roughly *doubles*: 5,000 all-items is ~2,515 top-level, against 5,000 after.
- **But the cost the ceiling bounds doesn't move at all.** 51 requests and 5,100 records held is the worst case before and after. Only the useful fraction of that fixed budget changes. Which we think answers the question we'd left open — the constant shouldn't need lowering to compensate.

Related: the tool description claims the scan is `O(n²)` on titles. It isn't — the grouping is a single `groups.setdefault(...)` hash pass, so it's O(n), and doubling the admitted population doesn't quadruple anything. Corrected, along with the cap itself, which still advertised "> 5,000 items" when it now means 5,000 *top-level* items — on this library the difference between refusing at 16,859 and at 33,506, and the wrong number for anyone comparing it against what Zotero displays.

Smaller things in the same commit:

- The two refusal messages report different quantities — the precheck the library's real total, the paging guard where paging stopped. They now say which, with a note not to "fix" them into agreement.
- The fallback test stubbed `num_items` with a lambda that swallowed its own call, so nothing asserted the count had even been attempted and it passed against unpatched source. The stub now records, and the test asserts paging carried on. Four of six tests now fail without the source change, up from three of four.
- Added the two branches nothing covered: the paging guard refusing when no count is available, and a collection scan refusing without spending a request on a count that could never refuse. Those pass either way — they're guards for decisions this PR makes, not proof of them.

Suite: `main` is 5 failed / 2725 passed, this branch 5 failed / 2731 passed — exactly the six new tests, no existing test changing status, which also covers the blast radius of touching the shared `FakeZotero`. The 5 are the pre-existing pyzotero-version failures in `test_rate_limit_guard.py`.

54yyyu:
Sorry for the slow answer on this one. You read it correctly: the ceiling is not about memory or about the comparison, it is there to bound the number of sequential API requests a single tool call can make. The scan pages 100 items at a time through the API in both web and local mode, so without a limit a large library means hundreds of requests inside one call.

Given that, paging 5,100 items in order to report "too large" is the worst of both, and counting first is the right fix. Please go ahead with that part: ask for the count, refuse straight away if it is over the ceiling, and only page when the scan is actually going to run.

On paging only the items that get compared: I'm fine with that as long as the groups come out identical to a full scan for the same library, so a test that pins that equivalence would be the thing I'd look for.

Longer term the ceiling should not apply in local mode at all. Reads there go through the SQLite backend now, and a duplicate scan over SQLite is one query rather than hundreds of requests. That is a separate change and does not need to be part of this PR, I've opened #583 for it.

PR #508, opened 2026-08-31
fix(dedup): pass the title on the ISBN path too (and why the url path is left alone)

danmackinlay:
> **Stacked on #506.** This branch contains that PR's two commits as well, so the diff here shows all three. Only the last one — `fix(dedup): pass the title on the ISBN path too` — is new. Happy to rebase once #506 lands, or drop this if #506 goes a different way, since it depends on the `title=` parameter that one adds.

In #506 we noted the `isbn` and plain-`url` dedup paths "look like they have the same problem". We went and checked rather than assuming, and the answer differs between the two.

## ISBN: same problem, and the same fallback fits

Four items were selected *for having an ISBN*, then searched by their own ISBN:

| Item | ISBN | hits |
|---|---|---|
| 3X3TXTWX | `978-0-7803-9151-2` | 0 |
| 57TNTZZB | `978-0-387-00454-9` | 0 |
| WYI4U23M | `3-355-40948-0` | 0 |
| BXV9UMXU | `978-0-19-955204-7` | 0 |

Four of four, consistent with the DOI and arXiv results in #506 and with the docs note that `q` searches titles and creators only.

`add_by_isbn` checks twice — once before the Open Library / Google Books lookup and once inside the identifier lock afterwards. The second has a title in hand, so that's the one this changes. The first has no title and is left alone.

## Plain URL: same problem, but the fallback does *not* fit

Four items searched by their own stored `url` also returned zero hits each, so the search limitation is the same. But we're **not** proposing a change there, because the generic-webpage branch has no title to fall back to — it creates the item with the URL *as* the title:

```python
template = write_zot.item_template("webpage")
template["url"] = url
template["title"] = url
```

So `title=` would just be the URL again, which is the thing that doesn't match. That path needs a different answer — an exact-match endpoint, a local lookup, or normalising `http`/`https` and trailing slashes before comparing — and we don't have a confident proposal. Flagging it rather than papering over it.

Worth noting the plain-url matcher is also an exact string compare after `rstrip("/")`, so `http://` vs `https://`, a trailing slash, or a tracking query parameter all read as different items. We haven't measured how often that bites and haven't touched it.

## Tests

Two added, mirroring the arXiv ones in #506: the identifier-blind fake finds the book by title, and a same-title-different-ISBN is still rejected so the ISBN keeps deciding.

Full suite 2,741 passed; the 5 `test_rate_limit_guard.py` failures are pre-existing on an untouched `3cb3e2e`. `ruff check` unchanged.

## Where we could be wrong

- Everything in #506's "Where we could be wrong" applies here too — one library, Web API only, `ZOTERO_LOCAL` untested, and the extra call per miss.
- **Book titles are more collision-prone than paper titles.** Editions, translations and reissues share a title far more often than papers do. The ISBN still decides, so a wrong match shouldn't get through — but `limit=50` truncating the candidate list is a more realistic worry here than it is for arXiv IDs, and we haven't characterised it.
- **We only checked four ISBNs**, on one library. Fewer than the six DOIs in #506, because fewer items in that library carry one.

### DIFF
--- a/src/zotero_mcp/tools/_helpers.py
+++ b/src/zotero_mcp/tools/_helpers.py
@@ -810,11 +810,21 @@ def find_existing_items(zot, *, doi=None, arxiv_id=None, isbn=None, url=None,
         def _matches(data):
             return _normalize_doi(data.get("DOI") or "") == doi
     elif arxiv_id:
-        query = arxiv_id
+        # Compare on the version-independent identity, and search on it too:
+        # quick-search is a substring match, so the bare id finds a stored
+        # 'arXiv:2401.00001v2' while the versioned form would miss a stored
+        # bare one.
+        ident = _arxiv_identity(arxiv_id) or arxiv_id
+        query = ident
         def _matches(data):
-            if _normalize_arxiv_id(data.get("url") or "") == arxiv_id:
-                return True
-            return f"arxiv:{arxiv_id}".lower() in (data.get("extra") or "").lower()
+            # Zotero stores an arXiv identity in up to four places depending
+            # on how the item arrived (connector, DOI add, arXiv add, manual).
+            # Checking only url+extra misses connector- and DOI-sourced items,
+            # which is how a re-add duplicates a paper already in the library.
+            for field in ("url", "archiveID", "DOI"):
+                if _arxiv_identity(data.get(field) or "") == ident:
+                    return True
+            return f"arxiv:{ident}".lower() in (data.get("extra") or "").lower()
     elif isbn:
         query = isbn
         def _matches(data):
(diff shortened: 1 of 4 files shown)


danmackinlay:
**This PR is now redundant and closeable whenever you like.**

`0f07f88` on #506 makes the same ISBN change — the production code is character-identical, only the comment wording differs — and wires `_maybe_reuse_existing` besides, which this PR didn't. `git diff 9573bf4 origin/fix/arxiv-identity-dedup -- src/` shows #506 as a strict superset: everything proposed here is already there, plus `_title_search_query`, which this branch lacks.

So there is nothing left here to merge. We're leaving it open only so the ISBN measurements in the description stay easy to find; verification of that path against a real library is now on #506, along with one case the title cleaner doesn't reach.

Close it at your convenience — no need to review this one.

PR #509, opened 2026-08-31
feat(add): report the citation key when reusing an existing item

danmackinlay:
A deliberately partial change, and we'd rather propose the half we're confident in than guess at the other.

## The gap

The add tools report the item key. What actually goes into a document is the citation key, so a caller wanting it makes a second `zotero-cli get metadata <KEY> --output-format bibtex` call. That's how one of the duplicate items behind #506 got created here: the second call was wrapped in a shell command substitution around the write, so the write ran twice.

## What this changes

Only `_handle_existing_item` — the already-in-library path. There it's free: the dedup search already returned the full item, so `data.citationKey` is in hand and no extra request is needed. Measured on one real library, 496 of 500 top-level items carried one. An item without one omits the line, so nothing depends on Better BibTeX being installed.

```
Already in library: **RL's Razor** (`VLYF39QV`, matched by arXiv ID 2509.04259)

Item key: `VLYF39QV`
Citation key: `Shenfeld2025RLs`
```

Two tests: present and absent.

## The half we did not do, and why

The just-created path is genuinely awkward, and we don't think the asymmetry should be hidden:

- **Reuse:** the key is already in the response we hold. Free and reliable.
- **Create:** it isn't. Better BibTeX assigns citation keys on the desktop client, reacting to the item appearing in the local database. An item created through the Web API has to sync down before that happens. So a create can't read the key back without a re-fetch that may still come up empty — for an interval that depends on the user's sync, and that's unbounded if Zotero isn't running at all.

Options as we see them, none obviously right:

1. Leave create as-is (this PR). Asymmetric output, which is arguably confusing.
2. Re-fetch once after create and emit the key if present. Costs a request per add and still usually misses, since the sync won't have happened yet.
3. Say something explicit on create — "citation key not yet assigned" — so the asymmetry is stated rather than silent.
4. Have `add` take a flag to wait/poll for it. More machinery than the problem probably deserves.

We'd take (3) if you want symmetry, but it's your call about output stability, and we didn't want to impose one.

## Interaction with #506

This only becomes *visible* once the reuse path is actually reached. Per #506 the dedup search currently returns nothing against the Web API, so `if_exists='file'` rarely gets here. The code change is independent and correct either way, but if #506 is rejected this is mostly dead code in practice.

## Where we could be wrong

- **We did not test the create path**, because doing so means writing junk items into a real library. Everything above about BBT's assignment timing is reasoning from how the plugin works, not measurement — if you know it behaves differently through the API, that changes option (2) entirely.
- **99.2% is one library**, belonging to someone who uses Better BibTeX heavily. A library without BBT would see this line essentially never, which is fine but makes the feature less useful than that number suggests.
- **`data.citationKey` vs BBT's own store.** We read the native Zotero field. If BBT's pinned keys and that field can disagree — we don't know whether they can — then this could report a key that isn't the one BBT would export.
- **Output stability.** Adding a line to a rendered result may break anything parsing it. `add_from_file` extracts the item key from this block; we kept that line untouched and only appended after it, but you'd know of other consumers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **The add tools report the citation key when they reuse an item already in the library.** The item key is what the tools returned, but the citation key is what goes into a document, so a caller that wanted one made a second `zotero-cli get metadata <KEY> --output-format bibtex` call to get it — and wrapping that second call around a write is how at least one duplicate got created. On the reuse path the key is already in hand: the dedup search returns the full item, so `data.citationKey` costs no extra request. An item without one simply omits the line, so nothing here depends on Better BibTeX being installed. Only `data.citationKey` is read, never a `Citation Key:` line in `extra`: the two disagree — on a 497-item sample both were present on 77 items and differed on 11 — and the native field is the one Better BibTeX exports, because those `extra` lines are this package's own import residue, where `citation_import` preserves the *source document's* key. This is the reuse path only. On the just-created path the key is not available at the same cost: BBT assigns it on the desktop client after it sees the new item, so an item created through the Web API has to sync down first, and a re-fetch immediately after the create would usually still come up empty.
+
 ## [0.11.0] - 2026-08-25
 
 **Upgrading:** `zotero_semantic_search` now defaults to the active library instead of every indexed library. If you relied on the old implicit behaviour, pass `search_all_libraries=True`.
--- a/src/zotero_mcp/tools/write.py
+++ b/src/zotero_mcp/tools/write.py
@@ -327,6 +327,19 @@ def _handle_existing_item(write_zot, existing, coll_keys, tags, if_exists,
     dedup search already returned the full item, so ``data.citationKey`` is
     in hand. An item without one just omits the line, so nothing depends on
     Better BibTeX being present.
+
+    Deliberately only ``data.citationKey``, never a ``Citation Key:`` line in
+    ``extra`` — the two disagree, and ``extra`` is the wrong one. On a
+    497-item sample both were present on 77 items and differed on 11 of
+    those; every difference checked against BBT's own export resolved to the
+    native field. Those ``extra`` lines are this package's own import
+    residue: ``citation_import`` preserves the *source document's* key there
+    (``Citation Key: CoaseFirmMarket1988`` beside a native
+    ``Coase1988Firm``), so reading them would report a key BBT does not
+    export. Nothing is lost by ignoring them — no sampled item carried a key
+    in ``extra`` alone. Note this differs from ``search_by_citation_key``,
+    which consults both *on purpose*, so an item stays findable by the key it
+    was imported under.
     """
     item = existing[0]
     item_key = item.get("key")
--- a/src/zotero_mcp/tools/write.py
+++ b/src/zotero_mcp/tools/write.py
@@ -321,10 +321,18 @@ def _handle_existing_item(write_zot, existing, coll_keys, tags, if_exists,
 
     The report keeps the ``Item key: `KEY``` line that callers (and
     add_from_file's key extraction) rely on.
+
+    It also reports the citation key when the item has one, since that — not
+    the item key — is what goes into a document. It costs nothing here: the
+    dedup search already returned the full item, so ``data.citationKey`` is
+    in hand. An item without one just omits the line, so nothing depends on
+    Better BibTeX being present.
     """
     item = existing[0]
     item_key = item.get("key")
-    title = item.get("data", {}).get("title") or "(untitled)"
+    data = item.get("data") or {}
+    title = data.get("title") or "(untitled)"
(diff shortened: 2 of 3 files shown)


danmackinlay:
Pushed a follow-up commit. It changes no behaviour — it closes out the third uncertainty in the description, which turned out to have a measurable answer, and adds the tests and changelog entry that answer implies.

## `data.citationKey` vs BBT's own store — measured

The open question was whether the native field and BBT's pinned keys can disagree, and therefore whether this could report a key that isn't the one BBT exports. A review of this PR raised the natural follow-on: `search_by_citation_key` consults *both* `data.citationKey` and a `Citation Key:` line in `extra`, so shouldn't the reuse path fall back to `extra` too?

Measured over the Web API on the same library, 500 top-level items sampled:

```
sampled top-level items   : 497
native data.citationKey   : 496  (99.8%)
carried a key in Extra only:  0
both present              : 77, of which 11 disagree
```

So the fallback would be dead code — and dead code with a hazard attached, because the two sources do disagree, on 11 of the 77 items carrying both:

```
3E28CACM   native Coase1988Firm           extra CoaseFirmMarket1988
G5RIWQXV   native Kosoy2021InfraBayesian  extra KosoyAppel2021InfraBayesian
PDK3SR5M   native McCarthy1969Philosophical  extra McCarthyHayes1969Some
```

Asking BBT directly (`zotero-cli get bibtex`) resolves every one of them to the **native** field:

```
@book{Coase1988Firm, ...
@misc{Kosoy2021InfraBayesian, ...
@incollection{McCarthy1969Philosophical, ...
```

The reason is visible in the Extra field's shape:

```
Citation Key: KosoyAppel2021InfraBayesian
howpublished: AI Alignment Forum
```

That is this package's own import residue, not a BBT pin. `citation_import` writes `Citation Key: {citekey}` preserving the *source document's* key (`citation_import.py:504` and `:667`), alongside `arXiv:` and unmapped BibTeX fields. BBT later assigns its own key natively and exports that.

So reading `extra` here would report a key BBT never exports — worse than reporting nothing, since the caller has no reason to doubt it. Reading only `data.citationKey` is correct, and the follow-up commit records that reasoning at the call site so the next reader doesn't re-propose the fallback. Two tests pin it, so a later well-meaning fallback fails loudly.

Worth stating plainly that this is *not* an argument for changing `search_by_citation_key`: consulting both there is right, and deliberately so — it keeps an item findable by the key it was imported under. The two readers want different things. The docstring now says which is which.

## Also in the follow-up commit

- A test through `add_by_doi` rather than the private helper, which additionally guards `add_from_file`'s item-key extraction (the ``Item key: `KEY` `` regex at `write.py:4366`) against the new line landing somewhere that regex would reach first.
- A changelog entry under `## [Unreleased]`, since this is a user-visible output change.

## One correction to the description

It says this is "mostly dead code in practice" if #506 is rejected. That is a Web-API-only claim. #506 explicitly leaves `ZOTERO_LOCAL` untested, and `find_existing_items` goes through whichever API is configured — so if the local API does index identifiers in `q`, local-mode users reach the reuse path today and this is useful independent of #506. We haven't tested that either and don't want to assert it; flagging it because the description currently undersells the change if local search behaves differently.

## Noticed while measuring, not addressed here

Both concern that same misleading `Citation Key:` line in Extra, and both are separate from this PR — happy to open issues or PRs, whichever you prefer:

- **`semantic_search` stores the wrong citation key for these items.** `_get_item_metadata` reads the key from Extra only (`semantic_search.py:1244-1252`), so on the ~14% of both-present items where the two disagree, semantic search results carry a key that BBT does not export. That is the same bug this PR avoids, in a place that didn't have the native field to hand.
- **`annotations.py` mis-parses a lowercase Extra line.** `annotations.py:262-271` matches case-insensitively with `startswith` but strips with a case-*sensitive* `line.replace("Citation Key:", "")`, so a line spelled `citation key: fooBar2020` yields the whole line as the citation key rather than `fooBar2020`:

  ```
  'Citation Key: fooBar2020'  -> 'fooBar2020'
  'citation key: fooBar2020'  -> 'citation key: fooBar2020'   # wrong
  'citationkey: fooBar2020'   -> 'fooBar2020'
  ```

There are four hand-rolled parsers for this line in the tree (`_helpers._extra_has_citekey`, `client.py:417`, `semantic_search.py:1248`, `annotations.py:265`), each matching slightly differently. Consolidating them is probably worth doing, but it is a different change from this one and we didn't want to smuggle it in.

## Still yours to call

The create-path asymmetry is unchanged, and the four options in the description still stand. For what it is worth, the measurement above is mild evidence against option (2): the native field is populated for 99.8% of items *that have synced*, which is consistent with BBT assigning promptly on the desktop and not at all before the sync — so a re-fetch straight after a create would miss for a while and then succeed, which is the worst shape for a caller to reason about. We would still take (3) if you want symmetry.

54yyyu:
This one is good to go, please mark it ready! The citation key is already in the item the dedup search returned, so it costs nothing, and leaving the line out when there is no key is the right call. The failure you describe, a second lookup wrapped around the write, is a good reason on its own. It will need a rebase on current main.

Issue #512, opened 2026-08-31
Local-API mode leaks raw httpx exceptions under pyzotero >=1.15, bypassing typed error handling (silent duplicates under throttling)


danmackinlay:
`get_zotero_client()` and `get_local_zotero_client()` hand pyzotero a client built by `_make_local_http_client()` (`src/zotero_mcp/client.py:165`, used at `:235` and `:259`). That is an **httpx 0.28** client, since `src/` imports plain `httpx`. pyzotero 1.15.x internally uses **httpx2** (httpx 2.x, a separate package).

Responses produced by our client are therefore `httpx.Response`, and pyzotero's `_client._post_check` catches only `httpx2.HTTPError`:

```python
try:
    resp.raise_for_status()
except httpx2.HTTPError as exc:
    error_handler(self, resp, exc)
```

Disjoint hierarchies, so nothing is caught. `error_handler` never runs, and every HTTP error status reaches our code as a raw `httpx.HTTPStatusError` instead of the pyzotero exception the call sites are written against.

Verified against pyzotero 1.15.1 / httpx 0.28.1 / httpx2 2.12.0, constructing the client exactly as `_make_local_http_client` does and replaying canned error responses:

```
429 -> httpx.HTTPStatusError    (expected: TooManyRetriesError)
403 -> httpx.HTTPStatusError
404 -> httpx.HTTPStatusError
412 -> httpx.HTTPStatusError    (expected: PreConditionFailedError)
500 -> httpx.HTTPStatusError
```

## Why this matters beyond error typing

The 429 row reinstates a bug the codebase deliberately guards against. `find_existing_items` special-cases a throttled search precisely so it does not degrade to "not present" (`src/zotero_mcp/tools/_helpers.py:839`):

```python
except (TooManyRetriesError, TooManyRequestsError):
    # A rate-limited search is deliberately not swallowed. ...
    raise
except Exception as e:
    ...
    return []
```

With a leaked `httpx.HTTPStatusError` the first clause misses and the second catches. The dedup search reports "nothing found", and `if_exists='file'` creates a duplicate of an item that is present — the exact failure mode the comment in that block exists to prevent, now happening silently.

Two further consequences on the same path:

- **No retry and no backoff wait happen at all.** `error_handler` is what records the server's backoff duration, and the retry loop only continues when the response is a recognised 429. A throttled local client therefore hammers rather than backing off.
[...]

PR #513, opened 2026-08-31
test: build rate-limit fixtures from the HTTP library pyzotero uses (#511)

### LINKED ISSUE #511: Test suite fails on pyzotero >=1.15: the httpx to httpx2 migration bypasses pyzotero's error handling
`tests/test_rate_limit_guard.py` fails five ways on a clean checkout of main. It reproduces on an untouched `3cb3e2e` (`release: 0.11.0`) with no local changes, so this is environment drift rather than anything in the tree — main's last green run was [Aug 25](https://github.com/54yyyu/zotero-mcp/actions/runs/32799744535), and every PR opened since shows the same five red checks.

```
FAILED tests/test_rate_limit_guard.py::TestUpstreamRateLimitHandling::test_transient_429_is_retried_and_recovers
FAILED tests/test_rate_limit_guard.py::TestUpstreamRateLimitHandling::test_persistent_throttling_raises_rather_than_returning_bytes
FAILED tests/test_rate_limit_guard.py::TestUpstreamRateLimitHandling::test_429_without_backoff_header_raises
FAILED tests/test_rate_limit_guard.py::TestDedupUnderThrottling::test_throttled_search_does_not_read_as_no_match
FAILED tests/test_rate_limit_guard.py::TestDedupUnderThrottling::test_transient_throttling_still_finds_the_existing_item
5 failed, 1 passed
```

## Cause

pyzotero 1.15.x switched its HTTP library from `httpx` to `httpx2` (httpx 2.x, distributed under a separate package name). The two are disjoint class hierarchies.

The test module does `import httpx`, which resolves to httpx 0.28.1 — pulled in transitively by `mcp`/`httpx-sse`, not by pyzotero — and builds its fixtures as `httpx.Response(429, ...)`. pyzotero's `_client._post_check` then does:

```python
try:
    resp.raise_for_status()
except httpx2.HTTPError as exc:
    error_handler(self, resp, exc)
```

`httpx.HTTPStatusError` is not a subclass of `httpx2.HTTPError`, so it escapes uncaught. The retry loop in `_retrieve_data` never sees the 429, `error_handler` never records the backoff, and `TooManyRetriesError` is never raised.

That accounts for all five: the three in `TestUpstreamRateLimitHandling` leak `HTTPStatusError` where they expect a retry or `TooManyRetriesError`, and the two in `TestDedupUnderThrottling` degrade to `[]` because `find_existing_items`' generic `except Exception` catches the leaked error instead of the typed one above it.

Worth being precise about what did *not* change: the retry semantics these tests pin are intact in 1.15.1. `MAX_RETRY_ATTEMPTS` is still 3, and `errors.error_handler` still swallows a 429 carrying a backoff header and returns for the caller to retry. The tests assert the right behaviour — they just can't reach it through the wrong exception hierarchy.

## Why CI picked it up when it did

`pyproject.toml` has a load-bearing floor and no ceiling (`pyzotero>=1.13.5`), there is no lockfile, and the test job installs bare `pyzotero`:

```yaml
python -m pip install pytest pytest-timeout "fastmcp<4" pyzotero python-dotenv ...
```

So CI resolves whatever is newest — currently 1.15.1.

## Possible fixes

Changing the test module's import to `httpx2` alone makes all six pass locally. But that breaks the suite against pyzotero <1.15, which the floor still allows. So it probably wants one of:

1. **Bump the floor to `pyzotero>=1.15` and use `httpx2` in the tests.** Simplest, at the cost of dropping a range that currently works at runtime.
2. **Keep the floor and build fixtures from whatever HTTP library the installed pyzotero uses** (`try: import httpx2 as httpx` / `except ImportError: import httpx`). Keeps both ranges green and is explicit about the coupling.
3. **Add a ceiling (`pyzotero>=1.13.5,<1.15`)** as a stopgap — but please see the runtime issue linked below before treating 1.15 as merely a test problem.

Option 2 looks right to me, but the choice turns on how much you want to keep supporting pre-1.15, which I don't have the context to judge. Happy to send a PR for whichever you prefer.

## Not only the tests

The same library mismatch has a runtime consequence on the local-API path, filed separately as #512: our local client is an httpx 0.28 one, so pyzotero's error handling is bypassed there too and a throttled dedup search silently duplicates.

danmackinlay:
Fixes #511.

`tests/test_rate_limit_guard.py` fails five ways on a clean checkout of `main`, and has done since pyzotero 1.15.0 reached PyPI. Nothing in the tree changed — CI installs bare `pyzotero` against a floor with no ceiling and there is no lockfile, so it resolves whatever is newest. Every PR opened since [main's last green run](https://github.com/54yyyu/zotero-mcp/actions/runs/32799744535) shows the same five red checks.

## Cause

pyzotero 1.15 moved off `httpx` onto `httpx2` (httpx 2.x, published under a separate distribution name). The two are disjoint class hierarchies.

The test module did `import httpx`, which resolves to httpx 0.28.1 — pulled in transitively by `mcp`/`httpx-sse`, not by pyzotero — and built its fixtures as `httpx.Response(429, ...)`. pyzotero's `_client._post_check` then does:

```python
try:
    resp.raise_for_status()
except httpx2.HTTPError as exc:
    error_handler(self, resp, exc)
```

`httpx.HTTPStatusError` is not an `httpx2.HTTPError`, so the 429 escaped that clause uncaught. The retry loop in `_retrieve_data` never saw it, `error_handler` never recorded the backoff, and `TooManyRetriesError` was never raised. That accounts for all five: the three in `TestUpstreamRateLimitHandling` leak `HTTPStatusError` where they expect a retry or the typed error, and the two in `TestDedupUnderThrottling` degrade to `[]` because `find_existing_items`' generic `except Exception` catches the leaked error instead of the typed clause above it.

So the fixtures were wrong, not the behaviour they pin. pyzotero 1.15.1 keeps the retry semantics these tests exist to guard: `MAX_RETRY_ATTEMPTS` is still 3, and `errors.error_handler` still swallows a 429 carrying a backoff header and returns for the caller to retry.

## Fix

Build the fixtures from whatever HTTP library the installed pyzotero speaks, rather than from whichever one happens to import — option 2 of the three sketched in the issue.

I went looking for whether the range below 1.15 is worth keeping green before choosing, since option 1 (bump the floor to `>=1.15`) is simpler and the answer decides it. It is: the **full suite passes unmodified on pyzotero 1.14.0**, so nothing else in the tree has drifted past that range, and the floor comment in `pyproject.toml` argues at some length for keeping the floor where it is. Narrowing the supported range to fix a fixture seemed like the wrong trade. Say the word if you'd rather take option 1 and I'll cut it down.

One deviation from what I proposed in the issue. I suggested `try: import httpx2 as httpx / except ImportError: import httpx`, which does not actually distinguish the two — **both can be installed at once**. httpx arrives transitively whatever pyzotero does, and the comment on the `fastmcp<4` pin notes that FastMCP 4 swaps httpx for httpx2, which would put httpx2 on the path beside a pyzotero still using httpx and make that import guess wrong in exactly the direction this PR exists to fix. Asking pyzotero is unambiguous instead: the client it builds for itself is an instance of the library it speaks, on either side of 1.15. A future pyzotero whose client is neither raises with a message naming the class, so the next migration costs one legible error rather than five opaque failures.

## Verification

Full suite on both sides of the split, Python 3.12, dependencies installed exactly as the CI job does:

| | result |
|---|---|
| pyzotero 1.15.1 + httpx2 2.12.0 | 2715 passed, 45 skipped — *was 5 failed* |
| pyzotero 1.14.0 + httpx 0.28.1 | 2715 passed, 45 skipped |

The tests are not passing vacuously in either configuration: `test_transient_429_is_retried_and_recovers` only passes if the retry loop actually ran, and `TooManyRetriesError` is only raised if `error_handler` actually recorded the backoff — the two things that were being bypassed.

`ruff check` is clean. I did not run `ruff format` on the file: it was already non-conformant on `main`, and reformatting it would bury a 5-line change under unrelated churn.

## Not fixed here

**#512** — the same library mismatch on the local-API path at runtime, where `_make_local_http_client` hands pyzotero an httpx 0.28 client, so its error handling is bypassed for real and a throttled dedup search silently duplicates. That one needs a decision about whether `_make_local_http_client`'s HTTP/1.1 pinning (#160) carries over to httpx2 before it can be patched, and it is not a test-only change. Worth noting that this PR does not make the suite cover it: these fixtures now exercise the web-API path, which was never the broken one.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **The test suite was red on a clean checkout against pyzotero >=1.15 (#511).** pyzotero 1.15 moved off httpx onto httpx2 — httpx 2.x, published under a separate distribution name, and a disjoint class hierarchy. `tests/test_rate_limit_guard.py` did `import httpx`, which resolves to the httpx 0.28 that `mcp`/`httpx-sse` pull in regardless of pyzotero, and built its 429 fixtures as `httpx.Response`. pyzotero's `_post_check` catches only `httpx2.HTTPError`, so those responses escaped it uncaught: the retry loop never saw the 429, `error_handler` never recorded the backoff, and `TooManyRetriesError` was never raised. Five tests failed — three leaking `HTTPStatusError` where they expect a retry or the typed error, two degrading to `[]` because `find_existing_items`' generic `except Exception` caught the leaked error instead of the typed clause above it. Since CI installs bare `pyzotero` against a floor with no ceiling and there is no lockfile, it resolves whatever is newest, so every PR opened after 1.15.0 reached PyPI showed the same five red checks. The fixtures are now built from whatever HTTP library the installed pyzotero speaks — asked of pyzotero rather than guessed at import time, since both libraries can be installed at once — which keeps the declared `pyzotero>=1.13.5` range green rather than narrowing it. Nothing the tests pin actually changed: 1.15.1 still retries a rate-limited read three times and still waits out a server-supplied backoff. The same mismatch has a *runtime* consequence on the local-API path, where our own httpx 0.28 client bypasses pyzotero's error handling for real; that is #512 and is not addressed here.
+
 ## [0.11.0] - 2026-08-25
 
 **Upgrading:** `zotero_semantic_search` now defaults to the active library instead of every indexed library. If you relied on the old implicit behaviour, pass `search_all_libraries=True`.
--- a/tests/test_rate_limit_guard.py
+++ b/tests/test_rate_limit_guard.py
@@ -13,7 +13,8 @@ the symptom is a silent duplicate rather than an error, so these tests are the
 only thing standing between the two.
 """
 
-import httpx
+import importlib
+
 import pytest
 from conftest import DummyContext
 from pyzotero.zotero import Zotero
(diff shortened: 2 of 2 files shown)


danmackinlay:
Cross-reference, since these two are easy to read as competing: #514 (for #512, the runtime half of the same root cause) adds `tests/conftest.py::pyzotero_http_module()`, which resolves the library exactly the way this PR's private `_pyzotero_http()` does — off `type(zot.client).__module__`. Arrived at independently, but the suite shouldn't carry both.

More to the point, this PR *is* the follow-up #514's description offers to send: "the change is `import httpx` → `httpx = pyzotero_http_module()` in that one test module". I filed it standalone rather than waiting, because it fixes a red `main` — every open PR is currently carrying the same five failed checks — and that seemed worth decoupling from review of a runtime change.

Suggested sequencing, if it suits you: take this one as-is, and #514 drops the duplicate when it rebases. It already edits `tests/conftest.py`, so collapsing `_pyzotero_http()` into the shared helper is a one-line repoint there. Both PRs also add a `### Fixed` block under `## [Unreleased]`, so whichever lands second needs a CHANGELOG resolution regardless.

Happy to restack this on #514 and cut it to that single line instead, if you'd rather review the two together.

PR #514, opened 2026-08-31
fix(client): build the local HTTP client from pyzotero's own httpx (#512)

### LINKED ISSUE #512: Local-API mode leaks raw httpx exceptions under pyzotero >=1.15, bypassing typed error handling (silent duplicates under throttling)

danmackinlay:
Fixes #512. I said in the issue I'd work up a patch if you confirmed the direction — here it is as a concrete proposal rather than a wait. Happy to close it if you'd rather go a different way; the two open questions I flagged there are both answered below.

## What was wrong

Local mode is the one path where *we* hand pyzotero a client instead of letting it build its own. It exists to pin HTTP/1.1 for Zotero 8's local server (#160), and it built that client from `httpx`. pyzotero 1.15 moved to `httpx2`, so pyzotero's

```python
try:
    resp.raise_for_status()
except httpx2.HTTPError as exc:
    error_handler(self, resp, exc)
```

never fired for a response our client produced. Disjoint hierarchies, nothing caught, `error_handler` never runs.

## What the fix is

Build the local client from whichever HTTP library the installed pyzotero is built against, resolved from pyzotero itself:

```python
defining_module = sys.modules.get(zotero.Zotero.__module__)
for name in ("httpx2", "httpx"):
    module = getattr(defining_module, name, None)
    if module is not None:
        return module
```

That's `pyzotero._client` in every version the floor allows — it holds `httpx` in 1.13.5 and 1.14.0, `httpx2` in 1.15.x — so the `pyzotero>=1.13.5` range keeps working on both sides of the switch without a version comparison to keep in sync. A pyzotero whose library can't be identified warns once and falls back to `httpx`, which is today's behaviour rather than a crash.

## The two things I was uncertain about in the issue

**Does httpx2 need the same HTTP/1.1 pinning, and does it spell it the same way?** Yes and yes. `httpx2.HTTPTransport.__init__` has the identical `http1: bool = True, http2: bool = False` parameters, and `Client(transport=..., follow_redirects=True)` is unchanged. So #160's pin crosses untouched — same two keyword arguments, same defaults, and `_pool._http1/_http2` still set the way that test asserts.

**Is there a constraint keeping our own client on 0.28?** Not one I can find. `httpx` appears in `src/` only in `client.py`, and only for this helper — nothing else in the package constructs a client, catches an `httpx` exception, or annotates against one. `httpx>=0.27` stays in `dependencies` because the fallback path still needs it; I updated the comment there, which described the old reason.

## Evidence

Replaying canned responses through a client constructed exactly as the local path constructs one, on pyzotero 1.15.1 / httpx 0.28.1 / httpx2 2.12.0:

| status | before | after |
|---|---|---|
| 429 | `httpx.HTTPStatusError` | `TooManyRetriesError` |
| 403 | `httpx.HTTPStatusError` | `UserNotAuthorisedError` |
| 404 | `httpx.HTTPStatusError` | `ResourceNotFoundError` |
| 412 | `httpx.HTTPStatusError` | `PreConditionFailedError` |
| 500 | `httpx.HTTPStatusError` | `HTTPError` |

### Against a live local server

Since opening this I got hold of a running local Zotero — **9.0.6, serving HTTP/1.0**, 33.5k items — and reran the comparison unmocked, same server, both clients:

```
before (httpx 0.28): reads=3 ok, 404 -> httpx.HTTPStatusError
after  (fix)       : reads=3 ok, 404 -> pyzotero.errors.ResourceNotFoundError
```

So the #160 surface is intact — plain reads still work through the httpx2 client, transport still negotiates HTTP/1.1, `_pool._http1/_http2` still `True/False` — and a real 404 from the real server now arrives typed instead of leaking. `find_existing_items` was also run against a DOI known to be present in that library and returned the match, so the dedup path is confirmed positively and not just in its failure mode.

The 429 row stays mocked: inducing real throttling against someone's local server to watch it back off is not a test worth running. What the live run establishes is that the client swap is transparent to Zotero and that the typed-exception plumbing works end to end on the wire.

And the consequence that actually costs data: with the leaked error, `find_existing_items` misses its `except (TooManyRetriesError, TooManyRequestsError): raise` clause and lands in the generic `except Exception: return []` below it, so a throttled dedup search reports "nothing found" and `if_exists='file'` duplicates an item that is present. After the fix it propagates, which is what that clause is there for.

## Tests

New `tests/test_local_client_error_typing.py`: five tests (seven with parametrisation) that build a local-mode client through `_make_local_http_client` and replay responses through a mock transport. The responses are constructed from *that client's own* HTTP library, not from a fixed import — that coupling is the thing under test, and pinning a library name in the fixture would make the test assert the fix's own opinion instead of the requirement. All seven fail against the pre-fix client.

`tests/test_local_http1_transport.py` (the #160 regression test) asserted `isinstance(c, httpx.Client)` against a bare `import httpx`, so it would have gone red on this change. It now takes its client and transport classes from a `pyzotero_http_module()` helper in `conftest.py`, which reads them off a client pyzotero constructs for itself — an oracle independent of how `client.py` picks, so `test_local_client_is_the_client_pyzotero_would_have_built` is a real assertion rather than a restatement.

Both modules pass against **pyzotero 1.14.0** as well as 1.15.1, which is the point of resolving rather than pinning.

Full suite: **2,720 passed, 42 skipped**, plus the 5 `tests/test_rate_limit_guard.py` failures that are pre-existing on an untouched `3cb3e2e` — that's #511, untouched here, and it also passes on 1.14.0 in the same run, which is consistent with it being 1.15 drift rather than anything in the tree. `ruff check` clean on every file this touches.

## What this does *not* fix

**#511.** Same root cause, different surface, and its fix turns on a policy call I don't have the context to make — you listed three options and I didn't want to pick one inside a runtime bugfix. If you want option 2, `conftest.pyzotero_http_module()` added here is exactly what it needs, and the change is `import httpx` → `httpx = pyzotero_http_module()` in that one test module. Say the word and I'll send it, or fold it in here if you'd rather have one PR.

## Where I could be wrong

- **One live server, one Zotero version.** The live run above is Zotero 9.0.6 on macOS. It closes the "does the real server still answer an httpx2 client" question for that version, but Zotero 8 specifically — the version #160 was reported against — I have not put behind it. The HTTP/1.1 pin is byte-identical across the two libraries, so I'd be surprised, but it is one data point rather than a matrix.
- **`_transport` and `_pool._http1` are private attributes.** The tests reach into them, following what the #160 test already does. If either library renames them the tests break, loudly, which is the failure mode I'd want.
- **The fallback is a judgement call.** If pyzotero restructures so neither name is a module global, this warns and uses `httpx`, silently restoring today's bug until someone reads the log. Raising instead would be louder but would take down local mode over a naming change. I picked the quieter one; happy to flip it.
- **`lru_cache` on the resolver** assumes pyzotero isn't swapped underneath a running process. True for any normal run, not for a test that reloads pyzotero — none does today.


danmackinlay:
I've sent the #511 follow-up this description offers, as #513 — option 2 from that issue, `import httpx` → the resolved module in `tests/test_rate_limit_guard.py`. Green on all five test legs.

One overlap worth flagging rather than leaving you to find it: #513 carries a private `_pyzotero_http()` in that test module which resolves via `type(zot.client).__module__` — the same mechanism as `conftest.pyzotero_http_module()` here. Independent convergence on the same technique, but the suite shouldn't end up with both.

Cheapest resolution is probably to take #513 first, since it clears the five failed checks currently showing on every open PR, and drop the duplicate here on rebase: this PR already touches `tests/conftest.py`, so it's a one-line repoint in `test_rate_limit_guard.py`. Both also add a `### Fixed` block under `## [Unreleased]`, which will want resolving in the same pass.

Or fold #511 into this PR and I'll close #513 — your call on whether one change or two is easier to review.

PR #516, opened 2026-08-31
fix: report page truncation in get_item_fulltext (#448)

### LINKED ISSUE #448: Issue: zotero_get_item_fulltext returns only a partial excerpt for large PDFs
# Issue: `zotero_get_item_fulltext` returns only a partial excerpt for large PDFs

**Project:** zotero-mcp-server  
**Severity:** Medium – incomplete literature analysis for long articles

## Problem

For large PDFs, `zotero_get_item_fulltext` returns only the first ~50 pages (or a character-limited excerpt) instead of the entire document text.

## Observed behavior

- Items with long PDFs (> ~50 pages) are truncated.
- The tool provides no parameter to request a specific page range or continuation token.
- Users must fall back to `zotero_read_pdf_pages` and call it blockwise.

## Expected behavior

One or more of the following would solve the problem:

1. `zotero_get_item_fulltext` returns the complete text, or documents clearly when truncation occurs.
2. A `page_start` / `page_end` parameter is added to `zotero_get_item_fulltext`.
3. The tool returns a pagination cursor so callers can fetch the rest.

## Environment

- zotero-mcp-server via SSH/Local API
- Observed with textbooks and long review articles (e.g., 442-page PDFs)

## Workaround currently used

Call `zotero_read_pdf_pages` repeatedly in blocks. This works but is slower and more complex than a single full-text call.

## Suggested implementation

If the limitation comes from context-window or API constraints, expose explicit pagination. If it is a bug in the extraction logic, return the full extracted text that Zotero already stores in its SQLite full-text index.

lots-o:
Fixes #448.

> Rebased onto `main` after the pdf-inspector migration. The fix got substantially smaller: `ExtractedDoc` already reports `page_count` and `truncated`, so nothing needs measuring — the facts only had to reach the caller. The earlier revision hand-rolled a page-counting subprocess on pdfminer; all of that is gone.

## The problem

`zotero_get_item_fulltext` stops at `fulltext_display_max_pages`, which defaults to 10. Nothing said so. The string returned for the first ten pages of a 442-page book was byte-for-byte indistinguishable from the string returned for a complete three-page note — same `## Full Text` heading, no marker, no log line — and the tool description promised the opposite:

```
"WARNING: returns the entire paper (often 10K+ tokens)."
```

So an agent summarizes and cites the excerpt as if it had read the paper. The reporter's phrasing — "incomplete literature analysis for long articles" — is exactly the failure, and it is invisible at the call site.

Verified against a real 51-page PDF in my library on unpatched `main`: ten pages back, nothing to distinguish it from the whole document.

## The fix

Both capped paths carry the parser's own page bookkeeping through to the heading:

```
## Full Text (pages 1-10 of 51 — TRUNCATED)

> Only the first 10 of 51 pages were extracted. Raise
> `semantic_search.extraction.fulltext_display_max_pages` in the zotero-mcp config
> to read more of the document in one call, or use
> zotero_read_pdf_pages(item_key='PWYKK9U7', start_page=11) to read on from where
> this stops.
```

A read that fits under the cap keeps the plain `## Full Text` heading — no notice, no change for short papers, which is the common case.

Both blocks are real output from the patched code against my own library, not a mock-up.

### The seam change, and why

`LocalZoteroReader._extract_text_from_file` becomes `_extract_doc_from_file` and returns the whole `ExtractedDoc` rather than `doc.text`. Two reasons:

- **One source of truth.** How long a document is and whether the cap cut it short are the parser's to report. A caller that re-derives them (extract, then count pages separately, then compare against the cap) creates a second answer that can disagree — and `extract.py`'s own docstring makes the case for deciding page provenance once, at the seam.
- **Nothing wanted text-only.** After this change `_extract_text_from_file` had no production callers left; it would have survived purely as a test stub point.

That moves 8 stubs in 4 existing test files from the text seam to the document seam. Mechanical — `lambda _path: "text"` becomes `lambda _path: extracted_doc("text")` via a new `extracted_doc` helper in `conftest.py`. Those tests are about attachment *selection*, so their subject is unchanged.

`_extract_fulltext_for_item` keeps its `(text, source)` contract, so the semantic indexer and `get_items_with_text` are untouched.

### Details worth flagging

- **Neither cache is reported as truncated.** The transient fulltext cache and Zotero's own `.zotero-ft-cache` store flat text with no page bookkeeping, and neither is page-capped to begin with. A notice on those would be a lie; there's a test that pins it.
- **The download+parse fallback applies the same cap** (`convert_to_markdown(..., max_pages=max_pages)`), so it had the same bug. It now calls `extract_file` directly for the page facts. `convert_to_markdown` is left as-is — it's re-exported from `server.py`, and changing its return type would be a breaking change for a helper that legitimately wants text.
- **The Zotero-index path is not capped** and is deliberately untouched: `zot.fulltext_item` serves Zotero's own complete text.
- **The tool description** now states the cap and points at `zotero_read_pdf_pages`. It grows 209 → 299 tokens, under the 450 hard cap in `test_description_tokens.py`; the tool isn't in `TOOL_BUDGETS`.

## What this deliberately does not do

The issue also floats `page_start`/`page_end` parameters on `zotero_get_item_fulltext`. I left them out: `zotero_read_pdf_pages(item_key, start_page, end_page)` already does exactly that, and adding a second page-ranged reader would duplicate the surface #422 worked to cut. The notice names that tool and the first unread page instead, so the excerpt becomes a place you can continue from rather than a dead end. Happy to add them in a follow-up if you'd rather.

## Verification

`tests/test_fulltext_truncation.py`, 11 tests: page bookkeeping reaching the caller; the complete-read case; the `.zotero-ft-cache` fallback and the transient-cache hit both reporting no page limits; `_extract_fulltext_for_item` still returning `(text, source)`; heading rendering including that the notice resumes at `start_page=11` and not page 1; and end-to-end `get_item_fulltext` on both capped paths.

Full suite with `pdf-inspector` installed: **2736 passed, 5 failed** — the same 5 that fail on clean `main` in my environment (4 are the pyzotero >=1.15 httpx migration from #511; 1 reads my real Zotero prefs and picks up a Google Drive data directory). I diffed the failure sets before and after: **no new failures**. `ruff` clean on every file this touches.

Live check against my own library, on this branch: a 51-page PDF reports `pages 1-10 of 51 — TRUNCATED`; a 10-page PDF under the 10-page cap reports plain `## Full Text` with no notice; a `.zotero-ft-cache` item likewise.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **`zotero_get_item_fulltext` says when a document was cut short.** Reads stop at `fulltext_display_max_pages` (10 by default), but the text returned for the first ten pages of a 442-page book was byte-for-byte indistinguishable from the text returned for a complete three-page note: same `## Full Text` heading, no marker anywhere, and a tool description that promised "the entire paper". An agent summarized and cited the excerpt as if it had read the whole thing. The parser already reports `page_count` and `truncated` on every extraction, so both capped paths — local storage and the last-resort download+parse — now carry them through to the heading: `## Full Text (pages 1-10 of 442 — TRUNCATED)`, followed by the config key to raise and the `zotero_read_pdf_pages` call that resumes at the first unread page. A read that fits under the cap is unchanged. Zotero's own `.zotero-ft-cache` and the transient fulltext cache are not page-capped and are never reported as truncated. `LocalZoteroReader`'s parser seam returns the whole `ExtractedDoc` instead of just its text, so page bookkeeping has one source of truth rather than being re-derived by callers (#448).
+
 ## [0.11.0] - 2026-08-25
 
 **Upgrading:** `zotero_semantic_search` now defaults to the active library instead of every indexed library. If you relied on the old implicit behaviour, pass `search_all_libraries=True`.
--- a/src/zotero_mcp/local_db.py
+++ b/src/zotero_mcp/local_db.py
@@ -22,6 +22,7 @@ from . import fulltext_cache
 from . import search_semantics as _semantics
 from .config import load_config
 from .extract import (
+    ExtractedDoc,
     categorize_attachment,
     extract_file,
     normalize_attachment_priority,
@@ -65,6 +66,23 @@ class KeyGroupMap(NamedTuple):
 
     groups: dict[str, int]
     excluded_keys: set[str]
+
+
+@dataclass
+class FulltextExtraction:
+    """Extracted attachment text, plus how much of the document it covers.
+
+    ``page_count`` is the document's real length and ``truncated`` says the
+    page cap dropped pages from the tail — both straight from
+    :class:`~zotero_mcp.extract.ExtractedDoc`. They carry their defaults on
+    the cache paths, which are not page-capped and keep no page bookkeeping
+    (#448).
+    """
+
+    text: str
+    source: str
+    page_count: int | None = None
+    truncated: bool = False
 
 
 def _read_string_pref(prefs_path: Path, pref: str) -> str | None:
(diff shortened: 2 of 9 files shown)


lots-o:
CI note: the five failing tests are all in `tests/test_rate_limit_guard.py` and are not from this branch — they are the pyzotero >=1.15 httpx migration (#511, fix in #513). I pulled the failure lists from this PR's run and from `main`'s own run of the same workflow (33395995391 vs 33391898366) and they are identical, with nothing unique to this branch:

```
TestUpstreamRateLimitHandling::test_transient_429_is_retried_and_recovers
TestUpstreamRateLimitHandling::test_persistent_throttling_raises_rather_than_returning_bytes
TestUpstreamRateLimitHandling::test_429_without_backoff_header_raises
TestDedupUnderThrottling::test_throttled_search_does_not_read_as_no_match
TestDedupUnderThrottling::test_transient_throttling_still_finds_the_existing_item
```

Everything else is green: 2736 passed locally with `pdf-inspector` installed.

PR #521, opened 2026-09-02
fix: follow pyzotero 1.15 to httpx2 — local-mode errors escaped raw, and CI has been red since 08-31

josk0:
## The bug

CI on `main` has been red on all five platforms since [the 2026-08-31 run](https://github.com/54yyyu/zotero-mcp/actions/runs/33391898366) — on the 0.11.0 release commit, which passed on 08-25. The difference between the two runs is pyzotero: 1.14.0 then, 1.15.1 now. pyzotero 1.15.0 (2026-08-29) replaced `httpx` with `httpx2` (urschrei/pyzotero@272634c).

That breaks more than the tests. For the local Zotero server, `_make_local_http_client()` hands pyzotero an `httpx.Client` pinned to HTTP/1.1 (#160). pyzotero turns error responses into its exception classes with `raise_for_status()` inside `except httpx2.HTTPError` — which does not catch the `httpx.HTTPStatusError` that an `httpx` response raises. So with pyzotero 1.15.1 in local mode, every non-2xx escapes raw:

```
$ python -c "..."   # Zotero() with an httpx.Client whose transport answers 404
BUG: httpx.HTTPStatusError - Client error '404 Not Found' for url 'https://api.zotero.org/users/1/items/NOPE'
```

instead of `ResourceNotFoundError`. Nothing downstream that catches pyzotero's errors (`PreConditionFailedError` in the write helpers, `TooManyRetriesError`/`TooManyRequestsError` in `find_existing_items`, ...) sees them any more.

The five `test_rate_limit_guard.py` failures are the same mismatch from the other side: the tests fabricate `httpx.Response` objects, whose 429 never reaches pyzotero's retry loop, hence `httpx.HTTPStatusError: Client error '429'` on every platform.

## The fix

`client._pyzotero_http()` returns whichever library pyzotero itself imports (`httpx2` from 1.15, `httpx` before; read off `pyzotero._client`, the one place the choice is recorded), and `_make_local_http_client()` builds from that. The two libraries share the `Client` / `HTTPTransport(http1=, http2=)` / `MockTransport` API, so nothing else changes. The rate-limit and HTTP/1.1 transport tests fabricate their responses from the same module, so they hold on either side of the switch.

Not touched: the `httpx>=0.27` entry in `pyproject.toml`. After this change nothing in `src/` imports `httpx` directly, so it could go, but that is a separate decision.

## Verification

- New `tests/test_pyzotero_http_module.py`: the helper names the library a real `Zotero()` instance uses; the local client is that library's `Client`; and the regression itself — a 404 through our client is a `ResourceNotFoundError`.
- The three affected test files pass against pyzotero 1.15.1 (locked) and against 1.14.0 (`uv run --with "pyzotero==1.14.0" pytest ...`).
- Full suite: 2733 passed, 39 skipped, 0 failed (was 5 failed on `main`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **With pyzotero 1.15, every error response from the local Zotero API escaped as a raw `httpx.HTTPStatusError` instead of a pyzotero error.** pyzotero 1.15.0 replaced `httpx` with `httpx2`. For the local server we hand pyzotero our own client, pinned to HTTP/1.1 (#160), and that client was still an `httpx.Client`. pyzotero classifies error responses by calling `raise_for_status` inside `except httpx2.HTTPError`, which does not catch the `httpx` error the old client's responses raise — so a 404 was no longer a `ResourceNotFoundError`, a 412 no longer a `PreConditionFailedError`, and nothing downstream that catches pyzotero's errors saw them. The same mismatch is why the rate-limit tests, which fabricate `httpx.Response` objects, failed on every platform against 1.15.1: the 429 never reached pyzotero's retry loop. The local client is now built from whichever library pyzotero itself imports (`httpx2` from 1.15, `httpx` before — the two share the `Client`/`HTTPTransport` API), and the tests fabricate their responses from the same one, so both hold on either side of the switch. Verified against pyzotero 1.14.0 and 1.15.1.
+
 ## [0.11.0] - 2026-08-25
 
 **Upgrading:** `zotero_semantic_search` now defaults to the active library instead of every indexed library. If you relied on the old implicit behaviour, pass `search_all_libraries=True`.
--- a/src/zotero_mcp/client.py
+++ b/src/zotero_mcp/client.py
@@ -13,7 +13,6 @@ from dataclasses import dataclass
 from pathlib import Path
 from typing import Any
 
-import httpx
 from dotenv import load_dotenv
 from pyzotero import zotero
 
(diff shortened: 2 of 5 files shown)


danmackinlay:
poss duplicate #514

danmackinlay:
Apologies for the collision — I'd sent the same fix as #514 (runtime) and #513 (the test fixtures) the day before, no way you'd have known.

Same diagnosis, same fix, and as far as I can tell yours resolves the same module as mine everywhere it matters: on pyzotero 1.13.5 (the floor), 1.14.0 and 1.15.1, `Zotero.__module__` is `pyzotero._client`, so reading the attribute off `_client` and reading it off the class agree across the whole supported range. Either works.

Two differences I think are real, though I'm obviously not the neutral party here:

- Yours imports `pyzotero._client` unguarded, so if that private module is ever renamed the `import httpx` fallback just below can't run — the call raises `ImportError` instead of degrading. Mine goes via `zotero.Zotero.__module__` and warns on fallback. Speculative failure mode, low stakes.
- I mutated `_make_local_http_client()` back to a plain `httpx.Client` under 1.15.1, leaving the helper intact, to see which suite catches the regression. Mine fails 7 tests in `test_local_client_error_typing.py` (403/404/412 typed errors, the 429 retry-and-recover, and `find_existing_items` raising rather than reading as "no match" — the silent-duplicate path). Yours fails the `isinstance` check, but `test_error_responses_still_become_pyzotero_errors` still passes, because it assigns its own client to a `Zotero()` rather than going through `_make_local_http_client()` — so it pins pyzotero's behaviour rather than ours.

Where yours is better: it's one patch rather than two, it has a proper CHANGELOG entry, and it drops the now-dead top-level `import httpx` that mine deliberately keeps as the fallback. One thing to catch either way — the `httpx>=0.27` comment in `pyproject.toml` says client.py imports httpx directly, which stops being true with your version.

Genuinely happy either way, @54yyyu's call. If #521 is the one, `tests/test_local_client_error_typing.py` from #514 should graft on unchanged and I'll close mine. And an idea better than both, if anyone wants it: resolve the library as `type(Zotero(...).client).__module__` — public API only, survives any internal reshuffle. My tests already do that; neither PR does it in `src/`.

josk0:
Thanks for the careful comparison, and no apology needed. You were two days ahead and filed the issues; I should have checked the open PRs before opening this one.

I've read #513 and #514 against mine and agree on every point. Your tests go through `_make_local_http_client` where mine sidestep it, your fallback degrades with a warning where mine would raise on a rename of `pyzotero._client`, and I left the httpx comment in `pyproject.toml` stale. The only thing #521 had over the pair was being a single patch, and that is not worth the duplication.

Closing in favour of #513 and #514. @54yyyu, from a bystander's seat: #513 is green and one-file, and merging it first turns CI green on every open PR. #514 then rebases onto it with a one-line repoint in `test_rate_limit_guard.py`, as Dan describes.

One suggestion for #514, take it or leave it: your idea of resolving the library as `type(Zotero(...).client).__module__` in `src/` as well as the tests would remove the last dependency on a private module name. Happy to test either version against my Zotero 9 local instance on macOS if that helps.

PR #530, opened 2026-09-07
fix: cap pyzotero at <1.15 — 1.15 drops TooManyRetriesError, breaking 429 handling

jianxing-chen:
## Problem

CI is currently red on `main` and on open PRs (e.g. run [34116303775](https://github.com/54yyyu/zotero-mcp/actions/runs/34116303775): 25 failures — all of `tests/test_rate_limit_guard.py` across the 5-job matrix):

```
httpx.HTTPStatusError: Client error '429 Too Many Requests' for url 'https://api.zotero.org/users/1/items'
```

## Root cause

pyzotero 1.15 removed `TooManyRetriesError` (only `TooManyRequestsError` remains). The internal retry path then raises a raw `httpx.HTTPStatusError` on a persistent 429, which `find_existing_items` does not catch — the exact escape the `>=1.13.5` floor was added for (urschrei/pyzotero#352).

This is not hypothetical: with pyzotero 1.15.1 installed, all five rate-limit tests fail; with 1.14.x they pass (verified on both 1.15.1 and 1.14.0 locally, and in CI logs on both sides of the version bump).

## Fix

- Cap `pyzotero>=1.13.5,<1.15` in `pyproject.toml`, extending the existing floor's comment.
- Apply the same pin in the workflow's test-dependency install line: `pip install -e . --no-deps` skips pyproject's constraints entirely, so the unpinned install resolves 1.15.1 regardless of the cap. Also pinned `bibtexparser>=1.4,<2` there to match pyproject (v1-only API), for the same bypass reason.

## Notes

- The cap is a stopgap; a follow-up could adapt `find_existing_items`' error handling to 1.15's exception shape and lift it.
- With these two pins the full 5-job matrix goes green again (same fix verified end-to-end on my fork's identical workflow: [green run](https://github.com/jianxing-chen/zotero-mcp/actions/runs/34122584767)).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -38,7 +38,9 @@ jobs:
       - name: Install package and test dependencies
         run: |
           python -m pip install --upgrade pip
-          python -m pip install pytest pytest-timeout "fastmcp<4" pyzotero python-dotenv "pdf-inspector==0.2.6" markdownify chromadb unidecode bibtexparser tiktoken pymupdf
+          # pyzotero must be pinned here too: the editable install below uses
+          # --no-deps, which skips pyproject's dependency constraints.
+          python -m pip install pytest pytest-timeout "fastmcp<4" "pyzotero>=1.13.5,<1.15" python-dotenv "pdf-inspector==0.2.6" markdownify chromadb unidecode "bibtexparser>=1.4,<2" tiktoken pymupdf
           python -m pip install -e . --no-deps
       - name: Run tests
         run: pytest --ignore=tests/test_lifespan.py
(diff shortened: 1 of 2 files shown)


54yyyu:
Thanks for chasing the red CI, and for spotting that `release.yml` installs pyzotero without the pyproject constraints.

I'm closing this in favour of #513 + #514 rather than capping. The root cause turned out to be different from the one in the description: `TooManyRetriesError` is still there in 1.15.1 (`pyzotero/_client.py` raises it after `MAX_RETRY_ATTEMPTS`, and it is exported from `pyzotero.zotero_errors`). What changed is that 1.15 moved from `httpx` to `httpx2`, and the two have unrelated exception classes. Our test fixtures built `httpx` responses and our local-mode client was an `httpx.Client`, so pyzotero's `except httpx2.HTTPError` never fired and the 429 escaped as a raw `HTTPStatusError`.

#513 builds the fixtures from whichever library pyzotero uses, and #514 does the same for the local client, which also fixes the runtime side (silent duplicates under throttling in local mode, #512). With both merged the full suite passes on pyzotero 1.14.0 and 1.15.1, so the supported range can stay open.

Both are on the `maint/0.12.0` integration branch (#544) and will ship in 0.12.0.

Issue #499, opened 2026-08-27
zotero-cli --json search returns empty in local mode (local API itemKey filter returns children first, limit=len(chunk) truncates parents)


5258MF:
## Summary

`zotero-cli --json search "<query>"` returns `count: 0` (empty items) when running in **local mode** (`ZOTERO_LOCAL=true`), even though the same search works in markdown mode and returns items via the web API. The markdown (`non-JSON`) path works fine, so the JSON rendering pipeline is where the bug lives.

## Environment

- `zotero-mcp-server` 0.11.0 (latest on PyPI as of today)
- Windows 11, Zotero 9 desktop, local API enabled on port 23119
- Hybrid config: `ZOTERO_LOCAL=true`, `ZOTERO_API_KEY`, `ZOTERO_LIBRARY_ID` (local userID), `ZOTERO_SEARCH_BACKEND=sqlite`

## Reproduction

```bash
# markdown mode — works, finds the item
zotero-cli search "GoLLIE"

# JSON mode — returns empty despite the same item existing
zotero-cli --json search "GoLLIE" --limit 5 --detail keys_only
# => {"ok": true, "data": {"query": "GoLLIE", "count": 0, "items": []}}
```

The same JSON search works when local mode is disabled (web API only).

## Root cause

The JSON path (`cmd_search` → `_fetch_projected` in `cli_standalone.py`) reconstructs results by fetching the markdown-extracted item keys back from the API:

```python
for i in range(0, len(keys), 50):
    chunk = keys[i:i + 50]
    batch = zot.items(itemKey=",".join(chunk), limit=len(chunk))
```

Two things combine to break it against the **local** API:

1. The Zotero local API answers an `itemKey` filter on the items endpoint with the requested items **plus their children and unrelated neighbours**, and it returns children **first**. Requesting a single parent key `USNDPGWX` returns `[QC7AMKID(child attachment), USNDPGWX(parent), 43SQ2G2A(note)]` in that order:

   ```
   GET http://127.0.0.1:23119/api/users/<id>/items?itemKey=USNDPGWX&limit=10
   -> QC7AMKID type=attachment parent=USNDPGWX
   -> USNDPGWX  type=preprint   parent=-
   -> 43SQ2G2A  type=note       parent=USNDPGWX
   ```

[...]

PR #542, opened 2026-09-13
fix(search): page past note matches in titleCreatorYear mode so child notes can't crowd papers out of the limit

Zhou-Ruichen:
### Problem

On the Web API backend, `zotero_search_items` (and `zotero-cli search`) in the default `titleCreatorYear` mode can silently drop real papers from small result sets.

`_search_with_variants` fetches `/items?q=…&limit=N`, which includes child items. The default `itemType=-attachment` excludes attachments but not notes, and Zotero's server-side quicksearch matches a note's *content* in `titleCreatorYear` mode. The client then drops notes via `_exclude_note_content_matches` — but by then the server has already spent the whole limit budget, so papers ranked just below the cutoff never arrived.

Reproduced on 0.11.0 with a real library (Web API mode): a query whose title contains a common keyword returns 9 items with `--limit 25`, missing papers whose titles contain the keyword. Raising `--limit` makes the missing papers appear — the classic signature of the budget being eaten by items that are filtered out afterwards.

The `collection_key` path of `search_items` has the same flaw: `_paginate(zot.collection_items, …, max_items=limit)` fills the budget with child items before the same note filter runs.

Related: #499 describes the same "children steal limit slots" symptom family on the local API's CLI JSON re-fetch path (`_fetch_projected`); this PR fixes the search path.

### Fix (reworked after review — paging, not `/items/top`)

The first version of this PR queried `/items/top` in `titleCreatorYear` mode. Two review findings rule that endpoint out:

- **Web API:** `/items/top?q=` returns the *parent* of a matching child. dataserver's `search()` only skips the `COALESCE(ITL.topLevelItemID, I.itemID)` projection when `q` is unset, so a child note whose content matches comes back as its parent paper — a different result set that `_exclude_note_content_matches` cannot catch (what arrives is a `journalArticle`, not a note). That is exactly the behaviour a561b84 removed for parity with the SQLite backend.
- **Local API:** `/items/top?itemType=-attachment` still returns child notes (identical to `/items`), so the swap fixes nothing there.

So this keeps `/items` and pages, per the review suggestion:

- `_search_with_variants`: in `titleCreatorYear` mode, page per variant (`start += limit`) until the note-filtered result reaches `limit` or a page comes back short. The count of surviving items is tracked while deduplicating, and the cascade budget is re-checked before each continuation page. `everything` never pages — nothing gets dropped afterwards, so one page is the whole answer.
- `search_items(collection_key=…)`: the same through `_paginate`, which gains a `keep` predicate — only items surviving the note filter count toward `max_items`, and the short-page test stays on the raw page length so a full page of notes is not read as exhaustion. The predicate also does the dropping on this path; the rule itself lives in `_is_note`, shared with `_exclude_note_content_matches`.

Same endpoint, same matching semantics on both backends; the crowding goes away wherever it happens.

### Tests

- New: `titleCreatorYear` pages past a full page of notes (library-wide and collection-scoped); `everything` makes exactly one request on both paths.
- All existing note-filtering tests unchanged in their assertions.

### Verification

Against a real Web API library (read-only key), `titleCreatorYear`, `--limit 25`:

- `deep`: single page → 11 papers; this branch → 13 papers (2 recovered, e.g. *A Deep-Learning Neural Network for Postseismic Deformation…*), 0 notes
- `network`: 12 → 16 papers (4 recovered, all with the keyword in their titles), 0 notes
- `everything`: unchanged — full page, notes kept
- non-crowded queries (total matches < limit) return identical results to before

`pytest tests/` (excluding live): 2728 passed; the 6 failures reproduce identically on a pristine `main` checkout and are unrelated (rate-limit fixtures, webdav test pollution).


54yyyu:
Thanks for the careful writeup, the budget problem is real. I don't think `/items/top` fixes it without changing what matches, though, and I checked two things before saying so.

**1. On the Web API, `/items/top?q=` returns the parent of a matching child note.** From zotero/dataserver `model/Items.inc.php`, `search()`:

```php
$skipITLI = $onlyTopLevel
    // /top?itemKey=[child key]
    && !$itemKeys
    // /top?itemType=annotation
    && empty($params['itemType'])
    // /top?q=[child note title]
    && empty($params['q'])
    ...
if ($onlyTopLevel && !$skipITLI) {
    $itemIDSelector = "COALESCE(ITL.topLevelItemID, I.itemID)";
```

With `q` set, matching runs on the child row (`INo.title LIKE ?`, and the `itemType` filter is `I.itemTypeID NOT IN (...)` on the child too) and then the result is projected onto its top-level parent. So a child note whose title matches now brings back its parent paper. `_exclude_note_content_matches` can't catch that, because what arrives is a `journalArticle`, not a note. That is the exact behaviour a561b84 removed for parity with the SQLite backend. In an arXiv-heavy library, where every paper carries a `Comment: 21 pages, 5 figures, ICML` child note, a titleCreatorYear search for `ICML` or `pages` would return papers whose own title, creators and year don't contain it.

**2. The local API does not behave like the Web API here.** Against a local Zotero 10.0.2, `/items/top?itemType=-attachment` still returns child notes (identical to `/items`), so on local mode this change doesn't remove children from the budget either. I also tried the obvious alternative, `itemType=-attachment || note`: the dataserver negates the whole list, but the local API returns 54 attachments for it, so that isn't portable.

**What I think works on both backends:** keep `/items`, and in titleCreatorYear mode keep paging (`start += limit`) until the note-filtered result reaches `limit` or a page comes back short. Same endpoint, same matching semantics, and the crowding goes away wherever it happens. The collection path can do the same through `_paginate`. Your two `everything`-mode tests stay valid as they are.

Happy to push that onto this branch if you'd rather, since maintainer edits are enabled. Just say which you prefer.

Zhou-Ruichen:
Went with the paging version. Thanks for the review.

Reworked in 93567ed:

- `_search_with_variants` keeps `/items` and pages per variant (`start += limit`) until the note-filtered result reaches `limit` or a page comes back short. The cascade budget is checked before each continuation page.
- The collection path does the same through `_paginate`. `_paginate` now takes a `keep` predicate: only items that pass the note filter count toward `max_items`. The short-page test uses the raw page length; a full page of notes is not read as exhaustion.
- `everything` is untouched on both paths.

Verified against a real Web API library, titleCreatorYear, limit 25: `deep` 11 → 13 papers, `network` 12 → 16. The recovered papers all carry the keyword in their titles; 0 notes in the results. `everything` returns the same full page with notes kept. pytest: 2728 passed, 6 known-unrelated failures.

54yyyu:
Merged in #580, in 0.12.5. Thanks for reworking it to the paging version! The collection branch had moved behind the library backend since you branched, so I carried your `keep` predicate over to `ApiBackend.search_items`. Doing that also showed the collection path had stopped sending `q` at all since 0.12.1, which is fixed in the same release.

Issue #287, opened 2026-05-20
Annotations indexed with placeholder `"No authors listed"` instead of annotation.text — embeddings collapse to one vector, pollutes semantic search


JBTaylor124:
## Summary

When `zotero-mcp update-db --fulltext` indexes a library, every annotation item is stored in ChromaDB with the literal string `"No authors listed"` as its document text — the placeholder rendered from the empty creators field. The annotation's actual highlighted text (from `itemAnnotations.text` in `zotero.sqlite`) is never read. Every annotation therefore receives a bit-identical embedding vector. With thousands of identical-vector entries in the index, `zotero_semantic_search` returns these annotation entries as the top results for every query, blocking real items (books, journal articles, webpages) from ever surfacing.

## Environment

- Package: `zotero-mcp-server[semantic]` installed via `uv tool install`
- Python 3.10
- macOS
- **Zotero 9.0.3** (newer than the README's documented `Zotero 7+` requirement — see the version note at the bottom)
- Embedding model: default (`all-MiniLM-L6-v2`)
- Connection: `ZOTERO_LOCAL=true`, target is a group library (`groupID=<group_id_redacted>`)
- Library size: 2,533 non-attachment items + 3,700 annotations across journalArticles, books, bookSections, webpages with PDF and HTML snapshots
- Index built with `update-db --fulltext --force-rebuild`

## Reproducer — semantic search returns the same items for unrelated queries

Three topically unrelated queries, against the same chroma_db, all return the same eight annotations with identical similarity scores tied within each query:

| Query | All 8 results identical at: |
|---|---|
| `"Nowrasteh culture crutch lazy social science explanations"` | −0.909 |
| `"water rights Phoenix Arizona urban growth desert"` | −0.874 |
| `"indigenous sovereignty self-determination tribal nations"` | −0.832 |

Returned items: `ANNO_A`, `ANNO_B`, `ANNO_C`, `ANNO_D`, `ANNO_E`, `ANNO_F`, `ANNO_G`, `ANNO_H`. All `item_type: "annotation"`, all "Untitled", all "No authors listed" as matched content. The actual relevant items (e.g., the article whose title matches the first query) are never returned.

## Confirmed root cause — direct ChromaDB inspection

Using `chromadb.PersistentClient` against `~/.config/zotero-mcp/chroma_db`:

```
ANNO_A via direct -> chroma_id=ANNO_A
  embedding norm = 1.000000
  document[:160] = 'No authors listed'
[...]

Issue #385, opened 2026-07-14
Is it possible to set reranker model to a different base url?


step21:
Hey, so I set the embedding model to a privately hosted model, with base url and api key. This also works. However when then setting a reranking model it seems it is not possible to set a base url and api key for this, is this accurate? Does it make sense to do this/add this?
Like the embedding model, it would be to have it always available, also when my local ollama for example is not available.

54yyyu:
Accurate, and it's an inconsistency rather than a deliberate design choice.

The embedding side takes a `base_url` and `api_key` through `embedding_config`, so you can point it at anything OpenAI-compatible. The reranker has no equivalent: it's wired to load either a local cross-encoder or whichever provider it defaults to, with no way to give it its own endpoint or credentials.

It makes sense to add, and for the reason you give rather than the obvious one. If your embeddings come from a hosted endpoint that's always up, having the reranker silently depend on a local Ollama being alive means search degrades in a way that's confusing to diagnose from the client side.

Shape I'd take: a `reranker_config` block mirroring `embedding_config` (`base_url`, `api_key`, `model_name`), so the two are configured the same way and neither inherits from the other. Inheriting the embedding endpoint would be convenient but wrong, since reranking and embedding are usually served by different models and often by different providers.

Backlog rather than 0.6.3, which is going out today with a batch of bug fixes. Happy to take a PR if you want it sooner. The reranker lives in `semantic_search.py` and is cached across requests (#365), so the config needs to be part of that cache key or a config change won't take effect until restart.

Issue #118, opened 2026-01-31
Feature request: section-aware full-text indexing (e.g., Abstract + Introduction)


YuChenXin-ZJU:
Hi! Thanks for zotero-mcp — it’s been very helpful.

When running `zotero-mcp update-db --fulltext`, could you add an option to generate embeddings from selected paper sections only (structure-aware indexing), instead of the entire extracted PDF text?

Proposed CLI / config (either works):
- `--section-filter abstract,introduction`
- `fulltext_section_filter: ["abstract", "introduction"]`

Suggested heuristic behavior:
- Abstract: prefer Zotero metadata (e.g., `abstractNote`); fallback to detecting an “Abstract” section in extracted text.
- Introduction: detect “Introduction” heading; extract until the next major section heading (Methods/Results/etc.); fallback to a fixed-length snippet after the heading if needed.

Motivation: reduces content sent to embedding providers, lowers cost and indexing time, while keeping the most useful text for semantic retrieval.

I’m happy to test and share PDFs/logs that help tune the heuristics. Thanks!

54yyyu:
Partial update: the retrieval-quality half of this is superseded by passage chunking (0.6.0, #350), which indexes full text as overlapping passages instead of one truncated vector. The cost-control half (index only Abstract+Introduction to cut embedding spend) is still unimplemented, and chunking actually increases embedded volume, so keeping this open for a section-filter option.

PR #547, opened 2026-09-14
feat(semantic): section labels, per-passage study context, and hosted rerankers

ArneBouten:
Three independent improvements to chunked semantic search, found while
indexing a ~1,900-item library (156k passages) with Gemini embeddings.
Each is a separate commit and they can be taken independently.

### 1. Label each passage with its paper section

With chunking on, a passing mention in someone's Introduction scores as
well as a paper's own finding, and the result gives the reader no way to
tell them apart short of opening the PDF.

Records the section each passage falls in (Abstract, Introduction,
Methods, Results, Discussion, Conclusion, References, Appendix) as chunk
metadata and shows it first in the Location line. Headings are matched by
shape as well as name — own line, optional numbering, nothing after but a
colon — since extracted PDF text keeps line breaks but loses styling.
Unrecognised headings leave the passage unlabelled rather than
mislabelled.

Because it lands in chunk metadata it is filterable through the existing
`filters` argument with no further change:
`{"section": {"$ne": "Introduction"}}` drops secondhand mentions,
`{"section": {"$in": ["Results", "Discussion"]}}` keeps only findings.

### 2. Give every passage its study's context before embedding

`doc_text` is structured text + fulltext, so passage 0 carries title,
creators and abstract while every later passage is a bare window of PDF
prose. A sentence from page 14 is embedded with no signal about which
study it belongs to.

Prepends a compact header (title, year, creators, first 500 chars of the
abstract) to each passage after the first, for the embedded text only.
Offsets, page numbers and section labels still address the original
document, so citation provenance is unchanged. Skipped for annotations
and untitled items (cf. #287). Costs ~130 tokens per passage.

### 3. Allow re-ranking through a hosted endpoint

The local cross-encoder is the only option, and the models worth running
need a GPU to stay interactive — on CPU, bge-reranker-v2-m3 adds seconds
per query, so the practical choice is a weak model or none.

Adds `APIReranker`, interface-compatible with `CrossEncoderReranker`,
selected by a `provider` key. Voyage, OpenRouter, Cohere and Contextual
AI all take a query, documents and a cut-off and answer with
`{index, relevance_score}`, so one client covers them; `base_url` and
`api_key_env` cover anything else. `instruction` is passed only when
configured, for the instruction-following models.

Failures degrade rather than propagate: a missing key or unknown provider
disables re-ranking for the request, a failed call keeps the retriever's
order. The process-wide cache is keyed by provider as well as model.

### Compatibility

No config changes required. Without `chunking.enabled` the first two are
inert; without `reranker.provider` the third falls through to the
existing local cross-encoder unchanged.

### Testing

Developed against 0.12.0 and exercised on a real 1,900-item library.
The reranker client was tested against stubbed responses covering
success, HTTP 500, empty results, missing API key and unknown provider.
I have not run the repo's own test suite; worth a look before merge.


ArneBouten:
Added a fourth commit. A section label ran forward to the next recognised
heading, so in books every chapter inherited the previous one's label.

Measured on a real 156k-passage index, across its 8 longest books: false
`Conclusion` labels dropped from 1,864 to 112, with no correct labels lost.
The cost is about five resets per 200 article passages, all at `FIGURE n`
captions.

Two limits worth stating plainly. An `Appendix` leak survives that the rule
cannot see. And four of those eight books yield no labels at all under any
rule, because their extracted text carries no headings the parser can read.
So this reduces confidently-wrong labels rather than producing correct ones
for books.

54yyyu:
Thanks for this, and for being upfront about the limits on books. The three commits are independent like you said, so I'd like to handle them separately:

**1. Section labels (plus the fourth commit):** yes. It only adds chunk metadata, the embeddings are untouched, and it is what #118 asks for. Please note in the docs that an existing index needs a rebuild before the labels show up, and that books mostly come out unlabelled.

**3. Hosted rerankers:** yes. It is inert without `reranker.provider`, it degrades to the retriever's order on failure, and it covers #385. One request: the docs should say plainly that the query and the passage text are sent to the provider.

**2. Study context before embedding:** I'd like to hold this one. It changes what every passage embeds, so everyone has to re-embed their library to get it (about 130 tokens times every passage, which is real money on the paid providers), and I can't tell yet which way it moves retrieval. The header should help a passage from page 14 find its study, but it also makes all passages of one paper look more alike, which can turn results into "grouped by paper" rather than "matched on content". I have an eval set from earlier retrieval work and will run this as an A/B before deciding. If you have any numbers from your own library, even rough ones, they would be useful here.

Could you split 1 and 3 into their own PRs, rebased on current main? The PR has no tests at the moment and you mentioned the suite had not been run. For 1, a few cases for `_section_offsets` / `_section_at` (numbered headings, an unrecognised heading ending a section, text with no headings) would do. For 3, the stubbed cases you already described (success, HTTP 500, empty result, missing key, unknown provider) are the right set, they just need to be in the repo.

PR #554, opened 2026-09-14
Profiler script

mronkko:
Profiler script as asked in #502


54yyyu:
Thanks for the script! It's merged into the 0.12.3 branch and will land on main with the release.

I made the change we talked about on `maint/0.12.3`. The type filter now uses the `itemTypeID` subquery, and the ranking query skips the index when the library is at least 20% of `items`, so a small group library still uses it. Both plans return the same items in the tests and on my library.

I couldn't reproduce the extra time through `get_library_backend()`. With your script I get 0.8 ms either way, but my library is tiny. If you have a minute, could you run `profile_recent_items.py` on that branch against your library? That would tell us whether the scan choice helps at your size and whether the overhead is still there.

PR #561, opened 2026-09-15
feat: copy items between libraries (#560)

### LINKED ISSUE #560: Feature: Copy items between libraries
Zotero-mcp currently allows adding by DOI and by URL. These are hallucination safe, because they do not allow LLM to write arbitrary data to Zotero. Another hallucination safe operation would be to allow copying between libraries. Zotero API does not expose this feature, but it would be trivial to implement using read and write APIs.

mronkko:
Resolves #560.

### Overview
Adds a new tool `zotero_copy_items_between_libraries` in the `libraries` toolset (enabled by default in `DEFAULT_ON`) to copy items between Zotero libraries (e.g. personal to group, between groups, or from an RSS feed to a user/group library).

### Highlights
- **Hallucination-safe and clean cloning**: Re-creates items via destination library schemas using sanitized payloads; internal fields (`key`, `version`, `dateAdded`, `dateModified`, `parentItem`, `deleted`) are stripped.
- **Library independence**: Consistent with Zotero desktop UI behavior where libraries are independent, cross-library relation links (`dc:relation`/`owl:sameAs`) are omitted and internal relation links from the source are removed. Collections are library-specific and are not copied over; copied items land cleanly in the target library.
- **Child items replication**:
  - Replicates child notes with note HTML content and tags.
  - Replicates child attachments: linked URLs and stored files (PDFs/EPUBs) via local storage lookup and multi-source download fallback.
  - Notes PDF annotations with a `# TODO(human)` comment for future coordinate-mapping support.
- **Deduplication (`if_exists`)**: Supports `'duplicate'` (default), `'file'` (reuses existing match and adds missing tags), and `'skip'`.
- **Direct multi-library querying**: Added `temporary_active_library` context manager in `client.py` and direct `library_id` query support in `local_db.py`.
- **Test coverage**: 26 comprehensive unit tests in `tests/test_copy_items_between_libraries.py` covering validation, sanitization, single & batch copying, child notes/attachments, deduplication, and toolset registry validation.


mronkko:
This is useful feature in a couple of scenarios: When starting new cooperations, this allows automatically sharing the references and PDFs used in the article. It also helps cleaning up. Migrating groups that are discontinued has previously required a drag and drop to my library, followed by deduplication. This is cumbersome. With between-library copy, these tasks can be outsourced to an LLM agent.

54yyyu:
I'd like to take this, the group handover and migration cases make sense and there is no good way to do them today. Three things before it goes in:

1. Could it go in its own opt-in toolset rather than `DEFAULT_ON`? Every default tool is sent on every request for every user, and most people never touch a second library. Something like `ZOTERO_MCP_TOOLSETS=...,libraries-copy` (name is up to you) keeps it one setting away for the people who need it.
2. `if_exists` defaults to `'duplicate'`. With an agent calling this, a retried or repeated call then quietly doubles the items in the target library. I'd default to `'skip'` and leave `'duplicate'` as the explicit choice.
3. The `# TODO(human)` on PDF annotations: either is fine, implementing it or leaving it out, but if it stays out the tool description and the result text should say that annotations are not copied, so nobody assumes they were.

One smaller note: `temporary_active_library` swaps the global active library and relies on the API lock to be safe. That holds today. A short comment there saying it must only be used under `with_zotero_api_lock` would help whoever touches it next.

It also needs a rebase, `write.py` and `local_db.py` both moved in 0.12.5 and 0.13.0. Happy to look again once those are in!

mronkko:
Implentented

- Moved zotero_copy_items_between_libraries out of the default-on libraries toolset into a new opt-in libraries-copy toolset.
- Changed if_exists default from 'duplicate' to 'skip'.
Left PDF annotation copying unimplemented, replacing the internal TODO(human) with a plain statement in the tool description and result text that annotations aren't copied.
- Added the requested comment on temporary_active_library about the API lock requirement.
- Rebased onto upstream/main (46 commits) as requested, plus a docs/tools.md and CHANGELOG entry, and two new tests covering the safer default.

54yyyu:
Thanks for turning this around so quickly! The opt-in toolset, the `skip` default and the annotations wording are all in. Going through it again I found a few things in the copy path that should be fixed before it goes in:

1. **Trashed items get copied as live items.** `get_full_items` and `get_children_of` include trashed rows by default, and the sanitizer strips `deleted`, so a trashed note or attachment is recreated in the target as a normal item. Passing `include_trashed=False` (and skipping children that carry `deleted`) should cover it.
2. **Stored files aren't fetched from the Web API.** `download_attachment_file` is called without a `web_client`, so only local storage and WebDAV are tried. It also runs while the target library is active, so a client built there would point at the wrong library. Could you build the source client inside the source context and pass it in? The `attachment_path_for` call just before it does the same lookup as the download's first step, so it can go.
3. **Failed file copies are invisible.** `attachment_details` is only printed when at least one attachment was copied, so when every file fails the user sees nothing. Same for the note that annotations aren't copied. Both should always show.
4. **ISBN dedup misses hyphenated ISBNs.** `find_existing_items` expects `_normalize_isbn` output but gets the raw value. Worth also saying in the description that items with no DOI, ISBN or URL are always copied, even with `skip`.
5. **Hybrid mode, personal library.** In local mode a `user` target resolves to `"0"`, and when writes go through the web client that becomes `/users/0`.
6. **Some reuse.** `_resolve_library_spec` overlaps with `_parse_library_id_param` plus `validate_library_switch`, and the source read could be `with temporary_active_library(src): get_library_backend()` instead of the hand-rolled SQLite/API branch. The results dict built in about eight places could be one small helper. The tool is long, so anything that shrinks it helps.
7. The description has a garbled sentence ("Defaults to active library if copy_notes...") and "Hallucination-safe" can go.

Tests for the trashed child, a failed download, a hyphenated ISBN and the SQLite source path would pin these down. Happy to look again once they're in!

mronkko:
All seven are addressed in 115598a, rebased onto 0.13.1. Two of them turned out a bit different from how the review described them:

1. **Trash.** Trashed children were half right: `get_children_of` already excludes them in SQL (`del.itemID IS NULL`). The trashed *parent* was the real hole: `get_full_items` and the API's `item()` fallback both return it. So the check is now on the `deleted` flag. A trashed source item is reported as "in the trash", and any child that carries `deleted` is skipped on either backend.
2. **Web API downloads.** The source clients are built in the source context before the target becomes active, and passed to `download_attachment_file`. The `attachment_path_for` call is gone.
3. **Failures.** Failed note and attachment copies are listed on every copied item, including when nothing succeeded. The annotations note is shown whenever attachments are copied.
4. **ISBN.** It's normalized before matching, taking the first valid ISBN from a multi-value field. The description now says items with no DOI, ISBN or URL are always copied.
5. **Hybrid `/users/0`.** I fixed this in `apply_library_override`, not in the tool. When the override is the local `"0"` and the client is a user-scoped web client, it keeps the client's userID. The same bug hit `zotero_switch_library` → personal library → any write in hybrid mode, so that path is fixed too (CHANGELOG under Fixed).
6. **Reuse.** `_resolve_library_spec` is gone, replaced by `_parse_library_id_param` + `validate_library_switch`. User and group sources read through `temporary_active_library(src)` + `get_library_backend()`. **Feeds keep a direct read by libraryID:** `get_active_group_id()` returns 0 for a feed, so the generic backend would read the personal library and report every key as not found. `SqliteBackend` would need a feed scope first, which seems better as its own PR. Results come from one helper, and the tool body is split into `_copy_one` / `_copy_attachment` / `_format_copy_results`.
7. Description rewritten, "Hallucination-safe" removed.

Tests added: trashed child, trashed parent, a failed download, the download using the source-scoped client, a hyphenated multi-ISBN field, the SQLite source path scoped to the source group, the feed source read by libraryID, and the `/users/0` override.

While checking item 2 I found a related bug outside this PR: with WebDAV configured and web-API writes, an upload into a **group** library goes to the personal WebDAV. `zotero_add_item` / `zotero_attach_file` then leave the group attachment with no file, because WebDAV-first sends the file only there. The storage helpers don't take a library argument, so they can't tell groups apart. Filed as #591.

54yyyu:
Thanks, all seven are in and I checked each one. The fix in `apply_library_override` is the right place for the `/users/0` bug, and the feed read by libraryID is well reasoned. There are no conflicts with main and CI is green.

One thing left before merging: **an exception partway through a batch loses the report after items were already written.** In `_copy_one`, only `create_items` is inside the try/except. These three calls are unguarded:
- `find_existing_items`, which deliberately re-raises `TooManyRequestsError`/`TooManyRetriesError`
- `_update_item_with_version_retry` on `if_exists='file'`
- `item_template_for` in the linked-URL attachment branch

So if item 1 is created and item 2's dedup search hits a rate limit, the tool raises: item 1 is already in the target, but the caller gets no report. Items with no DOI/ISBN/URL are always copied even with `skip`, so the natural retry duplicates them.

Could you catch exceptions per item around the `_copy_one` call and turn each into `_copy_result(key, title, error=...)`? On a rate limit it is fine to stop the batch, but return the partial report, marking the remaining keys as not attempted. A test would pin it down: second item's lookup raises, first item still shows as copied.

A minor point, optional: with `target_library_type='user'`, any id maps to your own personal library, so a wrong id silently writes there rather than erroring. Fine to leave as it is if that's intended for web mode.

---

PR #565, opened 2026-09-17
fix: ChromaDB can call embed_query on our embedding functions without a TypeError

josk0:
## The bug

ChromaDB defines this method on every embedding function (since chromadb 1.1.0):

```python
def embed_query(self, input) -> Embeddings   # a list of texts in, a list of vectors out
```

Our embedding functions (OpenAI, Gemini, Ollama, HuggingFace) define a method with the same name but a different signature:

```python
def embed_query(self, text: str) -> list[float]   # one text in, one vector out
```

ChromaDB calls the method with a keyword argument: `embed_query(input=[...])`. Our method has no `input` parameter, so the call fails:

```
TypeError: RemoteEmbeddingFunction.embed_query() got an unexpected keyword argument 'input'
```

## When it happens

Normally the bug is hidden. `ChromaClient.search` embeds the query itself and never lets ChromaDB call `embed_query`.

The bug appears when both of these are true:

1. The index was built with one of our embedding functions.
2. The client's embedding model is set to `default` (ChromaDB's built-in embedding function). This happens when the config file is not found, or when code calls `create_chroma_client()` without a config path.

In this case ChromaDB opens the existing collection and rebuilds our embedding function from the configuration stored in that collection. Then it calls `embed_query(input=...)` on that function. The search fails with the `TypeError` above.

## The fix

- The method for one text is now named `embed_query_text`. Its behaviour did not change.
- `embed_query(input)` now follows ChromaDB's signature. It takes a list of texts and returns a list of vectors. It calls `embed_query_text` once for each text, so there is one request for each query text. `ChromaClient.search` already worked this way.
- If `embed_query` gets a single string, it raises a `TypeError` that names `embed_query_text`. Without this check, Python would iterate over the string and embed every character separately.
- `embed_query` does not call `super().embed_query`. That method does not exist in chromadb 1.0.x, and we support `chromadb>=1.0.0`.
- `ChromaClient.search` now calls `embed_query_text`. It also uses the existing `CUSTOM_EMBEDDING_FUNCTIONS` tuple instead of a second hard-coded list of our embedding functions. The two lists have the same four classes.

## Behaviour changes

- Code that calls `ef.embed_query("some text")` with a string now gets a `TypeError`. It must call `embed_query_text`. I updated all callers in this repository, including `tests/live/`.
- In the situation described above, search now works. If the index was built with OpenAI, Gemini or Ollama, the client now sends requests to that provider, and these requests can cost money. If the API key is missing, the user now gets the provider's authentication error instead of the `TypeError`. ChromaDB uses the stored embedding function for documents and for queries, so all vectors in the index come from the same model.

## Tests

- `tests/test_embed_query_chromadb_regression.py` reproduces the bug with a real ChromaDB collection on disk. On `main`, two of its tests fail with the `TypeError` above. It also checks that the collection and its stored embedding-function configuration are unchanged after the client opens it.
- `tests/test_embed_query_contract.py` checks the new behaviour for each of our embedding functions: list input, the string check, and the Gemini query path for v1 and v2 models. One test checks that the parameter of `embed_query` is named `input` in each of them, so a new embedding function cannot bring the bug back.

The full test suite passes.

https://claude.ai/code/session_016U1NCekFtqCJb4sAmdPeQz


54yyyu:
Merged in #580, in 0.12.5. Thanks for both, and for writing up the repair step for people who already used `--limit`, that went into the changelog. On #565 I removed a few tests that overlapped with existing coverage (the Gemini task type and prefix ones, and the positional/empty input cases), the rest is unchanged.

Issue #458, opened 2026-08-17
`zotero_find_related_papers`: citations direction always returns zero, plus four more silent under-reports


Aboottogo:
**Version:** zotero-mcp-server 0.9.1 (uv tool install, `[all]` extra)
**Python:** 3.12.7 · **OS:** macOS 26.4.1 · **Transport:** stdio
**Toolsets:** `scite,libraries,search-admin,pdf-geometry,discovery`

All five issues below are in `zotero_mcp/tools/discovery.py`. They share a failure mode: each one silently under-reports, and none surfaces an error, so the output is indistinguishable from a correct answer. `zotero_library_coverage` is unaffected — I tested it against a 125-item library and it was exact.

I have a patch ready and will open a PR if you would like one.

---

### What happens

```
zotero_find_related_papers(identifier='10.1007/s11142-021-09582-z', direction='citations', limit=20)
-> "Found 0 citations, 0 already in library."
```

OpenAlex reports 33 citing works for that DOI. No error is raised, and the output is identical in shape to a paper that genuinely has no citations.

### Root cause

OpenAlex no longer returns `cited_by_api_url` on work records. The guard at `discovery.py:203-210` has no `else`, so the entire branch is skipped:

```python
if want_cites:
    cited_by_url = work.get("cited_by_api_url")
    if cited_by_url:          # None today -> whole block skipped
        ...
```

`citations` stays `[]`, and line 226 renders `f"{len(citations)} citations"` as "0 citations".

### Reproduction

```
curl -s "https://api.openalex.org/works/https://doi.org/10.1007/s11142-021-09582-z?mailto=you@example.com" \
  | python3 -c "import json,sys; d=json.load(sys.stdin); print('cited_by_count:', d['cited_by_count']); print('cited_by_api_url:', d.get('cited_by_api_url'))"
```

```
cited_by_count: 33
cited_by_api_url: None
```

I checked four DOIs across four publishers, including one with 4,021 citations. The field is absent on all of them.

### Fix

Build the query from the work's own ID when the field is missing:

```python
work_id = _short_id(work.get("id"))
cited_by_url = f"{_OPENALEX_BASE}/works?filter=cites:{work_id}"
```

Verified working: `filter=cites:W3160028372` returns `"count": 33`. Note `_openalex_get` passes params via `requests.get(url, params=...)`, which merges correctly into a URL that already carries a query string, so no other change is needed.

---

### What happens

```
[...]

PR #566, opened 2026-09-19
fix(discovery): look up citing works via the cites: filter

feiiiiii5:
fix(discovery): citing works always came back empty after OpenAlex removed `cited_by_api_url`

## The problem

`zotero_find_related_papers` reports **0 citations for every paper**, because the branch that fetches citing works is keyed on a field OpenAlex no longer sends:

```python
cited_by_url = work.get("cited_by_api_url")
if cited_by_url:                      # None today -> the whole block is skipped
    data = _openalex_get(cited_by_url, ...)
```

`citations` stays `[]`, and the summary line renders `Found 0 citations.` There is no error, so the output is indistinguishable from a paper that genuinely has never been cited — which is the worst possible failure mode for a literature-review tool, since a "no citations" answer reads as evidence about the paper.

Reproduced against the live API on `8778e62`, two publishers, one very-cited and one lightly-cited work:

```
$ curl -s "https://api.openalex.org/works/https://doi.org/10.1007/s11142-021-09582-z?mailto=..."
  cited_by_count: 36    cited_by_api_url present: False
$ curl -s "https://api.openalex.org/works/https://doi.org/10.1038/nature12373?mailto=..."
  cited_by_count: 1987  cited_by_api_url present: False
```

## The fix

When the field is absent, build the same query from the work's own OpenAlex id, which is what the `cites:` filter expects. The old path is kept so a response that does carry `cited_by_api_url` still uses it.

```python
if not cited_by_url:
    work_id = _short_id(work.get("id"))
    if work_id:
        cited_by_url = f"{_OPENALEX_BASE}/works"
        cite_params["filter"] = f"cites:{work_id}"
```

Verified against the live API: `GET /works?filter=cites:W3160028372` returns `meta.count = 36`, matching the `cited_by_count` of the same record, so the fallback finds exactly the works the removed URL used to point at.

This is item 1 of #458, and the query shape is the one @Aboottogo proposed there. Items 2-5 of that issue (URL-form DOI matching, `limit` erosion on dead referenced works, and the failed-lookup-reads-as-empty-result problem) are separate root causes and are untouched here.

## Tests

`tests/test_discovery.py` pins the citations behaviour, but every fixture there synthesises a `cited_by_api_url` — which is why the suite stayed green while the tool was broken in the field. The new test uses a payload shaped like what OpenAlex actually returns now, and also asserts the guard: with no `id` to name the work by, no request is invented and the answer stays `0 citations`.

Same command, base versus head:

| tree | `pytest tests/test_discovery.py -q` |
|---|---|
| `8778e62` + new test | `1 failed, 10 passed` — fails with `AssertionError: no citations lookup was attempted` |
| + this patch | `11 passed` |

Full suite (`pytest --ignore=tests/test_lifespan.py -q`), same environment:

| tree | result |
|---|---|
| unmodified `8778e62` | `4 failed, 3009 passed, 84 skipped` |
| with this patch | `4 failed, 3010 passed, 84 skipped` |

None of those four are in discovery: they are `test_local_write_client.py::test_carries_the_key_and_server_id`, two `test_rate_limit_guard.py` tests and `test_webdav.py::test_download_attachment_file_falls_back_to_webdav`. Two of the families are pyzotero-version sensitive: keeping the same venv and the same checkout and installing `pyzotero==1.15.1` instead of `1.15.2` takes `test_rate_limit_guard.py` + `test_local_write_client.py` from 6 failed to all passing, and the webdav one passes in CI. That version bisect is written up in #567, since it turns the CI matrix red for every branch, not just this one. The `1 failed, 10 passed` -> `11 passed` evidence above lives inside `test_discovery.py` and is unaffected by it.

Lint: `ruff check src/zotero_mcp/tools/discovery.py tests/test_discovery.py` passes, and `ruff format --check` is clean on the test file. On `discovery.py` it reports two pre-existing comprehensions in `zotero_library_coverage` (~lines 302-313) that would be reformatted; that is reported identically on an unmodified `8778e62`, so I left formatting alone rather than mixing unrelated reformatting into this diff.

## Scope and limits

- One code path changed: the citations lookup. References, library-membership flagging and the coverage tool are untouched.
- `per-page` is still `min(limit, 50)` and the results are still sorted by citation count locally, so the most-cited-first promise holds only over the first page — that is item 3's territory and deliberately left alone here.


54yyyu:
Merged in #580, in 0.12.5. I checked the OpenAlex change against the live API before merging and `cited_by_api_url` is indeed gone. Thanks for splitting #458 into separate root causes, that made these easy to review!

PR #569, opened 2026-09-19
Fix test suite under pyzotero 1.15.2 (127.0.0.1 local endpoint, per-request headers)

jianxing-chen:
## Problem

The `Build and publish to PyPI` test matrix is currently red on open PRs (e.g. all five `test (...)` jobs failing on e731262d and 024a94c2, while the same workflow passed on 09-17). The breakage window matches pyzotero 1.15.2's release on 2026-09-18, and `pyproject.toml` keeps the range open (`pyzotero>=1.14.0`), so CI now installs 1.15.2.

Reproduced on a pristine checkout of `main` @ 8778e62 with pyzotero 1.15.2:

```
FAILED tests/test_local_write_client.py::TestGetLocalWriteClient::test_carries_the_key_and_server_id
FAILED tests/test_rate_limit_guard.py::TestUpstreamRateLimitHandling::test_transient_429_is_retried_and_recovers
FAILED tests/test_rate_limit_guard.py::TestDedupUnderThrottling::test_transient_throttling_still_finds_the_existing_item
```

plus, intermittently, the rest of the 429 family — why below.

## Root cause

Two intentional changes in pyzotero 1.15.2:

1. [Use 127.0.0.1 for the local API endpoint](https://github.com/urschrei/pyzotero/commit/f9b1f81ed0) — a local client's `endpoint` is now `http://127.0.0.1:23119/api`; our suite pins `localhost`.

2. [Send the default headers with each request](https://github.com/urschrei/pyzotero/commit/8b90f8054d) — requests now go through a new `Zotero._send()` that calls `client.request(method, url, **kwargs)` instead of `client.get(url, ...)`. The rate-limit fixture mocks `client.get`, so under 1.15.2 the mock is **silently bypassed**: those tests issue real requests to `api.zotero.org`, fail with `403 Invalid key` / SSL errors depending on the runner's network, and the suite becomes network-dependent — which is why the failure list looks flaky.

## Fix

- `tests/test_rate_limit_guard.py`: the replay mock now patches `client.request` as well as `client.get` — whichever the installed pyzotero speaks; only one is ever called, and both replay the same sequence.
- `tests/test_local_write_client.py`: accept either endpoint host, since the floor is pyzotero 1.14, which still reports `localhost`.
- `src/zotero_mcp/client.py` (`probe_local_server_id`): probe `127.0.0.1:23119` instead of `localhost`, mirroring pyzotero's rationale — loopback by IP keeps the probe off HTTP proxies configured to intercept `localhost`.

## Verification

- Full suite under pyzotero **1.15.2**: `3318 passed, 6 skipped` (previously 3 deterministic failures plus the intermittent live-network ones).
- The touched test files also pass under **1.15.1**, so the supported range keeps working.

pyzotero adapted their own tests for both changes in the same release; this mirrors that on our side.


54yyyu:
Thanks for this! #577 landed first for the test side, with the replay moved down to a mock transport so it holds for both dispatch paths. Your `127.0.0.1` probe fix is the part that affects real users, so I took that commit into #580 with you as co-author, and it is out in 0.12.5. Closing this one since both halves are in.

PR #574, opened 2026-09-20
fix: resolve Zotero base fields per item type, not by hardcoded fieldID (#570)

### LINKED ISSUE #570: SQLite backend: keyword search returns 0 results for titles the API backend finds (Zotero 10.0.3, v0.12.4)
**Environment:** zotero-mcp-server 0.12.4 (uv tool, [all]), macOS Apple Silicon, Zotero 10.0.3, ZOTERO_LOCAL=true, personal library ~8,000 items, imported attachments.

I hit this while checking why a known paper didn't show up in title search.

**Repro:**

ZOTERO_BACKEND=sqlite zotero-cli search "Rethinking prognosis" --limit 3
# 0 results, falls back to semantic search

ZOTERO_BACKEND=api zotero-cli search "Rethinking prognosis" --limit 3
# finds 4UYS3M8I (Prosperini et al. 2026)

ZOTERO_BACKEND=sqlite zotero-cli search "The dual nature of neuroticism" --limit 5
# 0 results; simplified-to-"The" fallback lists items with title rendered as "Untitled" or as the item URL (e.g. XZH5TRDS, RLTLL7RY)

ZOTERO_BACKEND=sqlite zotero-cli get metadata XZH5TRDS
# title correct: "Evidence of early cortical atrophy in MS: relevance to white matter changes and disability"

**Observations:**
- Single-item reads on the SQLite path return the correct title.
- Looks like the search query isn't matching on title, and the result renderer reads an empty title for the same rows.
- Not WAL staleness (#536): items are years old; Zotero has been restarted since.
- DOI, citekey and creator conditions work.

I'm using ZOTERO_BACKEND=api as a workaround.

mronkko:
Fixes #570.

## The bug

Zotero maps the base fields `title`, `date` and `publicationTitle` onto a
type-specific column for some item types — a case's title is `caseName`, a
statute's is `nameOfAct`, an email's is `subject`; a case's date is
`dateDecided`, a webpage's publicationTitle is `websiteTitle`.

Several SQL queries in the SQLite backend hardcoded a fixed `fieldID` (or a
fixed field name) for these columns. Items of those types never matched a
title search, rendered as "Untitled", sorted wrong, and indexed into semantic
search with an empty title *and* an empty abstract — even though a single-item
read rendered them correctly, because that path fetches every field
generically.

## Root cause, and the rule this establishes

Zotero's own client never picks a fieldID up front. It loads `itemData`
generically and resolves a base field at read time via
`Zotero.ItemFields.getFieldIDFromTypeAndBase`, whose effective rule is
`override || baseFieldID`. That rule holds at two independent layers:

- `_baseTypeFields` (`itemFields.js:524-557`) is built from a *cartesian*
  `itemTypesCombined LEFT JOIN fieldsCombined`, so every `(type, baseField)`
  cell is filled — the override, else the base fieldID itself when valid for
  the type, else `false`.
- `getField` turns that `false` straight back into the base fieldID
  (`item.js:279-281`).

This PR mirrors that rule in SQL against `baseFieldMappingsCombined`, the same
table that cache is built from, as `COALESCE(map.fieldID, f.fieldID)`.

**The rule this establishes: never a literal fieldID in SQL.** Base-mapped
fields go through a resolver; fields with no mapping resolve by name.

## What changed

Four helpers in `local_db.py`, sharing one documented rationale that cites the
Zotero sources above:

- `_base_field_resolved_join(alias, base_field, item_alias="i")` — the join
  form, for projections. Now takes the outer `items` alias, so a *parent's*
  title can resolve against `pi`/`gpi`.
- `_base_field_resolved_subquery(...)` — the correlated-scalar-subquery form,
  for WHERE conditions. A `value_expr` parameter lets the three date variants
  share it.
- `_plain_field_join` / `_plain_field_subquery` — resolve by name for
  `abstractNote`, `DOI`, `extra`, `url` and an attachment's own title, which
  provably have **no** base-field mapping, so a mapping lookup could only ever
  miss.

Call sites fixed:

- **`_SIMPLE_FIELD_SQL` title/publicationTitle and the date/year condition
  SQL.** This was the sharpest defect: `advanced_search_sql` hydrated through a
  base-resolved `SELECT` while matching through a non-resolved `WHERE`, so one
  statement contradicted itself — it could return a row whose displayed title
  was exactly the string its own condition had just failed to match on.
- `get_items_with_text` (the semantic-search corpus builder).
- The note and annotation parent-title lookups — a note filed under a case
  rendered `parent_title: "Unknown"`.
- `get_feed_items` and `get_attachment_by_key`.

**The pyzotero path is fixed too** (`tools/search.py`). It did no base
resolution either (`_extract_values` read `data[field]` raw), and the SQLite
backend is opt-in — so fixing only the SQL would have left the default path
broken and silently diverged the two backends. `_extract_values` now routes the
field through `schema.resolve_field` on the `date`, `year` and generic
branches.

## Testing

Offline suite: **3342 passed, 0 failed** (up from a 3314 baseline).

The shared parity corpus gains a `case` item whose title lives in `caseName`,
materialised into both the SQL fixture and the API fixture from one
declaration. This matters: the corpus previously contained no base-field-mapped
item type at all, so a SQL-vs-pyzotero divergence here would have gone
unnoticed — and a parity assertion passes just as happily when *both* backends
find nothing. The new cases were checked to be non-vacuous (the item is
genuinely returned by both paths, and the SQL backend serves the query rather
than falling back).

Verified end-to-end against a real library's `statute` and `email` items, and
by three live tests that discover base-field-mapped types generically rather
than hardcoding item types.

`tests/live/test_search_backend_parity.py::test_search_items_free_text_parity`
fails, but is **pre-existing**: it fails identically with the working tree
stashed, and independently — every divergent key is a
`journalArticle`/`conferencePaper`, while the test library holds only 4
base-field-mapped items in total, so base-field resolution cannot be the cause.

## Deliberately out of scope

- `item_display_title` still lacks Zotero's display-title special cases
  (`letter`/`interview` creator-derived titles, `case`'s `caseName (reporter)`
  suffix, annotation titles). Those render "Untitled" today and deserve their
  own issue.
- `row_to_api_item` writes hydrated values under the **base** names, whereas
  the Web API returns `caseName`/`dateDecided`, so the two backends differ in
  item *shape* for remapped types. Harmless today — both display helpers fall
  back to the plain key — and documented in the corpus fixture, which is why
  the corpus case item deliberately carries no date.


54yyyu:
Merged in #580 and released in 0.12.5, thanks for tracking this down to how the client resolves fields!

One change on top: the `LEFT JOIN fields x_f ON x_f.fieldName = '...'` form scans `fields` for every outer row, since there is no index on `fieldName`. On my library keyword search went from 3.6 ms to 9.5 ms. I replaced it with an uncorrelated `(SELECT fieldID FROM fields WHERE fieldName = '...')`, which SQLite evaluates once per statement: 1.3 ms, same rows. I also shortened the rationale comment to the rule itself, the full reasoning is in your PR description for anyone who wants it.

PR #576, opened 2026-09-20
fix: compose the display titles Zotero composes, instead of reading a field (#575)

### LINKED ISSUE #575: Annotations render as "Untitled": item_display_title reads a field where Zotero composes one
## Summary

`utils.item_display_title` resolves a title by *reading a field*. For four item
types Zotero doesn't store a title at all — it **composes** one. Those items
fall through to the final `or "Untitled"`.

The material case is annotations. In my library that is **461 items**, every one
of which renders `## Untitled`.

## Reproduction

Any path that formats an annotation as a top-level item. The
`zotero://items/{item_key}` MCP resource is the simplest: it calls
`format_item_result` on whatever key it is given, with no type filter
(`resources.py:55`).

```python
from zotero_mcp.local_db import LocalZoteroReader
from zotero_mcp.utils import item_display_title, format_item_result

rd = LocalZoteroReader(db_path="~/Zotero/zotero.sqlite")
d = rd.get_full_items(["ZR495LQ2"], group_id=None)["ZR495LQ2"]["data"]

d["annotationType"]     # 'highlight'
d["annotationText"]     # 'Sample 7: employees evaluated reward expectations …'
d["annotationComment"]  # '-if you ask a question this wa…'

item_display_title(d)                                  # 'Untitled'
format_item_result({"key": "ZR495LQ2", "data": d})[0]  # '## Untitled'
```

The data is present and correct throughout — only the presentation loses it.
Zotero's own client shows
`“Sample 7: employees evaluated reward expectations…” -if you ask a question…`

`collection_items_resource` filters out `attachment` but not `annotation`, so
annotations reach the same heading there too.

## Root cause

`Zotero.Item.prototype.updateDisplayTitle`
(`chrome/content/zotero/xpcom/data/item.js:903-1035`) composes a display title
for four types. `item_display_title` implements none of them except notes:

| type | Zotero renders | we render |
|---|---|---|
| `annotation` (highlight/underline) | `“{text…}” {comment…}`, each component capped at 50 chars | `Untitled` |
| `annotation` (image/ink/text, empty) | `Image Annotation` / `Ink Annotation` | `Untitled` |
| `case` (common law) | `{caseName} ({reporter})`, else `({court})` | bare `caseName` |
| `letter` / `interview`, untitled | `[Letter to Thoreau]` / `[Interview by …]` | `Untitled` |

Zotero's exact strings, for reference
(`chrome/locale/en-US/zotero/zotero.properties:378-385`,
`chrome/locale/en-US/zotero/reader.ftl:20-27`):

```
pane.items.letter.oneParticipant     = Letter to %S
pane.items.letter.twoParticipants    = Letter to %S and %S
pane.items.letter.threeParticipants  = Letter to %S, %S, and %S
pane.items.letter.manyParticipants   = Letter to %S et al.
pane.items.interview.*               = Interview by …
reader-highlight-annotation          = Highlight annotation
reader-underline-annotation          = Underline annotation
reader-note-annotation               = Note Annotation
reader-text-annotation               = Text Annotation
reader-image-annotation              = Image Annotation
reader-ink-annotation                = Ink Annotation
```

## A second, prerequisite bug

`local_db._ANNOTATION_TYPES` is `{1: highlight, 2: note, 3: image, 4: ink,
5: underline}`. Zotero added **`ANNOTATION_TYPE_TEXT = 6`**
(`chrome/content/zotero/xpcom/annotations.js:36`, named `'text'` at
`data/items.js:538`). Type 6 therefore maps to `""` today.

This is already visible without any display-title change:

- `tools/retrieval.py:869` renders `f"  - [{key}] {ann_type}: {text}..."` →
  `"  - [KEY] : ..."`
- `cli_json.py:184` emits `"type": ""`

My library has 3 such annotations, all comment-only (`text` empty,
`comment` 1/33/16 chars). Any type-keyed display-title logic has to handle
them, so this wants fixing first.

## Why it matters

Same argument as #452: the search layer finds these items correctly, so an
affected item is **findable but unidentifiable**. An agent reading a list of
annotations sees a column of `## Untitled` and has to spend a second call per
item to discover what each one says — against the near-zero cost of printing
the text already in hand.

## Scope note

The `case` civil-law form (a case with no `caseName`, rendered
`[court, date, author]`) depends on `Zotero.Date.multipartToSQL`, whose
implementation is not present in my Zotero checkout, so I have not verified its
exact date format. That sub-branch is rare and I would leave it out rather than
guess. Likewise, `letter`/`interview` and `case` are **conformance** work — my
library contains no `case` at all, and its one letter and two interviews all
have titles, so only the annotation half is backed by an observable symptom
here.

mronkko:
Fixes #575.

## The bug

`utils.item_display_title` found a title by *reading a field*. For four item
types Zotero stores none — it **composes** one, in
`Zotero.Item.prototype.updateDisplayTitle`
(`chrome/content/zotero/xpcom/data/item.js:903-1035`). Those items fell
through to the final `or "Untitled"`.

The material case is annotations: **461** in my library, every one rendering
`## Untitled` through the `zotero://items/{item_key}` resource, which formats
whatever key it is given with no type filter (`resources.py:55`).
`collection_items_resource` filters `attachment` but not `annotation`, so they
reach the same heading there too.

```python
d = rd.get_full_items(["ZR495LQ2"], group_id=None)["ZR495LQ2"]["data"]
d["annotationText"]    # 'Sample 7: employees evaluated reward expectations …'
d["annotationComment"] # '-if you ask a question this wa…'
item_display_title(d)  # 'Untitled'   ← the text was right there
```

Same shape as #452: the search layer finds these items correctly, so an
affected item is **findable but unidentifiable**, and recovering the name costs
an extra call per item.

## What changed

| type | now renders |
|---|---|
| `annotation` highlight/underline | `“{text…}” {comment…}`, each component capped at 50 chars |
| `annotation` image/ink/text/note, empty | `Image Annotation`, `Ink Annotation`, … |
| `case` | `Marbury v. Madison (5 U.S. 137)` — reporter, else court |
| `letter`/`interview`, untitled | `[Letter to Thoreau]`, `[Interview by Terkel]` |

Three helpers in `utils.py` — `annotation_title`, `_case_title`,
`_participant_title` — each mirroring the corresponding branch of
`updateDisplayTitle`.

Every rendered string is Zotero's own, verbatim, including its inconsistent
capitalisation (`Highlight annotation` but `Note Annotation`), from
`zotero.properties:378-385` and `reader.ftl:20-27`.

**Two behaviours are mirrored deliberately rather than improved**, because
conformance with the client is the point:

- an empty highlight renders as `“”` — Zotero closes the quotes *before*
  testing whether the title is empty (`item.js:1013-1021`), so it never
  reaches the type-name fallback. Unreachable in practice: 0 of 445 highlights
  in my library have empty text.
- the authors Zotero collects for a letter are never used; only the recipients
  reach the string.

## A prerequisite bug, also fixed

`_ANNOTATION_TYPES` stopped at 5, but Zotero added `ANNOTATION_TYPE_TEXT = 6`
(`annotations.js:36`, named `'text'` in `data/items.js:538`). Those annotations
carried `"annotationType": ""`, which was **already user-visible** before any
display-title change — `- [KEY] : ...` from `zotero_get_item_children`, and an
empty `"type"` in the CLI's JSON. A display title keyed on the type needs this
right first. My library has 3 such items, all comment-only.

## One existing expectation changes, with intent

#452's case fixture carries a `court` and no `reporter`, so it now renders
`DSG Retail Limited v The Information Commissioner (Upper Tribunal)` — which is
what the Zotero client shows for that item. #452's own point, that the name
comes from `caseName` rather than `title`, is untouched and still asserted by
the surrounding tests.

## Backend shape note

`_case_title` consults the base `title` key as well as the resolved `caseName`.
The SQLite backend hydrates a case's name under the **base** key while the web
API returns `caseName`, so reading only the resolved key would silently drop
the qualifier for every SQL-served case. There is a test for that shape.

## Not implemented

The civil-law form for a case with no name at all (`[court, date, author]`).
It depends on `Zotero.Date.multipartToSQL`, whose implementation is not present
in my Zotero checkout, so I could not verify its date format — and guessing one
is worse than leaving a rare branch alone.

`letter`/`interview` and `case` are **conformance** work: my library contains no
`case` at all, and its one letter and two interviews all have titles. Only the
annotation half is backed by an observable symptom here.

## Testing

Offline suite **3349 passed, 0 failed**. New `tests/test_composed_display_titles.py`
(38 cases) covers each branch, the 50-character clipping boundary (`>`, not
`>=`), HTML reduction, the participant count forms, and guards that ordinary
articles, notes, attachments and #452/#570 base-field resolution are unchanged.

Verified against my library's real annotations — highlight-with-comment, empty
ink, empty image, and the comment-only type-6 items — all of which previously
rendered `## Untitled`.


54yyyu:
Merged in #580, in 0.12.5. Two changes to mention:

- `annotationText` and `annotationComment` are plain text, and running them through `html_to_text` removed anything between `<` and `>`, so a highlight like `p < 0.05 and n > 30` lost its middle. I dropped the HTML pass and collapse whitespace instead, so a multi-line highlight stays on one heading line.
- I left out the letter/interview participant titles for now, since nobody has reported those rendering badly. Happy to take them as a separate small PR if you run into it!

Annotations, cases and the type 6 fix are all in.

PR #577, opened 2026-09-20
test: track pyzotero 1.15.2's request dispatch and local host

### LINKED ISSUE #567: CI is red for every PR since pyzotero 1.15.2: 429 handling and the local API base URL both changed
## What happens

The `test (ubuntu / macos / windows, ...)` matrix now fails on branches that touch nothing related. On #566 all five jobs failed at `Run tests` with 6 failures, none of them in the files that PR edits:

```
FAILED tests/test_rate_limit_guard.py::TestUpstreamRateLimitHandling::test_transient_429_is_retried_and_recovers - pyzotero.errors.UserNotAuthorisedError:
FAILED tests/test_rate_limit_guard.py::TestUpstreamRateLimitHandling::test_persistent_throttling_raises_rather_than_returning_bytes - pyzotero.errors.UserNotAuthorisedError:
FAILED tests/test_rate_limit_guard.py::TestUpstreamRateLimitHandling::test_429_without_backoff_header_raises - pyzotero.errors.UserNotAuthorisedError:
FAILED tests/test_rate_limit_guard.py::TestDedupUnderThrottling::test_throttled_search_does_not_read_as_no_match - Failed: DID NOT RAISE TooManyRetriesError
FAILED tests/test_rate_limit_guard.py::TestDedupUnderThrottling::test_transient_throttling_still_finds_the_existing_item - AssertionError: assert [] == ['EXIST001']
FAILED tests/test_local_write_client.py::TestGetLocalWriteClient::test_carries_the_key_and_server_id - AssertionError: assert 'http://127.0.0.1:23119/api' == 'http://localhost:23119/api'
6 failed, 3257 passed, 134 skipped
```

`Build and publish to PyPI` on `main` was green at `8778e62e3` (2026-09-15), and #564 / #565 were green on 2026-09-17.

## Cause

`pyzotero 1.15.2` reached PyPI at **2026-09-18T21:37:51Z** — after those green runs, before the red ones. The test job installs it unpinned (`.github/workflows/release.yml:41`, `python -m pip install pytest pytest-timeout "fastmcp<4" pyzotero ...`), and the requirement in `pyproject.toml:35` is `"pyzotero>=1.14.0"`, so every run since then resolves the new release. Two independent symptoms show up, both in the area this repo already had to repair for the 1.15 httpx -> httpx2 migration (#511, #512):

1. A mocked `429` with a `Backoff` header no longer reaches pyzotero's retry path: the tests get `UserNotAuthorisedError` where they expect `TooManyRetriesError`, and the dedup test sees an empty list instead of a raised throttle. That is the exact silent-`no match` failure `tests/test_rate_limit_guard.py` exists to guard, so it is worth checking whether the new release also degrades a throttled search into "no such item" for real users, not just under the mock.
2. The local-API base URL is now `http://127.0.0.1:23119/api` instead of `http://localhost:23119/api` — related to #148.

Bisected on one machine: same checkout (`8778e62` + #566's branch), same venv, only the pyzotero version differs.

With `pyzotero==1.15.2`:

```
$ .venv/bin/pytest tests/test_rate_limit_guard.py tests/test_local_write_client.py -q
6 failed, 38 passed
$ .venv/bin/pytest --ignore=tests/test_lifespan.py -q
4 failed, 3010 passed, 84 skipped
```

Then, in the same venv, with nothing else changed:

```
$ .venv/bin/pip install 'pyzotero==1.15.1'
$ .venv/bin/pytest tests/test_rate_limit_guard.py tests/test_local_write_client.py tests/test_discovery.py -q
55 passed
$ .venv/bin/pytest --ignore=tests/test_lifespan.py -q
1 failed, 3013 passed, 84 skipped
```

The one failure left at 1.15.1 is `tests/test_webdav.py::test_download_attachment_file_falls_back_to_webdav`, which passes in CI, so it looks like something in my local environment rather than a repo problem.

Reproduce:

```bash
python3 -m venv .venv && .venv/bin/pip install -e . && \
  .venv/bin/pip install pytest pytest-timeout fastmcp pyzotero bibtexparser pymupdf tiktoken chromadb
.venv/bin/pytest tests/test_rate_limit_guard.py tests/test_local_write_client.py -q   # fails on 1.15.2
.venv/bin/pip install 'pyzotero==1.15.1'
.venv/bin/pytest tests/test_rate_limit_guard.py tests/test_local_write_client.py -q   # passes
```

## Options, no strong recommendation

I do not know whether 1.15.2's behaviour is intended upstream or a regression on their side, so I did not open a PR for it. Pinning `pyzotero<1.15.2` in the workflow turns CI green in one line but leaves symptom 1 unexamined, and that symptom looks like the same error-handling bypass as #511. Happy to do either once you say which way you want it, or to test against upstream pyzotero if that is more useful.

mronkko:
Fixes #567.

@feiiiiii5 diagnosed this exactly, down to the release and the two symptoms;
this is that diagnosis turned into a fix. I re-derived the bisect
independently before writing anything, and it holds: **1.15.0 and 1.15.1
still use `localhost` and `client.get`** — both changes are specific to
**1.15.2** (2026-09-18). The 1.15.0 httpx → httpx2 swap this repo already
absorbed (#511, #512) is a separate, earlier event.

Six tests fail on every job of every open PR, and on a pristine `main`. CI
installs `pyzotero` unpinned (`release.yml:41`) against a floor of
`pyzotero>=1.14.0`, so every run since 1.15.2 picks it up.

## Symptom 2 — the local host

pyzotero moved the local API off `localhost` onto `127.0.0.1`
(`_client.py:87`), so the endpoint assertion in
`test_carries_the_key_and_server_id` fails.

We pass `local=True` and never set the endpoint ourselves, so that assertion
was only ever restating upstream's host. It now reads the value back off
pyzotero, through a `pyzotero_local_endpoint()` helper placed beside the
existing `pyzotero_http_module()` and sharing its rationale — so it holds
whichever host upstream picks next.

## Symptom 1 — the dispatch method

Below 1.15.2 a read goes through `client.get`; from 1.15.2 it goes through
`_send()` and `client.request` (`_client.py:285-298`).

`_client()` stubbed `client.get` — so on 1.15.2 the stub was never consulted,
and the five rate-limit tests **issued real requests to api.zotero.org**,
failing with `UserNotAuthorisedError: 403 Invalid key` rather than on anything
they meant to assert. They had stopped exercising the retry path entirely
while still appearing to.

Stubbing `client.request` instead would just pin the other version. The replay
now goes in a `MockTransport` underneath both, which every supported version
dispatches through and which cannot silently fall through to the network.

### This also answers the open question in #567

> it is worth checking whether the new release also degrades a throttled
> search into "no such item" for real users, not just under the mock

**It does not.** The mock was bypassed, not the retry path. With the transport
replaced and `socket.connect` patched to raise, pyzotero 1.15.2 asks the
transport twice for the transient case and returns the item, and three times
for the persistent case before raising `TooManyRetriesError` — `_retrieve_data`
still loops over `MAX_RETRY_ATTEMPTS` and still raises once they are spent
(`_client.py:495-530`). So this is not another #511-shaped error-handling
bypass, and `find_existing_items` still propagates the throttle rather than
degrading to "no match". Those counts are asserted as part of the fix rather
than taken on faith.

That also means pinning `pyzotero<1.15.2` in the workflow was not needed —
nothing in the product is broken by 1.15.2, only two test fixtures that had
written upstream internals out literally.

## Testing

Test-only; no product code changes.

Run across the whole supported range and both HTTP libraries — pyzotero
1.14.0 (httpx) and 1.15.2 (httpx2) — each giving **3305 passed, 1 failed**.

The one failure is `test_webdav.py::test_download_attachment_file_falls_back_to_webdav`,
which is the same one @feiiiiii5 saw at 1.15.1 and correctly set aside as
environmental: it expects the host machine's real "Local Zotero: not
available" string, and fails only because these runs isolate `HOME`. It passes
standalone under either `HOME`, and passes in CI.

## Not included

The same 1.15.2 host change left a matching bug in product code —
`probe_local_server_id` still probes `http://localhost:23119/api/` on the
stated guarantee that it hits exactly where pyzotero writes. That guarantee
now runs backwards. Kept out of this PR to leave it test-only; happy to open
it separately if useful.


54yyyu:
Merged, thanks for checking the retry counts rather than only getting the run green. The probe fix went into 0.12.5 too.

PR #578, opened 2026-09-21
fix: propagate metadata search failures as MCP errors

iamnotmili:
When a Zotero metadata request fails, `_search_with_variants` currently discards the exception and can return an empty or incomplete list. `zotero_search_items` then reports successful results with `isError=false`; the CLI can exit 0 with "No items found" even though it never completed the lookup.

Propagate failed variant requests and raise FastMCP `ToolError` at the `search_items` boundary. A backend failure now stops the cascade and reaches the existing MCP/CLI error handling. Successful empty results, normalization, and normal fallback searches are unchanged. This does not change pagination, optional semantic fallback, or other tools.

Validation on upstream `8778e62`:

- 14 new real in-memory MCP/CLI regressions: **11 failed / 3 passed before**, all pass after. Covers connection failures, timeouts, permission errors, collection reads, a later variant failing after empty or populated results, valid empty/populated results, same-session recovery, and text/JSON CLI exit 1.
- New plus existing search, backend-routing, and global-search tests: **88 passed, 2 existing CI skips**.
- Minimum supported FastMCP 2.14.0 with MCP SDK 1.30.0: **12 protocol tests passed**. CLI compatibility at that minimum was not tested.
- CI-style full suite: **3263 passed, 148 skipped, 2 failed**. Both failures are Windows symlink privilege error 1314 in existing BibTeX/CSL-JSON fixtures. They reproduce on clean upstream; the failure identities match exactly. No test was skipped or changed to hide them.
- Ruff, diff whitespace checks, and sdist/wheel build pass.

Related work: #533 already fixed `get_recent`. Open #542 changes the same search helper for pagination; this patch only changes request-error propagation and should remain separate from that behavior.

No production dependency, installed MCP, live Zotero library, credential, or configuration changes.


54yyyu:
Thanks for the PR, the core point is right: a failed request is not an empty result. I took a narrower version into #580 (0.12.5). It raises when a request failed and nothing was found, and surfaces as a `ToolError`. The difference from yours: if the literal query already found items and a later normalization variant fails, those items are still returned, because the variants only exist to add recall. Closing this in favour of that, let me know if you see a case it misses!

iamnotmili:
Thanks for taking this into 0.12.5. Keeping valid matches when a later normalization variant fails makes sense to me.

I found one edge case in the note-filtering path. In `titleCreatorYear` mode, if a full first page contains only notes and the next-page request fails, `all_items` is nonempty when the failure guard runs. The notes are then filtered out, so `_search_with_variants` returns `[]` despite the failed request.

I reproduced this in an isolated check of the 0.12.5 helper with synthetic backend responses. If the failed page also exhausts the cascade time budget, the tool function can still return “No items found.”

Could the failure guard check the filtered result instead? That would cover this case while preserving the intended behavior when genuine matches were found.

iamnotmili:
> Thanks for the PR, the core point is right: a failed request is not an empty result. I took a narrower version into #580 (0.12.5). It raises when a request failed and nothing was found, and surfaces as a `ToolError`. The difference from yours: if the literal query already found items and a later normalization variant fails, those items are still returned, because the variants only exist to add recall. Closing this in favour of that, let me know if you see a case it misses!

Thanks for taking this into 0.12.5. Keeping valid matches when a later normalization variant fails makes sense to me.

I found one edge case in the note-filtering path. In titleCreatorYear mode, if a full first page contains only notes and the next-page request fails, all_items is nonempty when the failure guard runs. The notes are then filtered out, so _search_with_variants returns [] despite the failed request.

I reproduced this in an isolated check of the 0.12.5 helper with synthetic backend responses. If the failed page also exhausts the cascade time budget, the tool function can still return “No items found.”

Could the failure guard check the filtered result instead? That would cover this case while preserving the intended behavior when genuine matches were found.

54yyyu:
Good catch, thanks for checking the released version! You're right: the guard looked at the unfiltered list, so a full page of notes followed by a failed page passed the check and then got filtered down to nothing. #584 moves the check after the note filter, with a test for exactly that sequence. Matches found before a failure are still returned, as before.

The cascade budget case you mention goes away with it, since the failure now raises out of the helper instead of coming back as an empty list.

PR #579, opened 2026-09-21
fix: BibTeX export crash on parsed database, and missing booktitle for conference papers

dgcnz:
[disclaimer: this was 100% vibe-coded with claude. i needed to fix this for my fork and PR'd here just for completeness. feel free to ignore]

Two BibTeX bugs, found exporting a paper's references with `zotero-cli` in web API mode.

## `export --format bibtex` crashed

```
$ zotero-cli export --item-keys 59SCPEQQ,DBTDVFHQ --format bibtex
Error exporting bibliography: 'BibDatabase' object is not iterable
```

pyzotero parses a `format=bibtex` response into a bibtexparser `BibDatabase` (`pyzotero._decorators._parse_bibtex`), not bytes. The export only decoded bytes, so `_render_entries` tried to iterate the database. It is now serialised back with `bibtexparser.dumps`. The existing tests stub this call with bytes, which is why they passed; the new test returns a real `BibDatabase`.

This is the path that uses Zotero's own BibTeX translator, so with it working, multi-item and collection exports come back complete (venue, pages, DOI).

## `get bibtex` dropped the venue of conference papers

```
$ zotero-cli get bibtex 59SCPEQQ
@inproceedings{Govindarajan2024_59SCPEQQ,
  title = {On Partial Prototype Collapse in the DINO Family of Self-Supervised Methods},
  ...                      # no booktitle
```

A `conferencePaper` keeps its venue in `proceedingsTitle`, and `generate_bibtex`'s field table mapped `publicationTitle` and `bookTitle` but not it. One line adds `("proceedingsTitle", "booktitle")`.

## Tests

- `test_synthesis.py::test_export_bibliography_bibtex_from_parsed_database` fails on `main` with the error above and passes here.
- `test_bibtex_proceedings_title.py` fails on `main` (no `booktitle`) and passes here. It patches Better BibTeX out, so a running Zotero on the test machine cannot answer instead of the local generator.
- Full suite: the only failures are in `test_local_write_client.py` and `test_rate_limit_guard.py`, and the same tests fail on `main` on this machine.

Both fixes also checked against a live library over the web API.


54yyyu:
Merged in #580, in 0.12.5. Both fixes were right, thanks for sending them upstream!

PR #585, opened 2026-09-22
Clip update-db progress lines by display width (CJK/narrow terminals)

jianxing-chen:
## Problem

`update-db --fulltext` renders two single-line `\r` progress displays, and both can wrap the terminal — after which every later redraw repaints only the last physical line and the rows above turn to overlapping residue:

- the indexing line (`\r  [ 42%] 12/28 — Title`) has **no width clipping at all**: a long title runs past the right edge;
- the extraction line (`\r  Processing 3/28 (...) — Title`) clips by **character count**: a CJK title of 57 characters occupies ~114 columns (wide characters render 2 cells), so it wraps an 80-column terminal even though `len()` said it fit;
- the three clear-writes print a fixed 120 spaces, which itself wraps an 80-column terminal and leaves a stray blank line.

Reading Chinese/Japanese/Korean papers — where titles routinely hit 60 characters — makes this a routine garble rather than an edge case.

## Fix

`semantic_search.py` gains three small helpers — `_term_width()`, `_display_cols()` (East-Asian wide characters count as 2, via `unicodedata.east_asian_width`), and `_clip_to_width()` — and:

- both progress lines are clipped to `terminal_width - 1` **display columns** and padded with spaces to the same width, so a shorter redraw fully erases the previous one;
- the clear-writes use the measured terminal width (falling back to 120 when stderr isn't a tty, preserving today's behavior for piped output).

ASCII-only output that already fit is byte-for-byte unchanged.

## Verification

- New regression tests in `tests/test_regression_bugs.py` pin that a 57-char CJK title cannot produce a line wider than the terminal, and that ASCII clips like a plain slice.
- Full suite: `3416 passed, 6 skipped` (deselecting `test_pdf_cascade.py::test_cascade_order` and the two `test_user_agent.py` publisher-PDF cases, which fail on my machine on pristine `main` as well — environment-specific, not from this change).


54yyyu:
Merged, thanks! It went out in 0.13.1. Clipping by display columns is the right fix, and CJK titles were garbling that line for a lot of people.

PR #586, opened 2026-09-22
fix: derive every duplicate-matching key from one module (#496)

### LINKED ISSUE #496: Consolidate scattered duplicate-matching logic behind shared match-key primitives
## Problem

Four places in the codebase answer "are these the same work?", and they disagree substantively — not just stylistically.

| Site | Question | DOI key | Title key |
|---|---|---|---|
| `find_existing_items` (`tools/_helpers.py:787`) | does an item with *this* id exist? | `normalize_doi()`, case **preserved** | none, by design |
| `_maybe_reuse_existing` (`tools/write.py:5039`) | batch-import reuse | delegates to the above | none, "out of scope" |
| `_collect_duplicate_groups` (`tools/write.py:3059`) | partition library into duplicate groups | `data["DOI"].strip().lower()`, **raw field** | `_normalize_dup_title` (`write.py:3048`) |
| preprint filter (`semantic_search.py:1404`) | preprint or published version? | local `norm()` | same `norm()` |

Three concrete defects follow:

1. **`zotero_find_duplicates(method='doi')` misses DOIs stored in URL form.** It keys on the raw field, so `https://doi.org/10.1234/x` and `10.1234/x` land in different groups — while `normalize_doi` handles exactly that, and the import path uses it. The result is that `zotero_add_item` correctly refuses to re-add an item that `find_duplicates` cannot see is duplicated.

2. **Two opposite case policies.** `_collect_duplicate_groups` lowercases DOIs; `normalize_doi` deliberately preserves case and `_matches` compares case-sensitively (`_helpers.py:809`). `10.1234/ABC` and `10.1234/abc` are a duplicate pair to one path and distinct works to the other.

3. **Two incompatible title normalizations.** `_normalize_dup_title` strips punctuation, drops a leading article and keeps spaces; the preprint filter's `norm()` lowercases and strips all whitespace while keeping punctuation. `"The Cat's Meow"` becomes `cats meow` under one and `thecat'smeow` under the other — neither would ever match the other's output for the same input.

## Proposal: share the key derivation, not the detector

`identifiers.py` already exists as the home for this. Its docstring is explicit that it was split out to be public, stdlib-only and free of package imports, precisely so any layer can use the canonical normalisation cheaply. It currently exports only `normalize_doi`.

Add to it:

- `normalize_title_for_matching(title) -> str` — one definition, replacing both current ones
- `metadata_match_keys(item_like) -> set[tuple[str, str]]` — yielding `("doi", …)`, `("title", …)` and so on, with an adapter for API dicts vs `ZoteroItem` objects

Each existing call site then derives keys from one place and keeps its own control flow.

**I would not collapse the four into a single detector function.** They differ in arity (one-to-many lookup, O(n) partition, pairwise preference), in data source (web API dicts vs local-DB objects), and critically in error tolerance: a false positive in `find_existing_items` silently swallows a legitimate add, while a false positive in `find_duplicates` merely shows a human an extra group to review. One function serving all three needs a mode parameter that recreates the divergence internally.

## Hazard: this widens a destructive operation

Fixing defect 1 makes DOI matching strictly *wider*, and `zotero_merge_duplicates(auto=True)` acts on the same grouping via `_collect_duplicate_groups`. Auto-merge would begin merging pairs it currently leaves alone. That is the right behaviour, but it needs tests in `tests/test_duplicates.py` pinning the new grouping before the change lands, not after.

## Related

- Whatever ships first among the current dedup work should put its key derivation in `identifiers.py` rather than inline. Otherwise #493's cross-library fix becomes a fifth site and this consolidation gets harder.
- Separately noticed while mapping this: the preprint/journalArticle filter only runs on the local-DB full-scan path. There is no equivalent on the incremental `item_versions(since=V)` path, so a preprint arriving via incremental sync is indexed even when the published version is already present. Arguably a bug independent of this refactor, and an argument for lifting the preference rule out of the indexing loop regardless.

---
<sub>Code analysis by Claude Code at my request, checked against `main` @ `b48abeb`.</sub>

mronkko:
Fixes #496.

## The bug

Four places answer "is this the same work?", and each normalised identifiers its own way, so one pair of
items could be a duplicate to one of them and not to another. #496 lists the call sites. Three defects
follow from it, all measured on my library (28,419 items):

1. **`zotero_find_duplicates(method='doi')` keyed on the raw field, lowercased.** `10.1000/abc` and
   `https://doi.org/10.1000/abc` were two singletons instead of one group. The same rule made a DOI field
   holding an importer placeholder into a group: **one false group of 72 items keyed on the string
   `article`**, and another of 2 on `editorial`.
2. **`find_existing_items` compared DOIs case-sensitively.** Adding `10.1000/abc` when the library already held
   `10.1000/ABC` created a second copy. **647 items** store a DOI whose canonical form differs from it in case,
   so re-adding any of them in lowercase duplicated it.
3. **The title rules disagreed.** Duplicate grouping *deleted* punctuation, the semantic index's preprint
   filter deleted all whitespace, and neither folded a title the way the Zotero client does.

## What changed

`zotero_mcp.identifiers` holds the keys, and the four sites derive from it. The module stays stdlib-only, as its
docstring promises.

| In `identifiers.py` | What it is |
|---|---|
| `doi_match_key` | case-folded canonical DOI, for equality. `normalize_doi` stays case-preserving, for display. |
| `normalize_title_for_matching` | Zotero's `duplicates.js` `normalizeString`: markup stripped, HTML entities unescaped, diacritics removed, punctuation and symbols mapped to a space, case folded, a leading article dropped |
| `metadata_match_keys` | every `(kind, value)` an item could match under (`doi`, `arxiv`, `isbn`, `title`), over API dicts and `ZoteroItem`s alike. Which kinds decide stays each caller's policy. |
| `normalize_isbn`, `isbn_match_keys`, `normalize_arxiv_id`, `arxiv_identity` | moved from `tools/_helpers.py`. The private names stay there as aliases, so existing callers and patches keep working. |

The four sites:

- **`_collect_duplicate_groups`** (`tools/write.py`): groups on `metadata_match_keys`, filtered by `method`.
- **The auto-merge DOI-conflict guard** (`tools/write.py`): compares `doi_match_key`s, so a group's members count
  as conflicting only when their DOIs really differ. Two unparseable DOI fields still count as a conflict (pinned).
- **`find_existing_items`** (`tools/_helpers.py`): compares canonically. A query DOI that does not parse still
  matches nothing. Without the fallback that guarantees this, it would match every item whose DOI field is empty
  or unparseable, and an `if_exists='update'` add would overwrite an unrelated paper (pinned).
- **The preprint filter** (`semantic_search.py`): uses the same `doi`/`title` keys as duplicate grouping.

The auto-merge rule itself is unchanged: DOI-only by default, mixed item types skipped, genuinely conflicting
DOIs skipped, and nothing acts until the caller echoes back the plan token.

## Measured

`scripts/measure_match_keys.py` runs the old and new rules over the same library and diffs the groups. It is
read-only, works with Zotero open, and reproduces every number here:

|  | old | new | change |
|---|---:|---:|---:|
| DOI groups | 329 | 335 | +6 |
| items in them | 823 | 767 | −56 |
| DOI pairs gained / lost | | | +13 / −2,557 |
| title groups | 782 | 788 | +6 |
| items in them | 1,880 | 1,892 | +12 |
| title pairs gained / lost | | | +10 / −4 |

**Two things get narrower, on purpose:**

- **Placeholder DOIs no longer form groups.** A DOI field holding something that is not a DOI yields no key.
  All 2,557 lost DOI pairs come from the two placeholder groups (72 × `article`, 2 × `editorial`); **no pair of
  real, equal DOIs stopped matching.**
- **Title matching moves both ways**: 10 item pairs now match that did not, and 4 no longer do. The gains are
  titles that differ only in punctuation, dashes or markup: `Tool-an Analysis` vs `Tool – an Analysis`,
  `start up` vs `Start-Up`, `<i>SMJ</i>` vs `SMJ`. The losses are pairs the old rule joined only by gluing
  punctuation out: `U.S. Ventures` vs `US Ventures`, `Attackers` vs `attacker's`. Zotero's `normalizeString`
  keeps those apart too. A title match is a candidate, not a verdict, so the script reports movement in both
  directions rather than calling either side right.

## Carried over from 0.12.5

#571's fix for dotted legacy arXiv ids (`math.GT/0309136`) edited `_normalize_arxiv_id` in `tools/_helpers.py`.
That function now lives in `identifiers.py`, so the fix does too, and #571's own tests pass through the alias.

## Testing

**50 new tests**: `tests/test_identifiers_match_keys.py` and `tests/test_semantic_preprint_filter_keys.py`
(new), plus additions to `tests/test_duplicates.py` and `tests/test_idempotent_adds.py`. They pin the
duplicate-grouping invariants before the change, then each defect above, the placeholder narrowing, the
unparseable-DOI conflict rule and the title folding.

Suite in a venv built from CI's install line (Python 3.12, pyzotero 1.15.2, fastmcp 3.4.7): **3463 passed,
1 failed**. The failure, `test_webdav.py::test_download_attachment_file_falls_back_to_webdav`, fails
identically on `main` on my machine: it asserts the exact list of local probes, and that list depends on
whether the machine has a Zotero database. `ruff check` findings are unchanged from `main`.

This moves no module and depends on nothing else in flight.


54yyyu:
Thanks for this, pulling the matching rules into one place is a real improvement, and the canonical DOI grouping and dropping the placeholder-DOI group are both good catches. A few things before it goes in:

1. **Leading-article strip runs after punctuation is mapped to spaces.** So `A/B testing` and `A-B testing` both fold to `b testing` and match `B Testing`. Stripping the article before the punctuation step (or only when the next word is alphabetic) fixes it. Worth a test.
2. **Speed.** On a 44k-item library `metadata_match_keys` takes about 1.05 s against 0.26 s before, mostly the per-character `unicodedata.category` loop, and every caller only uses the DOI and title keys (0.84 s with just those). A `kinds=` parameter, or computing arXiv/ISBN keys only when asked, would keep the semantic full scan from paying for keys it throws away.
3. `_helpers.py` re-exports private names from `identifiers` under a `noqa` (`_TITLE_TAG_RE`, the ISBN checksum helpers). Either make them public or have the tests import them from `identifiers`.
4. Could `scripts/measure_match_keys.py` move to the PR description or a gist? It is a one-off. And the CHANGELOG entry can be two or three sentences.

Since this widens what auto-merge acts on (canonical DOI forms), I'd like it in the next minor release rather than a patch. Happy to look again once these are in!

mronkko:
All four addressed in e9a88aa, and the branch is merged with 0.13.1 (4e639ec).

1. **Leading article.** The strip now runs before punctuation becomes spaces and needs whitespace after the article, so `A/B testing` and `A-B testing` fold to `a b testing`. I tried "only when the next word is alphabetic" first, but it also kept real articles (`The R book`, `The 1999 Irish nurses' strike`, `The M-form Society`, 32 items in my library). With the whitespace rule, one item's key changes, and that one is correct: `THE* Rules of …`, where THE is an acronym. Tests cover both directions. While checking this I found my description was wrong: Zotero's `normalizeString` does not drop articles. That step comes from our old `_normalize_dup_title`. I've corrected the docstring and the description.
2. **Speed.** `metadata_match_keys(item, kinds=...)`, and both callers ask for `("doi", "title")`. ASCII titles (93% of mine) also skip the `unicodedata.category` loop via one regex, checked equal to the slow path on all 128 ASCII characters. Over 44,859 items: 0.56 s → 0.23 s for what the callers use (all kinds: 0.33 s; `main`'s old rule: 0.15 s).
3. **Private re-exports.** Gone. `_helpers.py` imports only what it uses. `TITLE_TAG_RE` is public, since `_title_search_query` uses it, and `test_add_by_isbn.py` imports the checksum helpers from `identifiers`.
4. **Script and CHANGELOG.** The script is out of the tree and linked from the description as a permalink. The CHANGELOG entry is three sentences, under Unreleased.

The measured group and pair counts don't change with these fixes. Next minor is fine.

54yyyu:
Merged, thanks! I checked all four: `A/B testing` / `A-B testing` no longer match `B Testing`, `kinds=` keeps the callers close to the old cost, the private re-exports are gone, and the script is out of the tree. I also tried unparseable query DOIs against items with empty or placeholder DOIs, and `find_existing_items` returns nothing for them, so `if_exists='update'` is safe. This will go out with the next minor release.

---

PR #587, opened 2026-09-22
RFC: a target package layout, and the scaffolding to move modules without breaking imports

mronkko:
**This moves no module and changes no runtime behaviour.** It is the groundwork for reorganising the flat
`src/zotero_mcp/` into packages. It holds a written target layout, plus the tooling that lets a module move
without breaking anyone's imports. Three moves built on it are ready on my fork (below); none is opened yet,
because each depends on this one.

## Why

`src/zotero_mcp/` has **40 modules at the root**. Related code is spread across names that don't say what the
modules hold: `pdf_utils.py`, `epub_utils.py`, `pdfannots_helper.py` and `extract.py` are one concern, and
`semantic_search.py`, `chroma_client.py`, `batch_common.py`, `openai_batch.py`, `gemini_batch.py` and
`embeddings/` are another. Moving them is mechanical, but it has two hazards that the tooling here is for:

1. **External callers import these paths.** A move has to leave the old path working and say where the code went.
2. **A patch on an old path silently stops working.** `monkeypatch.setattr("zotero_mcp.chroma_client.X", ...)`
   against a forwarding shim sets an attribute on the shim, not on the moved module, so the test passes
   having tested nothing.

## What's in it

- **`docs/architecture.md`**: the target package map (`backends/`, `attachments/`, `metadata_sources/`,
  `semantic_search/`, `cli/`, `formatting/`), with an import-cost rule for each package's `__init__`. It also
  states the naming rules (Zotero's own vocabulary where it exists; no `_utils` / `_helpers` / `_common`), the
  test layout, how shims work, the import-cost budget and a checklist for moving a module. Every entry is
  marked **now** or **target**, and the shim table is empty.
- **`_shim.forwarder`**: a lazy PEP 562 `__getattr__` for a moved module's old path. Importing the old path
  costs nothing, and a subprocess `-X importtime` test proves the target is not loaded. Each *use* warns with
  `MovedModuleWarning` (a `DeprecationWarning`) at the caller's line. **The release plan is two constants,
  `MOVED_IN` and `REMOVED_IN`,** and nothing else in the tree may spell either version out; a test enforces
  that. They are placeholders (0.14.0 / 0.15.0), meant to be set to whatever release the moves actually ship in.
- **`tests/test_module_layout.py`**: a `SHIMS` map, empty here, and two guards. No test and no production module
  may import or patch a shimmed path; each hit is reported as `file:line: old -> use new`. With the map empty
  the guards pass vacuously, so a separate test runs the scan over the real tree against a synthetic row.
- **A suite-wide gate** that turns `MovedModuleWarning` into an error, so a *lazy* import of an old path, which
  no static scan can see, fails the test that reaches it. It is armed from `conftest.py`, because the
  interpreter's `-W` flag can't survive conftest's `sys.modules` purge. There's a test that the gate really
  bites, and `ZOTERO_MCP_ALLOW_SHIM_PATHS=1` turns it off.
- **Two test-isolation fixes with value of their own:**
  - `conftest.py` points `HOME` at a temp dir, so a running `zotero-mcp` holding `update.lock` can no longer fail
    the suite. On my machine it failed 51 tests that way.
  - `test_webdav.py::test_download_attachment_file_falls_back_to_webdav` no longer asserts the exact list of
    local probes, which depended on whether the machine has a Zotero database.

`_shim.py` is the only file under `src/` this touches, and nothing imports it yet.

## Testing

Suite in a venv built from CI's install line (Python 3.12, pyzotero 1.15.2, fastmcp 3.4.7): **3447 passed,
0 failed**. On `main` in the same venv: 3412 passed, 1 failed (the WebDAV test above). `ruff check` findings
are unchanged from `main`.

## The moves, and what I'd ask

Each move is on my fork and sits on this branch. Each has the same shape: a pure `git mv` commit (so
`git log --follow` works), then the import rewrite, then the shims. Each also moves its tests (e.g. to
`tests/semantic_search/`) and records its shims in the table and in `SHIMS`.

| Move | Diff against this branch | Old paths kept as shims |
|---|---|---|
| `attachments/` | [refactor/02-attachments](https://github.com/mronkko/zotero-mcp/compare/refactor/00-scaffolding...refactor/02-attachments) | 7 |
| `semantic_search/` | [refactor/03-semantic-search](https://github.com/mronkko/zotero-mcp/compare/refactor/00-scaffolding...refactor/03-semantic-search) | 9 |
| `cli/` | [refactor/04-cli](https://github.com/mronkko/zotero-mcp/compare/refactor/00-scaffolding...refactor/04-cli) | 6 |

All three pass the full suite. `attachments/` and `cli/` also record before/after checks in their CHANGELOG
entries: `import zotero_mcp.server` loads the same modules, and `zotero-cli --json-schema` and `--help` are
byte-identical. `cli/` also keeps every console script that is already installed resolving silently.
`backends/`, `metadata_sources/`, `formatting/` and the `utils.py` split are not written yet.

Two things I'd value your view on before opening them:

1. **The names in the package map.** They are what's expensive to change once the moves land.
2. **Landing the three moves in one release**, so `MOVED_IN` stays a single value. I can send them one after
   another right after a release, rebasing each as the previous one lands. Or, if you'd rather, take them
   together on a branch the way you did with #502.

The shared-match-keys fix for #496 is a separate PR (#586). It moves no module and doesn't depend on this.


mronkko:
I have a bit of extra Claude Code usage for this month and thought I could do a bit of organization, removing duplicate code, and addressing some bugs related to search and semantic search. This will be a sequence of about 10 PRs.

54yyyu:
Thanks for writing this up so carefully. The layout reads sensibly and the lazy-`__init__` rule per package is the right constraint.

I'm going to hold this one for now rather than merge it. The scaffolding itself is inert, but merging it effectively commits us to moving 40 modules behind a deprecation cycle, and I'm not ready to make that call yet: a lot of the core is likely to move to a separate engine over the next while, and I'd rather not restructure the Python side twice. I'll come back to it once that direction is clearer.

One concrete thing if it does go ahead: `tests/conftest.py` points `HOME` at a temp directory at import time. In subprocess tests that also moves `site.USER_SITE`, so on an install where packages live under `~/.local` (mine does) `test_lifespan.py` and `test_context_cost_claim.py` fail with import errors (16 locally, main is clean). Isolating only `~/.config/zotero-mcp`, or keeping `PYTHONUSERBASE`, would avoid it.

Please don't start the three follow-up moves on this basis yet. Smaller cleanups like #586 that don't move files are very welcome in the meantime!

mronkko:
I will hold this off. But I think a bit of housekeeping is warranted. Good that somethign like this is planned.

